PR #77, opened 2024-01-19
feat: Remove magic strings from Source type

pmengelbert:
Replaces PR #57 , see #57 for all comments which have been addressed by now.

There was a snafu with the Azure/dalec repo going public, so I had to re-fork.

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -7,8 +7,6 @@ import (
 	"regexp"
 	"strings"
 	"time"
-
-	"github.com/invopop/jsonschema"
 )
 
 // Spec is the specification for a package build.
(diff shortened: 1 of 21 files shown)


spec.go
```diff
@@ -179,71 +177,95 @@ type ImageConfig struct {
...
+// Source defines a source to be used in the build.
+// A source can be a local directory, a git repositoryt, http(s) URL, etc.
+type Source struct {
+	// This is an embedded union representing all of the possible source types.
+	// Exactly one must be non-nil, with all other cases being errors.
+	//
+	// === Begin Source Variants ===
+	DockerImage *SourceDockerImage `yaml:"image,omitempty" json:"image,omitempty"`
+	Git         *SourceGit         `yaml:"git,omitempty" json:"git,omitempty"`
+	HTTPS       *SourceHTTPS       `yaml:"https,omitempty" json:"https,omitempty"`
```
cpuguy83, line 256: Nit: this should probably just be `http`.
Tried to think of a better name than that but am coming up short.

pmengelbert: The field name? The Struct name? The json tag? All of the above?

cpuguy83: All of the above.

pmengelbert: done

cpuguy83 (approved):

PR #80, opened 2024-01-24
passthrough custom frontends for integration framework

### LINKED ISSUE #52: Improve target forwarding test to include a real custom frontend
Right now the test just re-uses the existing frontend just under a different name in the spec (thereby creating a different build target that we can select).

I think we can build a custom frontend for testing purposes and upload to the internal test registry to produce whatever we need for testing to thoroughly exercise the codepath.

cpuguy83:
This Is replacing a PR #56 that GH eviscerated due to flipping the repo public.
Closes #52


adamperlin:
I already looked over this one and approved, so perhaps @pmengelbert should take a look?

pmengelbert (approved):
Adam has already reviewed and approved these changes.

PR #84, opened 2024-01-26
Passthrough Target Platform Build Vars

### LINKED ISSUE #67: Passthrough TARGET{PLATFORM,OS,ARCH<VARIANT} vars
These are things that Dockerfile supports, it would be nice to pass this through.

adamperlin:
Resolves #67. Adds support for passing through `TARGET{OS, ARCH, VARIANT,PLATFORM}` build variables, as would be available in a Dockerfile. This feature could be used to support multi-arch builds in a single Dalec spec file. See the following example for an idea of how this feature works:

```yaml
args:
    # MUST opt-in to the build arg by listing it in the spec args in order to have the value automatically injected
    TARGETARCH:
build:
  env:
    # must explicitly map target platform build args to environment variables
    TARGETARCH: ${TARGETARCH}
   steps:
     - command: |
          case ${TARGETARCH} in
                amd64)
                    export GOARCH=amd64;
                     ;;
                 arm64)
                    export GOARCH=arm64;
                    ;;
                  *)
                     echo "unsupported arch ${TARGETARCH}"
                     exit
                    ;;
            esac
            go build ./src
            ...
```

A notable change in this PR is a change to when build-arg substitution happens. Buildkit builds are triggered per-platform, and thus the value of `TARGET*` variables varies per build. Thus, we cannot bind the values of these variables until we are about to trigger the Buildkit build for a particular platform.  Because the values of existing build-args won't have changed at that point, the most straightforward course of implementation was to delay doing all build-arg substitution until we have platform information, which happens in the callback passed to `bc.Build()`.


docker-bake.hcl
```diff
@@ -152,6 +152,28 @@ target "test-fixture" {
     cache-to = DALEC_NO_CACHE_EXPORT != "1" ? ["type=gha,scope=dalec/${f}/${tgt}/${f},mode=max"] : []
 }
 
+target "test-multiplatform" {
+    name = "test-multiplatform-${f}-${replace(plat, "/", "-")}"
+    matrix = {
+        plat = ["linux/amd64", "linux/arm32/v5", "windows/amd64"]
```
cpuguy83, line 158:
```suggestion
    plat = ["linux/amd64", "linux/arm/v5", "windows/amd64"]
```

frontend/build.go
```diff
@@ -59,6 +60,30 @@ func makeRequestHandler(target string) dockerui.RequestHandler {
 	return h
 }
 
+var PassthroughGetters = map[string]func(ocispecs.Platform) string{
```
cpuguy83, line 63: Does this need to be exported?

test/fixtures/platform-vars-linux-amd64.yml
```diff
@@ -0,0 +1,46 @@
+args:
+  TARGETARCH:
+  TARGETOS:
+  TARGETVARIANT:
+  TARGETPLATFORM:
+
+name: dalec-test-nested-build
```
cpuguy83, line 7: Looks like this is a copy-paste from another spec.

frontend/build.go
```diff
@@ -59,6 +60,30 @@ func makeRequestHandler(target string) dockerui.RequestHandler {
...
+	},
+	"TARGETARCH": func(p ocispecs.Platform) string {
+		return p.Architecture
+	},
+	"TARGETVARIANT": func(p ocispecs.Platform) string {
+		return p.Variant
+	},
+	"TARGETPLATFORM": func(p ocispecs.Platform) string {
+		return platforms.Format(p)
+	},
```
cpuguy83, line 75: How about `BUIILDOS`, `BUILDARCH`, etc...?

cpuguy83 (changes requested):
Can you use the new integration test suite instead of adding more to the bake file?
We should be able to use the `debug/resolve` build target to get the fully resolved spec file out.

pmengelbert:
I'd like to see this added to the docs, especially since there's a requirement to opt in by putting it in the spec.

frontend/build.go
adamperlin: Implemented. Weirdly, the docker client stores an array of build platforms, but only one default build platform ever seems to be set, so I am assuming a singular build platform for now

adamperlin:
> Can you use the new integration test suite instead of adding more to the bake file? We should be able to use the `debug/resolve` build target to get the fully resolved spec file out.

Yes, done! It is much cleaner now, I think.

load.go
```diff
@@ -335,8 +342,30 @@ func (c *Command) processBuildArgs(lex *shell.Lex, args map[string]string, name
...
+			}
+			strip := DefaultPatchStrip
+			s.Patches[k][i].Strip = &strip
+		}
+	}
+
+	return s
+}
+
 func (s Spec) Validate() error {
```
adamperlin, line 363: After splitting out the build-arg substitution logic from spec loading, I think it makes sense to split out the validation and default filling logic too (there was a bug with duplicate validation and default filling)

frontend/build.go
```diff
@@ -59,6 +60,38 @@ func makeRequestHandler(target string) dockerui.RequestHandler {
...
+
+var passthroughGetters = map[string]func(ocispecs.Platform) string{
+	"OS":       getOS,
+	"ARCH":     getArch,
+	"VARIANT":  getVariant,
+	"PLATFORM": getPlatformFormat,
+}
+
+func fillPlatformArgs(prefix string, args map[string]string, platform ocispecs.Platform) map[string]string {
+	args = dalec.DuplicateMap(args)
```
cpuguy83, line 87: How come we are duplicating and returning a new map?

load.go
```diff
@@ -197,101 +207,98 @@ func (s *SourceBuild) validate(failContext ...string) (retErr error) {
...
-				continue
-			}
-			strip := DefaultPatchStrip
-			spec.Patches[k][i].Strip = &strip
-		}
+	return nil
+}
+
+// LoadSpec loads a spec from the given data.
+// env is a map of environment variables to use for shell-style expansion in the spec.
```
cpuguy83, line 289: Extra doc from the old impl.

load.go
```diff
@@ -335,8 +342,30 @@ func (c *Command) processBuildArgs(lex *shell.Lex, args map[string]string, name
 	return nil
 }
 
+func (s Spec) FillDefaults() Spec {
+	for _, src := range s.Sources {
```
cpuguy83, line 346: I think this is not doing what we want it to since `Sources` is `map[string]Source` and not `map[string]*Source`.

To make this work we either need to do something like:
```suggestion
	for i, src := range s.Sources {
		s := &src
		fillDefaults(&s)
		s.Sources[i] = *s
	}
```

Or I think we can just change `Sources` to be a `

cpuguy83: Also seems like we must be missing a test for this.

load.go
```diff
@@ -335,8 +342,30 @@ func (c *Command) processBuildArgs(lex *shell.Lex, args map[string]string, name
 	return nil
 }
 
+func (s Spec) FillDefaults() Spec {
```
cpuguy83, line 345: Can we do this on `*Spec` and not return another spec?

cpuguy83 (changes requested):

load.go
adamperlin: Good catch, thank you! I believe the original code was putting the filled source back in the source map as you have it here, but I must have forgotten that line. We should definitely add a test that would catch this

frontend/build.go
adamperlin: This should probably be moved. When this function is used, it is initially passed `bc.BuildArgs` from the docker client. I didn't want to mutate the underlying `BuildArgs` map as that seems like confusing behavior, but I could instead do the copy before ever calling this function, leading to only one copy instead of a copy on **every** call.

cpuguy83 (approved):

PR #87, opened 2024-01-30
Add support for jammy

cpuguy83:
What Dalec doesn't like a good jammy dodger?

image

### DIFF
--- a/frontend/deb/debroot.go
+++ b/frontend/deb/debroot.go
@@ -80,6 +80,11 @@ func Debroot(sOpt dalec.SourceOpts, spec *dalec.Spec, worker, in llb.State, targ
 	base := llb.Scratch().File(llb.Mkdir(dir, 0o755), opts...)
 	installers := createInstallScripts(worker, spec, dir)
 
+	const (
+		sourceFormat  = "3.0 (quilt)"
+		sourceOptions = "create-empty-orig"
+	)
+
 	debian := base.
 		File(llb.Mkdir(filepath.Join(dir, "source"), 0o755), opts...).
 		With(func(in llb.State) llb.State {
@@ -87,8 +92,8 @@ func Debroot(sOpt dalec.SourceOpts, spec *dalec.Spec, worker, in llb.State, targ
 				return in
 			}
 			return in.
-				File(llb.Mkfile(filepath.Join(dir, "source/format"), 0o640, []byte("3.0 (quilt)")), opts...).
-				File(llb.Mkfile(filepath.Join(dir, "source/options"), 0o640, []byte("create-empty-orig")), opts...)
+				File(llb.Mkfile(filepath.Join(dir, "source/format"), 0o640, []byte(sourceFormat)), opts...).
+				File(llb.Mkfile(filepath.Join(dir, "source/options"), 0o640, []byte(sourceOptions)), opts...)
 		}).
 		File(llb.Mkdir(filepath.Join(dir, "dalec"), 0o755), opts...).
 		File(llb.Mkfile(filepath.Join(dir, "source/include-binaries"), 0o640, append([]byte("dalec"), '\n')), opts...)
@@ -256,6 +261,8 @@ func createInstallScripts(worker llb.State, spec *dalec.Spec, dir string) []llb.
 	})
 
 	writeInstall := func(src, dir, name string) {
+		// This is wrapped in a sync.OnceFunc so that this only has an effect the
+		// first time it is called.
 		writeInstallHeader()
 
 		if filepath.Base(src) != name {
--- a/frontend/deb/debroot.go
+++ b/frontend/deb/debroot.go
@@ -34,12 +34,17 @@ func sourcePatchesDir(sOpt dalec.SourceOpts, base llb.State, dir, name string, s
 	for _, patch := range spec.Patches[name] {
 		src := spec.Sources[patch.Source]
 
+		copySrc := patch.Source
+		if patch.Path != "" {
+			src.Includes = append(src.Includes, patch.Path)
+			copySrc = filepath.Base(patch.Path)
+		}
 		st, err := src.AsState(patch.Source, sOpt, opts...)
 		if err != nil {
 			return nil, errors.Wrap(err, "error creating patch state")
 		}
 
-		st = base.File(llb.Copy(st, patch.Source, filepath.Join(patchesPath, patch.Source)), opts...)
+		st = base.File(llb.Copy(st, copySrc, filepath.Join(patchesPath, patch.Source)), opts...)
 		if _, err := seriesBuf.WriteString(name + "\n"); err != nil {
 			return nil, errors.Wrap(err, "error writing to series file")
 		}
@@ -97,7 +102,7 @@ func Debroot(sOpt dalec.SourceOpts, spec *dalec.Spec, worker, in llb.State, targ
 
 	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/build.sh"), 0o700, createBuildScript(spec)), opts...))
 	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/patch.sh"), 0o700, createPatchScript(spec)), opts...))
-	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/fix_file_backed_sources.sh"), 0o700, fixupFileBackedSourcesScript(spec)), opts...))
+	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/fix_sources.sh"), 0o700, fixupSources(spec)), opts...))
 	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/fix_perms.sh"), 0o700, fixupArtifactPerms(spec)), opts...))
 
 	customEnable, err := customDHInstallSystemdPostinst(spec)
@@ -157,33 +162,31 @@ func fixupArtifactPerms(spec *dalec.Spec) []byte {
 	return buf.Bytes()
 }
 
-// for debian sources, file backed sources are not in the correct format as expected by dalec.
-// This script fixes that by moving the file to the correct location.
-// It is called from `debian/rules` after the source tarball has been extracted.
-func fixupFileBackedSourcesScript(spec *dalec.Spec) []byte {
-	// first, don't bother doing this for patches, which we don't include in our source tarballs.
-	patches := map[string]struct{}{}
-	for _, patchList := range spec.Patches {
-		for _, patch := range patchList {
-			patches[patch.Source] = struct{}{}
-		}
-	}
-
+// For debian sources
+//  1. File backed sources are not in the correct format as expected by dalec.
+//  2. Sources with certain characters in the name had to be changed, so we need
+//     to bring those back.
+//
+// This is called from `debian/rules` after the source tarball has been extracted.
+func fixupSources(spec *dalec.Spec) []byte {
 	buf := bytes.NewBuffer(nil)
 	writeScriptHeader(buf)
 
 	// now, we need to find all the sources that are file-backed and fix them up
 	for name, src := range spec.Sources {
-		if _, ok := patches[name]; ok {
-			continue
-		}
+		dirName := sanitizeSourceKey(name)
+
 		if dalec.SourceIsDir(src) {
+			if dirName == name {
+				continue
+			}
+			fmt.Fprintf(buf, "mv '%s' '%s'\n", dirName, name)
 			continue
 		}
 
-		fmt.Fprintf(buf, "mv %s/%s %s.dalec.tmp\n", name, name, name)
-		fmt.Fprintln(buf, "rm -rf", name)
-		fmt.Fprintf(buf, "mv \"%s.dalec.tmp\" %q\n", name, name)
+		fmt.Fprintf(buf, "mv '%s/%s' '%s.dalec.tmp' || (ls -lh %q; exit 2)\n", dirName, name, name, dirName)
+		fmt.Fprintf(buf, "rm -rf '%s'\n", dirName)
+		fmt.Fprintf(buf, "mv '%s.dalec.tmp' '%s'\n", name, name)
 		fmt.Fprintln(buf)
 	}
 
--- a/frontend/deb/debroot.go
+++ b/frontend/deb/debroot.go
@@ -1,16 +1,22 @@
 package deb
 
 import (
+	"bufio"
 	"bytes"
+	"context"
 	_ "embed"
 	"fmt"
 	"io"
 	"path/filepath"
 	"runtime/debug"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/Azure/dalec"
+	"github.com/Azure/dalec/frontend/pkg/bkfs"
 	"github.com/moby/buildkit/client/llb"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/pkg/errors"
 )
 
@@ -57,7 +63,15 @@ func sourcePatchesDir(sOpt dalec.SourceOpts, base llb.State, dir, name string, s
 
 // Debroot creates a debian root directory suitable for use with debbuild.
 // This does not include sources in case you want to mount sources (instead of copying them) later.
-func Debroot(sOpt dalec.SourceOpts, spec *dalec.Spec, worker, in llb.State, target, dir string, opts ...llb.ConstraintsOpt) (llb.State, error) {
+//
+// Set the `distroVersionID` argument to a value suitable for including in the
+// .deb for storing the targeted distro+version in the deb.
+// This is generally needed so that if a distro user upgrades from, for instance,
+// debian 11 to debian 12, that the package built for debian 12 will be considered
+// an upgrade even if it is technically the same underlying source.
+// It may be left blank but is highly recommended to set this.
+// Use [ReadDistroVersionID] to get a suitable value.
+func Debroot(sOpt dalec.SourceOpts, spec *dalec.Spec, worker, in llb.State, target, dir, distroVersionID string, opts ...llb.ConstraintsOpt) (llb.State, error) {
 	control, err := controlFile(spec, in, target, dir)
 	if err != nil {
 		return llb.Scratch(), errors.Wrap(err, "error generating control file")
@@ -68,7 +82,7 @@ func Debroot(sOpt dalec.SourceOpts, spec *dalec.Spec, worker, in llb.State, targ
 		return llb.Scratch(), errors.Wrap(err, "error generating rules file")
 	}
 
-	changelog, err := Changelog(spec, in, target, dir)
+	changelog, err := Changelog(spec, in, target, dir, distroVersionID)
 	if err != nil {
 		return llb.Scratch(), errors.Wrap(err, "error generating changelog file")
 	}
@@ -442,4 +456,60 @@ func controlFile(spec *dalec.Spec, in llb.State, target, dir string) (llb.State,
 			File(llb.Mkdir(dir, 0o755, llb.WithParents(true))).
 			File(llb.Mkfile(filepath.Join(dir, "control"), 0o640, buf.Bytes())),
 		nil
+}
+
+// ReadDistroVersionID returns a string concatenating the values of ID and
+// VERSION_ID from /etc/os-release from the provided state.
+func ReadDistroVersionID(ctx context.Context, client gwclient.Client, st llb.State) (string, error) {
+	rootfs, err := bkfs.FromState(ctx, &st, client)
+	if err != nil {
+		return "", err
+	}
+
+	f, err := rootfs.Open("etc/os-release")
+	if err != nil {
+		return "", err
+	}
+
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	var (
+		id      string
+		version string
+	)
+
+	for scanner.Scan() {
+		k, v, ok := strings.Cut(scanner.Text(), "=")
+		if !ok {
+			continue
+		}
+		switch k {
+		case "ID":
+			id = unquote(v)
+		case "VERSION_ID":
+			version = unquote(v)
+		}
+
+		if id != "" && version != "" {
+			break
+		}
+	}
+
+	if scanner.Err() != nil {
+		return "", err
+	}
+
+	if id == "" || version == "" {
+		return "", errors.New("could not determine distro or version ID")
+	}
+
+	return id + version, nil
+}
+
+func unquote(v string) string {
+	if updated, err := strconv.Unquote(v); err == nil {
+		return updated
+	}
+	return v
 }
--- a/frontend/deb/debroot.go
+++ b/frontend/deb/debroot.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"path/filepath"
 	"runtime/debug"
-	"strings"
 	"sync"
 
 	"github.com/Azure/dalec"
@@ -259,7 +258,7 @@ func createInstallScripts(worker llb.State, spec *dalec.Spec, dir string) []llb.
 	writeInstall := func(src, dir, name string) {
 		writeInstallHeader()
 
-		if filepath.Base(src) != name || strings.Contains(src, "*") {
+		if filepath.Base(src) != name {
 			fmt.Fprintln(installBuf, src, "=>", filepath.Join(dir, name))
 			return
 		}
--- a/frontend/deb/debroot.go
+++ b/frontend/deb/debroot.go
@@ -131,7 +131,7 @@ func Debroot(sOpt dalec.SourceOpts, spec *dalec.Spec, worker, in llb.State, targ
 func fixupArtifactPerms(spec *dalec.Spec) []byte {
 	buf := bytes.NewBuffer(nil)
 
-	fmt.Fprintln(buf, "#!/usr/bin/env bash")
+	fmt.Fprintln(buf, "#!/usr/bin/env sh")
 	fmt.Fprintln(buf, "set -ex")
 	fmt.Fprintln(buf)
 
@@ -141,7 +141,7 @@ func fixupArtifactPerms(spec *dalec.Spec) []byte {
 		return nil
 	}
 
-	sorted := dalec.SortMapKeys(spec.Artifacts.Directories.Config)
+	sorted := dalec.SortMapKeys(spec.Artifacts.Directories.GetConfig())
 	for _, name := range sorted {
 		cfg := spec.Artifacts.Directories.Config[name]
 		if cfg.Mode.Perm() != 0 {
@@ -150,7 +150,7 @@ func fixupArtifactPerms(spec *dalec.Spec) []byte {
 		}
 	}
 
-	sorted = dalec.SortMapKeys(spec.Artifacts.Directories.State)
+	sorted = dalec.SortMapKeys(spec.Artifacts.Directories.GetState())
 	for _, name := range sorted {
 		cfg := spec.Artifacts.Directories.State[name]
 		if cfg.Mode.Perm() != 0 {
--- a/frontend/deb/debroot.go
+++ b/frontend/deb/debroot.go
@@ -1,0 +1,352 @@
+package deb
+
+import (
+	"bytes"
+	_ "embed"
+	"fmt"
+	"io"
+	"path/filepath"
+	"runtime/debug"
+	"sync"
+
+	"github.com/Azure/dalec"
+	"github.com/moby/buildkit/client/llb"
+	"github.com/pkg/errors"
+)
+
+//go:embed templates/patch-header.txt
+var patchHeader []byte
+
+// This creates a directory in the debian root directory for each patch, and copies the patch files into it.
+// The format for each patch dir matches what would normaly be under `debian/patches`, just that this is a separate dir for every source we are patching
+// This is purely for documenting in the source package how patches are applied in a more readable way than the big merged patch file.
+func sourcePatchesDir(sOpt dalec.SourceOpts, base llb.State, dir, name string, spec *dalec.Spec, opts ...llb.ConstraintsOpt) ([]llb.State, error) {
+	patchesPath := filepath.Join(dir, name)
+	base = base.
+		File(llb.Mkdir(patchesPath, 0o755), opts...)
+
+	var states []llb.State
+
+	seriesBuf := bytes.NewBuffer(nil)
+	for _, patch := range spec.Patches[name] {
+		src := spec.Sources[patch.Source]
+
+		st, err := src.AsState(patch.Source, sOpt, opts...)
+		if err != nil {
+			return nil, errors.Wrap(err, "error creating patch state")
+		}
+
+		st = base.File(llb.Copy(st, patch.Source, filepath.Join(patchesPath, patch.Source)), opts...)
+		if _, err := seriesBuf.WriteString(name + "\n"); err != nil {
+			return nil, errors.Wrap(err, "error writing to series file")
+		}
+		states = append(states, st)
+	}
+
+	series := base.File(llb.Mkfile(filepath.Join(patchesPath, "series"), 0o440, seriesBuf.Bytes()), opts...)
+
+	return append(states, series), nil
+}
+
+// Debroot creates a debian root directory suitable for use with debbuild.
+// This does not include sources in case you want to mount sources (instead of copying them) later.
+func Debroot(sOpt dalec.SourceOpts, spec *dalec.Spec, in llb.State, target, dir string, opts ...llb.ConstraintsOpt) (llb.State, error) {
+	control, err := controlFile(spec, in, target, dir)
+	if err != nil {
+		return llb.Scratch(), errors.Wrap(err, "error generating control file")
+	}
+
+	rules, err := Rules(spec, in, dir)
+	if err != nil {
+		return llb.Scratch(), errors.Wrap(err, "error generating rules file")
+	}
+
+	changelog, err := Changelog(spec, in, target, dir)
+	if err != nil {
+		return llb.Scratch(), errors.Wrap(err, "error generating changelog file")
+	}
+
+	if dir == "" {
+		dir = "debian"
+	}
+
+	base := llb.Scratch().File(llb.Mkdir(dir, 0o755), opts...)
+	installers := createInstallScripts(spec, dir)
+
+	debian := base.
+		File(llb.Mkfile(filepath.Join(dir, "compat"), 0o440, []byte("10")), opts...).
+		File(llb.Mkdir(filepath.Join(dir, "source"), 0o755), opts...).
+		File(llb.Mkfile(filepath.Join(dir, "source/format"), 0o440, []byte("3.0 (quilt)")), opts...).
+		File(llb.Mkfile(filepath.Join(dir, "source/options"), 0o440, []byte("create-empty-orig")), opts...).
+		File(llb.Mkdir(filepath.Join(dir, "dalec"), 0o755), opts...).
+		File(llb.Mkfile(filepath.Join(dir, "source/include-binaries"), 0o440, append([]byte("dalec"), '\n')), opts...)
+
+	states := []llb.State{control, rules, changelog, debian}
+	states = append(states, installers...)
+
+	dalecDir := base.
+		File(llb.Mkdir(filepath.Join(dir, "dalec"), 0o755), opts...)
+
+	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/build.sh"), 0o700, createBuildScript(spec)), opts...))
+	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/patch.sh"), 0o700, createPatchScript(spec)), opts...))
+	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/fix_file_backed_sources.sh"), 0o700, fixupFileBackedSourcesScript(spec)), opts...))
+	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/fix_perms.sh"), 0o700, fixupArtifactPerms(spec)), opts...))
+
+	patchDir := dalecDir.File(llb.Mkdir(filepath.Join(dir, "dalec/patches"), 0o755), opts...)
+	sorted := dalec.SortMapKeys(spec.Patches)
+	for _, name := range sorted {
+		pls, err := sourcePatchesDir(sOpt, patchDir, filepath.Join(dir, "dalec/patches"), name, spec, opts...)
+		if err != nil {
+			return llb.Scratch(), errors.Wrapf(err, "error creating patch directory for source %q", name)
+		}
+		states = append(states, pls...)
+	}
+
+	return dalec.MergeAtPath(in, states, "/"), nil
+}
+
+func fixupArtifactPerms(spec *dalec.Spec) []byte {
+	buf := bytes.NewBuffer(nil)
+
+	fmt.Fprintln(buf, "#!/usr/bin/env bash")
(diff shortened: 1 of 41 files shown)


frontend/deb/handle_debroot.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return llb.Scratch(), errors.Wrap(err, "error generating changelog file")
+	}
+
+	if dir == "" {
+		dir = "debian"
+	}
+	compat := llb.Scratch().
+		File(llb.Mkdir(dir, 0o755, llb.WithParents(true))).
+		File(llb.Mkfile(filepath.Join(dir, "compat"), 0o770, []byte("10")))
```
pmengelbert, line 84: I see this is [recommended to be hardcoded](https://www.debian.org/doc/manuals/maint-guide/dother.en.html#compat). How about a `const`?

pmengelbert: for `10` and `compat` both, I guess

frontend/deb/handle_debroot.go
```diff
@@ -0,0 +1,148 @@
...
+
+	return dalec.MergeAtPath(in, states, "/"), nil
+}
+
+func createBuildScript(spec *dalec.Spec) llb.State {
+	buf := bytes.NewBuffer(nil)
+
+	fmt.Fprintln(buf, "#!/usr/bin/env sh")
+	fmt.Fprintln(buf, "ls -lh")
+	fmt.Fprintln(buf, "echo $(pwd)")
```
pmengelbert, line 100: debug only?

frontend/deb/handle_debroot.go
```diff
@@ -0,0 +1,148 @@
...
+			return nil, nil, err
+		}
+
+		sm, err := dalec.Sources(spec, sOpt)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		for k, src := range sm {
+			sm[k] = llb.Scratch().File(llb.Copy(src, "/", k, dalec.WithCreateDestPath()))
```
pmengelbert, line 35: Why are we doing copies here and not mounts?

cpuguy83: Because we aren't running anything just pulling out the content for the debroot that will be exported to the client.

source.go
```diff
@@ -408,3 +411,64 @@ func PatchSources(worker llb.State, spec *Spec, sourceToState map[string]llb.Sta
...
+}
+
+// SourceModifier is a function that modifies the provided source LLB state.
+type SourceModifier func(key string, src *Source, st llb.State, opts ...llb.ConstraintsOpt) (llb.State, error)
+
+// Sources gets all the source LLB states from the spec.
+//
+// SourceModifiers can be provided to modify the source LLB state before it is returned.
+// The `key` argument to the SourceModifier is the name of the source.
+func Sources(spec *Spec, sOpt SourceOpts, mods ...SourceModifier) (map[string]llb.State, error) {
```
adamperlin, line 454: I like this a lot. It is very clean with the passed in modifiers, and the returned map prevents us from having to construct the name -> source state mapping later on

frontend/jammy/handle_container.go
```diff
@@ -0,0 +1,113 @@
...
+		repo := base.
+			Run(
+				shArgs("apt-ftparchive packages . | gzip -1 > Packages.gz"),
+				llb.Dir("/tmp/deb"),
+			).
+			AddMount("/tmp/deb", deb)
+
+		sources := base.
+			Run(
+				shArgs("set -e; echo 'deb [trusted=yes] file:/tmp/_dalec_apt_repo_deb/ /' > /tmp/sources/list; cat /etc/apt/sources.list >> /tmp/sources/list"),
```
pmengelbert, line 90:
```suggestion
				shArgs("set -e; echo 'deb [trusted=yes] copy:/tmp/_dalec_apt_repo_deb/ /' > /tmp/sources/list; cat /etc/apt/sources.list >> /tmp/sources/list"),
```
From the `souces.list` manpage:

"The copy scheme is identical to the file scheme except that packages are copied into the cache directory instead of used directly at their location."

reason: `mmdebstrap` regarded the file as "not downloaded" because it wasn't in `<root>/var/cache/apt/archives`, causing the `mmdebstrap` process to exit nonzero.

pmengelbert (commented):
The `docs/examples/go-md2man-2.yml` example doesn't work, due to 
- the `GOROOT` issue that's also coming up in moby-runc also
- `revision` is empty

pmengelbert:
The CI is failing because the `test/fixtures/frontend.yml` fixture tries to build the frontend within the mariner toolkit image. The latest toolkit image has go 1.20.10 as its latest version, but the frontend requires go 1.21, since it uses the `slices` library. As a temporary workaround, we can modify the frontend.yml fixture to look like this:

```yaml
# <. . .>
sources:
  # <. . .>
  goroot:
    path: /usr/local/go
    image:
      ref: mcr.microsoft.com/oss/go/microsoft/golang:1.21

build:
  env:
    GOPROXY: direct
    CGO_ENABLED: "0"
    GOGC: off
    GOFLAGS: -trimpath
    GOPATH: /go
  steps:
    - command: |
        export GOMODCACHE="$(pwd)/gomodcache"
        export GOROOT="$(pwd)/goroot"
        export PATH="${GOROOT}/bin:${PATH}"
        cd src
        go build -o ../frontend ./cmd/frontend
# <. . .>
```

frontend/deb/handle_debroot.go
```diff
@@ -0,0 +1,148 @@
...
+		fmt.Fprintln(buf, "(")
+
+		for k, v := range step.Env {
+			fmt.Fprintf(buf, "export %s=%s\n", k, v)
+		}
+
+		fmt.Fprintln(buf, step.Command)
+		fmt.Fprintln(buf, ")")
+
+		if i < len(spec.Build.Steps)-1 {
```
adamperlin, line 112: Why are we joining build step commands with `&&` here, but not when we generate the build section of an RPM SPEC?

cpuguy83: An oversight in the rpm impl, I think.

pmengelbert: We're joining here with an unescaped newline and then &&, which causes failure when there are multiple steps. It ends up looking like this which is invalid shell syntax:
```
)
&& \
(
```
This mean the above Fprintln will have to change to Fprintf so as to avoid the newline.

pmengelbert:
- [ ] We will want to add validation for the metadata fields, similar to the mariner2 target (see #161 )

cpuguy83:
This should be ready for review and is all green @adamperlin @pmengelbert

frontend/deb/debroot.go
```diff
@@ -0,0 +1,226 @@
...
+
+	"github.com/Azure/dalec"
+	"github.com/moby/buildkit/client/llb"
+	"github.com/pkg/errors"
+)
+
+//go:embed templates/patch-header.txt
+var patchHeader []byte
+
+func sourcePatchesDir(sOpt dalec.SourceOpts, base llb.State, dir, name string, spec *dalec.Spec, opts ...llb.ConstraintsOpt) ([]llb.State, error) {
```
adamperlin, line 18: maybe a comment here explaining what this function does?

frontend/deb/debroot.go
```diff
@@ -0,0 +1,226 @@
...
+		if err != nil {
+			return llb.Scratch(), errors.Wrap(err, "error creating patch directory")
+		}
+		states = append(states, pls...)
+	}
+
+	return dalec.MergeAtPath(in, states, "/"), nil
+}
+
+func fixupFileBackedSourcesScript(spec *dalec.Spec) []byte {
```
adamperlin, line 102: maybe a comment explaining why this fixup is necessary?

frontend/deb/pkg.go
```diff
@@ -0,0 +1,128 @@
...
+			Run(
+				dalec.ShArgs("set -e; git init .; git add .; git commit -m 'Initial commit'; \"${DEBIAN_DIR}/dalec/patch.sh\"; git add .; git commit -m 'With patch'; git diff HEAD~1 >> /work/out/dalec-changes.patch; echo 'dalec-changes.patch' > /work/out/series"),
+				llb.Dir("/work/sources"),
+				mountSources(sources, "/work/sources", nil),
+				// DEBIAN_DIR is used by the patch script to find the debian directory where we actually have the patches
+				llb.AddEnv("DEBIAN_DIR", "/work/debian"),
+				llb.AddMount("/work/debian", dr, llb.SourcePath("debian"), llb.Readonly),
+				dalec.WithConstraints(opts...),
+			).AddMount("/work/out", patchesMountInput)
+	}
```
adamperlin, line 71: I think this is substantial and important enough to be moved into a separate function

frontend/deb/template_changelog.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// Get the most recent changelog entry only
+	// TODO: this could be better... but it's not worth the effort right now
+	// Where "better" means list all changelog entries
+	// Where "not worth the effort" means the changelog entry type needs updating to include extra details like the version a change was made in
+	// Really the changelog model in debian does not line up well with the dalec model because it also wants distribution information as well.
+
+	// Sort so newest entry (by change date) is on top.
+	slices.SortFunc(entries, func(i, j dalec.ChangelogEntry) int {
+		if i.Date.Equal(j.Date) {
```
adamperlin, line 63: I think you can just do `return i.Date.Compare(j.Date)` here

frontend/deb/template_changelog.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		if i.Date.Before(j.Date) {
+			return 1
+		}
+		return -1
+	})
+
+	var entry dalec.ChangelogEntry
+	if len(entries) == 0 {
+		entry = dummyChangelogEntry
```
adamperlin, line 74: Do we always need at least one changelog entry?

frontend/deb/template_changelog.go
```diff
@@ -0,0 +1,94 @@
...
+		entry = entries[0]
+	}
+
+	buf := &strings.Builder{}
+
+	for _, change := range entry.Changes {
+		fmt.Fprintln(buf, "  *", change)
+	}
+
+	fmt.Fprintf(buf, " -- %s  %s", entry.Author, entry.Date.Format("Mon, 02 Jan 2006 15:04:05 -0700"))
```
adamperlin, line 85: Where is this date coming from?

frontend/rpm/template.go
```diff
@@ -109,23 +113,36 @@ func (w *specWrapper) Requires() fmt.Stringer {
...
-	if len(constraints) == 0 {
-		fmt.Fprintf(b, "%s: %s\n", kind, name)
+func writeDep(b *strings.Builder, kind, name string, constraints dalec.PackageConstraints) {
+	do := func() {
+		if len(constraints.Version) == 0 {
+			fmt.Fprintf(b, "%s: %s\n", kind, name)
+			return
+		}
+
+		sort.Strings(constraints.Version)
```
adamperlin, line 131: I believe this will mutate the underlying `constraints.Version` array, is that ok?

frontend/deb/debroot.go
adamperlin: ah, I see you added a comment with the script, maybe add a brief one here too?

frontend/deb/template_changelog.go
cpuguy83: Yes.

frontend/deb/template_changelog.go
cpuguy83: These dates are how go parses time: https://pkg.go.dev/time#example-Time.Format

```

	// The rest of this function demonstrates the properties of the
	// layout string used in the format.

	// The layout string used by the Parse function and Format method
	// shows by example how the reference time should be represented.
	// We stress that one must show how the reference time is formatted,
	// not a time of the user's choosing. Thus each layout string is a
	// representation of the time stamp,
	//	Jan 2 15:04:05 2006 MST
	// An easy way to remember this value is that it holds, when presented
	// in this order, the values (lined up with the elements above):
	//	  1 2  3  4  5    6  -7
```

frontend/deb/debroot.go
```diff
@@ -0,0 +1,226 @@
...
+	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/build.sh"), 0o700, createBuildScript(spec)), opts...))
+	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/patch.sh"), 0o700, createPatchScript(spec)), opts...))
+	states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/fix_file_backed_sources.sh"), 0o700, fixupFileBackedSourcesScript(spec)), opts...))
+
+	patchDir := dalecDir.File(llb.Mkdir(filepath.Join(dir, "dalec/patches"), 0o755), opts...)
+	sorted := dalec.SortMapKeys(spec.Patches)
+	for _, name := range sorted {
+		pls, err := sourcePatchesDir(sOpt, patchDir, filepath.Join(dir, "dalec/patches"), name, spec, opts...)
+		if err != nil {
+			return llb.Scratch(), errors.Wrap(err, "error creating patch directory")
```
DannyBrito, line 94: maybe, the error message could be more specific, like adding on which source the patch creation. failed at?

frontend/deb/debroot.go
```diff
@@ -0,0 +1,226 @@
...
+		if _, ok := patches[name]; ok {
+			continue
+		}
+		if dalec.SourceIsDir(src) {
+			continue
+		}
+
+		fmt.Fprintf(buf, "mv %q/%q %q.dalec.tmp\n", name, name, name)
+		fmt.Fprintln(buf, "rm -rf", name)
+		fmt.Fprintf(buf, "mv %q.dalec.tmp %q\n", name, name)
```
DannyBrito, line 126: 1. the not sure how the `%q/%q` path works combined, but just wanted to make sure that it handles spaces so maybe we should quote it?
```suggestion
		fmt.Fprintf(buf, "mv \"%q/%q\" \"%q.dalec.tmp\"\n", name, name, name)
		fmt.Fprintln(buf, "rm -rf", name)
		fmt.Fprintf(buf, "mv \"%q.dalec.tmp\" %q\n", name, name)
```
2. for the rm -rf does it need a newline?
```suggestion
		fmt.Fprintf(buf, "mv %q/%q %q.dalec.tmp\n", name, name, name)
		fmt.Fprintln(buf, "rm -rf %q\n", name)
		fmt.Fprintf(buf, "mv %q.dalec.tmp %q\n", name, name)
```

frontend/deb/debroot.go
```diff
@@ -0,0 +1,226 @@
...
+	for k, v := range spec.Build.Env {
+		fmt.Fprintln(buf, "export", k+"="+v)
+	}
+
+	for _, step := range spec.Build.Steps {
+		fmt.Fprintln(buf)
+		fmt.Fprintln(buf, "(")
+
+		for k, v := range step.Env {
+			fmt.Fprintf(buf, "	export %s=%s\n", k, v)
```
DannyBrito, line 169: Handle spaces in the value.
```suggestion
			fmt.Fprintf(buf, "	export %s=\"%s\"\n", k, v)
```
or it could be handle with %q not super sure on how it behaves
```suggestion
			fmt.Fprintf(buf, "	export %s=%q\n", k, v)
```

frontend/deb/debroot.go
```diff
@@ -0,0 +1,226 @@
...
+func createBuildScript(spec *dalec.Spec) []byte {
+	buf := bytes.NewBuffer(nil)
+
+	fmt.Fprintln(buf, "#!/usr/bin/env sh")
+	fmt.Fprintln(buf)
+
+	fmt.Fprintln(buf, "set -ex")
+
+	for k, v := range spec.Build.Env {
+		fmt.Fprintln(buf, "export", k+"="+v)
```
DannyBrito, line 161: a bit of consistency and handle spaces in value
```suggestion
		fmt.Fprintf(buf, "export %s=\"%s\"\n", k, v)
```

frontend/deb/debroot.go
```diff
@@ -0,0 +1,226 @@
...
+	patches := map[string]struct{}{}
+	for _, patchList := range spec.Patches {
+		for _, patch := range patchList {
+			patches[patch.Source] = struct{}{}
+		}
+	}
+
+	buf := bytes.NewBuffer(nil)
+	buf.WriteString("#!/usr/bin/env sh\n\n")
+	buf.WriteString("set -ex\n\n")
```
DannyBrito, line 113: nit: can the headers maybe of the scrips be consistent between fixupBackedSourceScript, createBuildScript and createPatchScript? It seems it have a space on it?
```go
	fmt.Fprintln(buf, "#!/usr/bin/env sh")
	fmt.Fprintln(buf)
```

frontend/deb/template_rules.go
```diff
@@ -0,0 +1,56 @@
...
+
+type rulesWrapper struct {
+	*dalec.Spec
+}
+
+func (w *rulesWrapper) Envs() fmt.Stringer {
+	b := &strings.Builder{}
+
+	for k, v := range w.Spec.Build.Env {
+		fmt.Fprintf(b, "export %s := %s\n", k, v)
```
DannyBrito, line 45: handle value with spaces
```suggestion
		fmt.Fprintf(b, "export %s := \"%s\"\n", k, v)
```

spec.go
```diff
@@ -276,17 +276,22 @@ type Source struct {
 	Excludes []string `yaml:"excludes,omitempty" json:"excludes,omitempty"`
 }
 
+type PackageConstraints struct {
+	Version []string `yaml:"version,omitempty" json:"version,omitempty"`
```
DannyBrito, line 280: general question, here version been an array, does it make Dalec able to handle install of multiple versions of same package?

test/fixtures/moby-runc.yml
```diff
@@ -36,8 +51,12 @@ targets: # Distro specific build requirements
       runtime:
         /bin/sh:
         libseccomp:
-          - ">= 2.3"
+          version:
+            - ">= 2.3"
     tests:
```
DannyBrito, line 56: should `jammy` target also have some tests?

test/mariner2_test.go
```diff
@@ -24,7 +23,7 @@ func testLinuxDistro(ctx context.Context, t *testing.T, buildTarget string) {
 		t.Parallel()
 		spec := dalec.Spec{
 			Name:        "test-build-commands-fail",
-			Version:     "v0.0.1",
+			Version:     "0.0.1",
```
DannyBrito, line 26: why is the drop of the `v` prefix?

cpuguy83: It's not technically correct.
Also, IIRC, even though rpmbuild doesn't seem to care, debbuild didn't like it.

frontend/deb/template_changelog.go
adamperlin: Ah got it, thank you for clarifying. I'd never run into that before

frontend/deb/template_rules.go
cpuguy83: This is `make` which would include literal quotes, which we don't want here.

frontend/rpm/template.go
cpuguy83: In the end I'm not sure why I was sorting this.
I just removed it.

frontend/deb/template_changelog.go
cpuguy83: Fixed. Nice catch.

frontend/deb/debroot.go
cpuguy83: Done.

frontend/deb/debroot.go
cpuguy83: Done.

frontend/deb/debroot.go
cpuguy83: DOne.

frontend/deb/debroot.go
cpuguy83: Done.

test/fixtures/moby-runc.yml
cpuguy83: I moved the manpage test from mariner into the main tests.
I can't think of any jammy specific ones to add here.

frontend/deb/debroot.go
cpuguy83: Done.

frontend/deb/debroot.go
cpuguy83: Done.

frontend/deb/pkg.go
```diff
@@ -0,0 +1,132 @@
...
+func BuildDeb(worker llb.State, spec *dalec.Spec, sOpt dalec.SourceOpts, targetKey string, opts ...llb.ConstraintsOpt) (llb.State, error) {
+	srcPkg, err := SourcePackage(sOpt, worker, spec, targetKey)
+	if err != nil {
+		return llb.Scratch(), errors.Wrap(err, "error creating debian source package")
+	}
+
+	dirName := filepath.Join("/work", spec.Name+"_"+spec.Version+"-"+spec.Revision)
+	st := worker.
+		Run(
+			dalec.ShArgs("set -e; ls -lh; dpkg-source -x ./*.dsc; ls -lh; cd "+spec.Name+"-"+spec.Version+"; ls -lh *; dpkg-buildpackage -b -uc -us; mkdir -p /tmp/out; cp ../*.deb /tmp/out; ls -lh /tmp/out"),
```
adamperlin, line 107: is the `ls -lh` for debugging?

spec.go
adamperlin: Can we add a comment noting that these are version constraints?

frontend/deb/pkg.go
cpuguy83: Yeah, when stuff goes south its nice to see the content there.

spec.go
cpuguy83: This is done.

frontend/rpm/template.go
```diff
@@ -112,23 +115,35 @@ func (w *specWrapper) Requires() fmt.Stringer {
...
+		do()
 		return
 	}
 
-	sort.Strings(constraints)
-	for _, c := range constraints {
-		fmt.Fprintf(b, "%s: %s %s\n", kind, name, c)
+	for _, arch := range constraints.Arch {
+		fmt.Fprintf(b, "%%ifarch %s\n", arch)
+		slices.Sort(constraints.Arch)
```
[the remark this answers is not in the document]
cpuguy83: Oh boy, good catch.
Because I wrote wildly incorrect code here.

frontend/jammy/handle_container.go
```diff
@@ -0,0 +1,197 @@
...
+		Run(
+			dalec.ShArgs("apt-get update && mmdebstrap --variant=custom --mode=chrootless --include="+spec.Name+" jammy /tmp/rootfs /etc/apt/sources.list && rm -rf /tmp/rootfs/var/lib/apt && rm -rf /tmp/rootfs/var/cache/apt"),
+			llb.AddEnv("DEBIAN_FRONTEND", "noninteractive"),
+			repoMount,
+			dalec.WithConstraints(opts...),
+		).
+		AddMount("/tmp/rootfs", llb.Scratch())
+}
+
+func addAptRepoForDeb(base llb.State, deb llb.State, opts ...llb.ConstraintsOpt) llb.RunOption {
```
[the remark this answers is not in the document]
cpuguy83: Added a doc string for this.

frontend/jammy/handle_container.go
```diff
@@ -0,0 +1,197 @@
...
+		deps := spec.GetTestDeps(targetKey)
+		if len(deps) == 0 {
+			return in
+		}
+
+		opts = append(opts, dalec.ProgressGroup("Install test dependencies"))
+
+		cleanupVars := dalec.GetOutputBaseImageRef(spec, targetKey) == ""
+
+		if cleanupVars {
```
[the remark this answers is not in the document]
cpuguy83: Added some comments around this.

frontend/rpm/template.go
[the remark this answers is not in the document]
cpuguy83: Fixed this.

cpuguy83:
As an update here, I've been carrying this through adding core spec features.
Right now I"m working on getting systemd units support and have hit a number of issues with dh_installsystemd:

1. It is not flexible enough to support even our test cases because you cannot differentiate between a "foo.service" and "foo.socket" because it only deals with "foo". So if you need to enable one you need to enable them all.
2. mmdebdstrap's only means of doing truly unprivileged bootstrapping, "chrootless" mode, does not necessarily work with all maintainer scripts from deb packages (e.g. postinst), including even "essential" packages for debian. This is a fairly generic problem, so even if our own package do not have an issue a dependent package may. It also appears that `dh_instalsystemd` does not work with it.


For 1, I think we'll need to write our own scripts to handle this. Not a big deal just not ideal.

For 2, this is a larger problem. mmdebstrap has other modes but they all require CAP_SYS_ADMIN because it tries to mount things from the worker rootfs into a chroot.
Likely we'll need to tease apart what mmdebstrap is doing to make apt/dpkg work in this case and handle privileged operations in LLB.
Unfortunately mmdebstrap does not like it if I pre-configure the mounts for it as it wants a completely empty root.

#### About dh_installsystemd
This is a debhelper script that is responsible for setting up postinst/postrm/etc scripts for managing the unit on install/upgrade/remove/etc.

cpuguy83:
Ok, this is ready.
A couple of things to know:

1. This does not build a distroless jammy image though we may in the future, its not a goal at the moment and there are some issues in making that happen.
2. There are some edge cases around artifact installation in into the deb package, particularly around handling wildcards. The tests we have are passing, though and I do not expect these situations to be normal and could be done differently in the spec file.

frontend/jammy/handle_container.go
```diff
@@ -0,0 +1,189 @@
...
+		workerFS, err := bkfs.FromState(ctx, &worker, client)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// Check if there there is a test repo in the worker image.
+		// We'll mount that into the target container while installing packages.
+		_, repoErr := fs.Stat(workerFS, testRepoPath[1:])
+		_, listErr := fs.Stat(workerFS, testRepoSourceListPath[1:])
+		if listErr == nil && repoErr == nil {
```
adamperlin, line 56: Can this manual test repo handling go away once we have support for custom repos in the spec?

frontend/jammy/handle_container.go
```diff
@@ -0,0 +1,189 @@
...
+			// installation of a lot of thigns, including doc files.
+			// This is mounting over that file with an empty file so that our test suite
+			// passes (as it is looking at these files).
+			llb.AddMount("/etc/dpkg/dpkg.cfg.d/excludes", tmp, llb.SourcePath("tmp")).SetRunOption(cfg)
+		}),
+	).Root().
+		With(installSymlinks)
+}
+
+func installTestDeps(worker llb.State, spec *dalec.Spec, targetKey string, opts ...llb.ConstraintsOpt) llb.StateOption {
```
adamperlin, line 174: The `worker` parameter appears to be unused here

frontend/deb/debroot.go
```diff
@@ -0,0 +1,438 @@
...
+	installers := createInstallScripts(worker, spec, dir)
+
+	debian := base.
+		File(llb.Mkdir(filepath.Join(dir, "source"), 0o755), opts...).
+		With(func(in llb.State) llb.State {
+			if len(spec.Sources) == 0 {
+				return in
+			}
+			return in.
+				File(llb.Mkfile(filepath.Join(dir, "source/format"), 0o640, []byte("3.0 (quilt)")), opts...).
```
adamperlin, line 90: Perhaps make the contents of the `source/format` and `source/options` files constants?

frontend/deb/debroot.go
```diff
@@ -0,0 +1,438 @@
...
+	base := llb.Scratch().File(llb.Mkdir(dir, 0o755, llb.WithParents(true)))
+
+	installBuf := bytes.NewBuffer(nil)
+	writeInstallHeader := sync.OnceFunc(func() {
+		fmt.Fprintln(installBuf, "#!/usr/bin/dh-exec")
+		fmt.Fprintln(installBuf)
+	})
+
+	writeInstall := func(src, dir, name string) {
+		writeInstallHeader()
```
adamperlin, line 259: Can you add an extra comment here that `writeInstallHeader` is in a `OnceFunc` so will only be called once?

frontend/deb/template_rules.go
```diff
@@ -0,0 +1,160 @@
...
+		enable := grouping[firstKey].Enable
+
+		b.WriteString("\tdh_installsystemd --name=" + basename)
+		if !enable || needsCustomEnable {
+			b.WriteString(" --no-enable")
+		}
+		b.WriteString("\n")
+	}
+
+	if includeCustomEnable {
```
adamperlin, line 153: Can this just be `if needsCustomEnable { ... }`? Do we need the `includeCustomEnable` flag var too?

frontend/deb/templates/custom_enable_postinst.tmpl
```diff
@@ -0,0 +1,17 @@
+# Automatically added by github.com/Azure/dalec
+# This is generally copied from dh_installsystemd
+# This is required because dh_installsystemd operates on the unit base name and
+# not the full unit name (e.g. "foo" instead of "foo.service") and sometimes
+# we want to enable one and not the other.
+if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ] ; then
+	# This will only remove masks created by d-s-h on package removal.
+	deb-systemd-helper unmask '{{.}}' >/dev/null || true
+
+	if deb-systemd-helper --quiet was-enabled '{{.}}'; then
```
adamperlin, line 10: How does this work if this is the first time the unit is being installed? Will the unit still be enabled?

frontend/deb/debroot.go
```diff
@@ -0,0 +1,438 @@
...
+		// We need to track this in cases where some units under a base are
+		// enabled and some are not since dh_installsystemd does not support this
+		// directly.
+
+		sorted := dalec.SortMapKeys(units)
+		for _, key := range sorted {
+			cfg := units[key]
+			name, suffix := cfg.SplitName(key)
+			if name != spec.Name {
+				name = spec.Name + "." + name
```
adamperlin, line 375: Can you explain why we are preprending the package name to the unit name here?

frontend/deb/debroot.go
```diff
@@ -0,0 +1,438 @@
...
+		}
+
+		fmt.Fprintln(installBuf, src, dir+"/")
+	}
+
+	if len(spec.Artifacts.Binaries) > 0 {
+		sorted := dalec.SortMapKeys(spec.Artifacts.Binaries)
+		for _, key := range sorted {
+			cfg := spec.Artifacts.Binaries[key]
+			writeInstall(key, filepath.Join("/usr/bin", cfg.SubPath), cfg.ResolveName(key))
```
adamperlin, line 273: Should `/usr/bin` be hardcoded here?

frontend/deb/debroot.go
```diff
@@ -0,0 +1,438 @@
...
+		buf := bytes.NewBuffer(nil)
+
+		sorted := dalec.SortMapKeys(spec.Artifacts.Directories.Config)
+		for _, name := range sorted {
+			fmt.Fprintln(buf, filepath.Join("/etc", name))
+		}
+
+		sorted = dalec.SortMapKeys(spec.Artifacts.Directories.State)
+		for _, name := range sorted {
+			fmt.Fprintln(buf, filepath.Join("/var/lib", name))
```
adamperlin, line 322: Should these paths be hardcoded? Is `/var/lib` the guaranteed convention?

frontend/jammy/handle_deb.go
```diff
@@ -0,0 +1,163 @@
...
+			Runtime: buildDeps,
+		},
+		Description: "Build dependencies for " + spec.Name,
+	}
+
+	pg := dalec.ProgressGroup("Install build dependencies")
+	opts = append(opts, pg)
+	deb, err := deb.BuildDeb(worker, depsSpec, sOpt, targetKey, append(opts, dalec.ProgressGroup("Create intermediate deb for build dependnencies"))...)
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating intermediate package for insalling build dependencies")
```
adamperlin, line 152:
```suggestion
		return nil, errors.Wrap(err, "error creating intermediate package for installing build dependencies")
```

frontend/deb/template_control.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (w *controlWrapper) Architecture() string {
+	if w.NoArch {
+		return "all"
+	}
+	return "linux-any"
+}
+
+func injectConstraints(in []string, deps map[string]dalec.PackageConstraints) {
```
adamperlin, line 31: Can you change this function to return a new slice with the constraints attached to each package instead of mutating the passed in one? That reads more clearly to me. If so, it could be renamed to something like `appendConstraints`?

frontend/deb/debroot.go
cpuguy83: Do you have something else in mind? I don't think there's any macro or anything to go with here.

frontend/deb/debroot.go
cpuguy83: Same as https://github.com/Azure/dalec/pull/87#discussion_r1725751256

frontend/deb/templates/custom_enable_postinst.tmpl
cpuguy83: This is all verbatim from the deb-systemd-helper scripts.
Running this command locally on an non-existant package:

```console
$ DPKG_MAINTSCRIPT_PACKAGE=1 deb-systemd-helper was-enabled totally-does-not-exist; echo $?                                                                                                                                          ─╯
enabled
0
```

cpuguy83: Reading the code for all this (and enabling debug on the script) it is reading some state in `/var/lib/systemd/deb-systemd-helper-enabled/`.
This directory has files for each unit that d-s-h manages.
Each file contains 1 or more references to systemd units linked into whatever target its being installed to (e.g. multi-user.target).
The `was-enabled` subcommand checks if all the listed links are present (and are symlinks).
If there is no file (or the file is not a symlink) it says yes it is enabled.

frontend/jammy/handle_container.go
cpuguy83: Yeah this is janky AF. 😢 
I was thinking we could just copy all extra repos, but this isn't necessarily always desirable.
We could override the base image but then we need to break out the test to be something jammy specific which isn't great either.

adamperlin: Hmm, I see your point, and I suppose there isn't another great solution for now. Well, maybe we can at least open an issue to revisit this? I'd definitely prefer that this not be the permanent solution

frontend/deb/templates/custom_enable_postinst.tmpl
adamperlin: Ok perfect, that was exactly my question: I was concerned about what the `was-enabled` command would return in the case where d-s-h had no record of a unit. But since this is handled properly then this seems fine to me (especially since it mimics the behavior of the deb-systemd-helper scripts anyways)

frontend/deb/debroot.go
adamperlin: Yeah, I suppose I was wondering if there was a macro or some kind of debhelper command which would return configured installation paths. But it's looking like that really doesn't exist in which case I guess hardcoding according to the proper conventions may be the best thing we can do

frontend/deb/template_control.go
cpuguy83: Classic case of "wtf was I thinking when I wrote this" moment.
All calls to `injectConstraints` make the same `injectConstraints(slices.Sort(maps.Keys(deps)), deps))`

cpuguy83: Done.

frontend/jammy/handle_deb.go
cpuguy83: Fixed.

cpuguy83:
This should be ready to go now.

cpuguy83:
Extra commit LGTM

test/testenv/buildx.go
```diff
@@ -293,6 +293,8 @@ func (b *BuildxEnv) RunTest(ctx context.Context, t *testing.T, f TestFunc, opts
 	withGHCache(t, &so)
 	withResolveLocal(&so)
 
+	// so.AllowedEntitlements = append(so.AllowedEntitlements, entitlements.EntitlementSecurityInsecure)
```
adamperlin, line 296: Can we delete this?

test/testenv/buildx.go
```diff
@@ -293,6 +293,8 @@ func (b *BuildxEnv) RunTest(ctx context.Context, t *testing.T, f TestFunc, opts
...
+
```
cpuguy83, line 297:
```suggestion
```

test/testenv/buildx.go
cpuguy83: Done.

frontend/deb/template_control.go
```diff
@@ -0,0 +1,153 @@
...
+func appendConstraints(deps map[string]dalec.PackageConstraints) []string {
+	if deps == nil {
+		return nil
+	}
+	out := dalec.SortMapKeys(deps)
+
+	for i, dep := range out {
+		constraints := deps[dep]
+		s := dep
+		if len(constraints.Version) > 0 {
```
adamperlin, line 42: Can we add a comment here that the format is specified here: https://www.debian.org/doc/debian-policy/ch-relationships.html and is `package (version constraint, ...) [arch constraint, ...]`

frontend/deb/pkg.go
```diff
@@ -0,0 +1,154 @@
...
+	work := worker.Run(
+		dalec.ShArgs("set -e; ls -lh; dpkg-buildpackage -S -us -uc; mkdir -p /tmp/out; ls -lh; cp -r /work/"+spec.Name+"_"+spec.Version+"* /tmp/out; ls -lh /tmp/out"),
+		llb.Dir("/work/pkg"),
+		llb.AddMount("/work/pkg/debian", dr, llb.SourcePath("debian")), // This cannot be readonly because the debian directory gets modified by dpkg-buildpackage
+		llb.AddMount("/work/pkg/debian/patches", patches, llb.Readonly),
+		llb.AddEnv("DH_VERBOSE", "1"),
+		llb.Network(pb.NetMode_NONE),
+		mountSources(sources, "/work/pkg", sanitizeSourceKey),
+		dalec.RunOptFunc(func(ei *llb.ExecInfo) {
+			for key, src := range debSources {
```
adamperlin, line 109: Can you add a comment clarifying the need for this RunOptFunc?

adamperlin (approved):
This is a very substantial PR and I tried to review it as thoroughly as I could, though I think there are definitely things I've missed. It certainly doesn't hurt to have more eyes on the PR, but at this point since all the functionality is there it seems like getting this merged and fixing anything that crops up later is maybe the best option!

frontend/deb/pkg.go
cpuguy83, line 109:
```suggestion
			// Mount all the tar+gz'd sources into the build which will get picked p by debbuild
			or key, src := range debSources
```

frontend/deb/template_control.go
cpuguy83, line 42:
```suggestion
		// Format is specified in https://www.debian.org/doc/debian-policy/ch-relationships.html#syntax-of-relationship-fields
		if len(constraints.Version) > 0 {
```

frontend/deb/pkg.go
cpuguy83: Done

frontend/deb/template_control.go
cpuguy83: Done

PR #95, opened 2024-02-02
Remove unused bkembed package

cpuguy83:
Old stuff that never got removed.


adamperlin (approved):
LGTM since this is no longer used. Remind me what this was used for originally?

cpuguy83:
It was accidentally included in a commit awhile ago via `git add .`
I originally intended to use it for testing but just wound up using a full buildkit instance.

PR #98, opened 2024-02-03
Add support for inline sources

cpuguy83:
This allows you to specify files or trees inline in yaml spec.

Example:

```yaml
sources:
  someFile:
    inline:
      file:
      contents: "hello!"
  someDir:
    inline:
      dir:
        files:
          foo:
            contents: "I am foo!"
```

You can also set permissions and ownership.

### DIFF
--- a/files.go
+++ b/files.go
@@ -92,7 +92,7 @@ func (s *SourceInlineDir) validate() error {
 
 	for k, f := range s.Files {
 		if strings.ContainsRune(k, os.PathSeparator) {
-			errs = append(errs, errors.Errorf("file name %q must not contain path separator", k))
+			errs = append(errs, errors.Wrapf(sourceNamePathSeparatorError, "file %q", k))
 		}
 		if err := f.validate(); err != nil {
 			errs = append(errs, errors.Wrapf(err, "file %q", k))
--- a/files.go
+++ b/files.go
@@ -1,0 +1,170 @@
+package dalec
+
+import (
+	goerrors "errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/moby/buildkit/client/llb"
+	"github.com/pkg/errors"
+)
+
+const (
+	defaultFilePerms = 0o644
+	defaultDirPerms  = 0o755
+)
+
+func (d *SourceInlineDir) PopulateAt(p string) llb.StateOption {
+	return func(st llb.State) llb.State {
+		perms := d.Permissions.Perm()
+		if perms == 0 {
+			perms = defaultDirPerms
+		}
+
+		st = st.File(llb.Mkdir(p, perms, llb.WithUIDGID(int(d.UID), int(d.GID))))
+
+		sorted := SortMapKeys(d.Files)
+		for _, k := range sorted {
+			f := d.Files[k]
+			st = st.With(f.PopulateAt(filepath.Join(p, k)))
+		}
+
+		return st
+	}
+}
+
+func (f *SourceInlineFile) PopulateAt(p string) llb.StateOption {
+	return func(st llb.State) llb.State {
+		perms := f.Permissions.Perm()
+		if perms == 0 {
+			perms = defaultFilePerms
+		}
+
+		return st.File(
+			llb.Mkfile(p, perms, []byte(f.Contents), llb.WithUIDGID(int(f.UID), int(f.GID))),
+		)
+	}
+}
+
+func (s *SourceInline) validate(subpath string) (retErr error) {
+	var errs []error
+
+	if s.File == nil && s.Dir == nil {
+		errs = append(errs, errors.New("inline source is missing contents to inline"))
+	}
+
+	if s.File != nil && s.Dir != nil {
+		errs = append(errs, errors.New("inline source variant cannot have both a file and dir set"))
+	}
+
+	if s.Dir != nil {
+		if err := s.Dir.validate(); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	if s.File != nil {
+		if subpath != "" {
+			errs = append(errs, errors.New("inline file source cannot have a path set"))
+		}
+		if err := s.File.validate(); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return goerrors.Join(errs...)
+}
+
+func (s *SourceInlineDir) validate() error {
+	var errs []error
+
+	if s.UID < 0 {
+		errs = append(errs, errors.Errorf("uid %d must be non-negative", s.UID))
+	}
+
+	if s.GID < 0 {
+		errs = append(errs, errors.Errorf("gid %d must be non-negative", s.GID))
+	}
+
+	for k, f := range s.Files {
+		if strings.ContainsRune(k, os.PathSeparator) {
+			errs = append(errs, errors.Errorf("file name %q must not contain path separator", k))
+		}
+		if err := f.validate(); err != nil {
+			errs = append(errs, errors.Wrapf(err, "file %q", k))
+		}
+	}
+	return goerrors.Join(errs...)
+}
+
+func (s *SourceInlineFile) validate() error {
+	var errs []error
+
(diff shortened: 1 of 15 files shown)


spec.go
```diff
@@ -225,6 +225,60 @@ type SourceBuild struct {
...
+	// Permissions is the octal permissions to set on the directory.
+	Permissions fs.FileMode `yaml:"permissions,omitempty" json:"permissions,omitempty"`
+
+	// UID is the user ID to set on the directory and all files and directories within it.
+	// UID must be greater than 0
+	UID int `yaml:"uid,omitempty" json:"uid,omitempty"`
+	// GID is the group ID to set on the directory and all files and directories within it.
+	// GID must be greater than 0
+	GID int `yaml:"gid,omitempty" json:"gid,omitempty"`
+}
```
pmengelbert, line 267: Inline files make a lot of sense for one-off things that don't deserve their own file, but I don't know if I would want to allow directories. It seems unnecessary to allow complex nested inlining when the filesystem is right there and available as a `SourceTypeContext`.

I think it could get quickly get very indented and cluttered. This functionality is already available by other means, for example a `Build` source type with a local dir provided as a context and a small inline dockerfile to set the perms and uid/gid of the files/directories, which can be more concise due to `(chown|chmod) -R`.

cpuguy83: Context is not really the same thing.
Context is client-side, different clients have different contexts.

Using `Build` is also significantly more complex (on the backend) and requires a remote image to do anything at all.

I thought I put this in the commit but apparently I did not: The main intent for these is to simplify testing and not require tests to hit registries.

pmengelbert: > Context is not really the same thing. Context is client-side, different clients have different contexts.
> 
Understood

> Using `Build` is also significantly more complex (on the backend) and requires a remote image to do anything at all.
> 
Make sense

> I thought I put this in the commit but apparently I did not: The main intent for these is to simplify testing and not require tests to hit registries.
Sounds good. Was wondering what the reason was for this.

files.go
```diff
@@ -0,0 +1,184 @@
...
+
+		return st
+	}
+}
+
+func (f *SourceInlineFile) PopulateAt(p string) llb.StateOption {
+	return func(st llb.State) llb.State {
+		perms := f.Permissions.Perm()
+		if perms == 0 {
+			perms = 0o644
```
adamperlin, line 45: Maybe add this as a const?

load.go
```diff
@@ -365,14 +371,17 @@ func (s *Spec) FillDefaults() {
 
 func (s Spec) Validate() error {
 	for name, src := range s.Sources {
+		if strings.ContainsRune(name, os.PathSeparator) {
```
adamperlin, line 374: On this note, have we specifically defined what are and are not valid source names?

spec.go
```diff
@@ -225,6 +225,60 @@ type SourceBuild struct {
...
 }
 
+// SourceInlineContent is used to specify the content of an inline source.
+type SourceInlineFile struct {
+	// Content is the contents.
+	Contents string `yaml:"content,omitempty" json:"content,omitempty"`
+	// Permissions is the octal file permissions to set on the file.
+	Permissions fs.FileMode `yaml:"permissions,omitempty" json:"permissions,omitempty"`
+	// UID is the user ID to set on the directory and all files and directories within it.
+	// UID must be greater than 0
```
adamperlin, line 235: It looks like in the validation code we're just testing `< 0`, could we clarify this comment vs. the validation logic?

spec.go
```diff
@@ -225,6 +225,60 @@ type SourceBuild struct {
...
+	// Directories with path separators in the key will be rejected.
+	//
+	// Directories are created before files are created.
+	Dirs map[string]*SourceInlineDir `yaml:"dirs,omitempty" json:"dirs,omitempty"`
+	// Permissions is the octal permissions to set on the directory.
+	Permissions fs.FileMode `yaml:"permissions,omitempty" json:"permissions,omitempty"`
+
+	// UID is the user ID to set on the directory and all files and directories within it.
+	// UID must be greater than 0
+	UID int `yaml:"uid,omitempty" json:"uid,omitempty"`
```
adamperlin, line 263: Same as above, just making sure comment matches validation

cpuguy83: Nice catch. I think this was copilot :)
It tries hard.

load.go
cpuguy83: We have not, but should indeed.

cpuguy83:
Also realized we can drop the "Inline" field in the SourceBuild struct.

spec.go
pmengelbert: > Context is not really the same thing. Context is client-side, different clients have different contexts.
> 

So actually I'm not sure I fully understand this. You mean that a context can be all kinds of things, depending on the client? For example, with `buildctl` I can do this:

```bash
buildctl build \                                                                                       --frontend=dockerfile.v0 \                                                                     --local context=. \                                                                            --oci-layout=foo=./oci \                                                                       --opt context:bar=oci-layout:foo@sha256:760c7b6<...> \                                         --local dockerfile=. \
        --output=type=local,dest=../hello

# Dockerfile
FROM bar
```

An in this case, the directory `oci` is not actually the context (because it's json blobs / tarballs of the oci layout dir), the actual context is the fs of the image *represented* by the `oci` layout dir.

Is that right?

cpuguy83: I mean a context is dependent on the build invocation.
The client tells you what the context is and is not declarative.

spec.go
cpuguy83: Fixed

files.go
cpuguy83: Done

files.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (s *SourceInline) Doc(w io.Writer, name string) {
+	if s.File != nil {
+		s.File.Doc(w, name)
+	}
+
+	if s.Dir != nil {
+		s.Dir.Doc(w, name)
+	}
+
```
adamperlin, line 139: nit: extra newline

frontend/debug/handle_sources.go
```diff
@@ -19,14 +19,15 @@ func HandleSources(ctx context.Context, gwc client.Client, spec *dalec.Spec) (cl
...
+		name := name
+		src := src
 		f := dalec.Source2LLBGetter(spec, src, name)
 		st, err := f(sOpt)
 		if err != nil {
 			return nil, nil, err
 		}
 
-		to := llb.Scratch().File(llb.Mkdir("/"+name, 0o755, llb.WithParents(true)))
-		sources = append(sources, to.File(llb.Copy(st, "/", "/"+name, dalec.WithDirContentsOnly())))
```
adamperlin, line 29: Was this copy operation not necessary?

cpuguy83: Yes, it was adding an extra nesting unexpectedly.

cpuguy83:
I'm going to move the new test to a unit test.

cpuguy83:
I've updated this to test inline sources via unit tests.

I've also removed support for nested directories.
We can add this back easily enough in the future if needed, but after discussing with @pmengelbert last week I think it is best to keep this simple.
If multiple directories are needed someone can use multiple sources and arrange things as part of the build step.

adamperlin:
Wait what was the reasoning for removing inline directories? In my opinion natively supporting that seems simpler than trying to achieve it through multiple sources/using build context. I think nested directories are very useful for testing purposes (I can already think of cases where this would have helped me), and I personally don't feel like it adds that much complication. Yes, it would get deeply nested if you used the feature beyond its intended purpose, but for putting in one or two levels of directory nesting it doesn't seem unreasonable to me. Definitely not a blocker by any means, but just thought I'd add my opinion

cpuguy83:
This is rebased and all green.

cpuguy83:
@adamperlin There is no way to specify nested directories at all.

adamperlin:
Ah yes, I see that in the spec for `SourceInlineDir` now

adamperlin (approved):
Looks good to me!

PR #108, opened 2024-02-05
Add support for Windows (cross compiled)

pmengelbert:
Very rough, mostly looking to get feedback. To test (of course with the re-built frontend):

```
docker build -f test/fixtures/moby-containerd.yml --target=windows/zip --output=_zip .
```


cpuguy83 (commented):
For tests, I'd like to move over to using the new integration test suite.
Additionally I really don't want to be building containerd (or runc) on every test run.

This is part of why I'm pushing for inline sources (#98).
We can test that a spec does everything we expect it to without having to build a bunch of projects.

We can of course add an e2e suite (later) that double checks real projects, but probably I'd only run it either on merge or before releases.

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,168 @@
...
+
+func createBuildScript(spec *dalec.Spec) llb.State {
+	buf := bytes.NewBuffer(nil)
+
+	fmt.Fprintln(buf, "#!/usr/bin/env sh")
+
+	for i, step := range spec.Build.Steps {
+		fmt.Fprintln(buf, "(")
+
+		for k, v := range step.Env {
```
adamperlin, line 151: This same logic of entering a subshell, adding env vars with `export` before command, and writing the command, is re-used now across several places (RPM, Jammy PR, now here). I wonder if we can abstract this out into a separate function or consider how best to reduce duplicated code here

pmengelbert:
I copied/pasted this from Brian's jammy PR, we should definitely abstract 🙂
________________________________

cpuguy83:
Looks like the rebase on this was incomplete.
Code is not compiling.

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+const (
+	workerImgRef    = "mcr.microsoft.com/mirror/docker/library/ubuntu:jammy"
+	outputDir       = "/tmp/output"
+	buildScriptName = "_build.sh"
+)
+
+var (
+	varCacheAptMount = llb.AddMount("/var/cache/apt", llb.Scratch(), llb.AsPersistentCacheDir("dalec-windows-var-cache-apt", llb.CacheMountLocked))
```
cpuguy83, line 26: For these to work, we also need to remove `/etc/apt/apt.conf.d/docker-clean` from the base image.
You should be able to acheive this by mounting over the file, like `llb.AddMount("/etc/apt/apt.conf.d/docker-clean", llb.Scratch())`

pmengelbert:
> Looks like the rebase on this was incomplete. Code is not compiling.

Working on it now

test/fixtures/moby-containerd.yml
```diff
@@ -0,0 +1,97 @@
...
+        GO111MODULE=on go build \
+          -mod=vendor \
+          -o bin/containerd-shim-runhcs-v1.exe \
+          ./cmd/containerd-shim-runhcs-v1
+
+artifacts:
+  binaries:
+    hcsshim/bin/containerd-shim-runhcs-v1.exe:
+    containerd/bin/containerd.exe:
+    containerd/bin/ctr.exe:
```
cpuguy83, line 97: Can we move this to an integration test?

We do have a (somewhat) generic test function that could possible be reused here.
It may need an extra option or something since we can't really test any container output at this time.

cpuguy83 (changes requested):

test/fixtures/moby-containerd.yml
pmengelbert: - [ ] Move yaml fixtures to integration test suit

pmengelbert:
Just pushed a few changes. We should have a working container target, but I haven't migrated the test yet.

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,188 @@
...
+
+type RunOptFunc func(*llb.ExecInfo)
+
+func (f RunOptFunc) SetRunOption(ei *llb.ExecInfo) {
+	f(ei)
+}
+
+func SpecToSourcesLLB(spec *dalec.Spec, sOpt dalec.SourceOpts) (map[string]llb.State, error) {
+	m := make(map[string]llb.State, len(spec.Sources))
+	keys := dalec.SortMapKeys(spec.Sources)
```
cpuguy83, line 73: This sorting isn't doing anything since it is putting everything back into a map.

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,188 @@
...
+	return ref, &image.Image{}, err
+}
+
+type RunOptFunc func(*llb.ExecInfo)
+
+func (f RunOptFunc) SetRunOption(ei *llb.ExecInfo) {
+	f(ei)
+}
+
+func SpecToSourcesLLB(spec *dalec.Spec, sOpt dalec.SourceOpts) (map[string]llb.State, error) {
```
cpuguy83, line 71: Can we unexport this?

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,188 @@
...
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+	ref, err := res.SingleRef()
+	// Do not return a nil image, it may cause a panic
+	return ref, &image.Image{}, err
+}
+
+type RunOptFunc func(*llb.ExecInfo)
```
cpuguy83, line 65: This should move to the root dalec package or be unexported.

cpuguy83 (changes requested):
I think we need to validate the spec for runtime dependencies for Windows, since TMK, we don't really have any way to support that.

frontend/windows/handle_zip.go
pmengelbert: The calls to `AsState` don't matter? I was thinking that because we're manipulating LLB we want to do them in the same order so we don't get cache misses. LMK if that doesn't matter here.

frontend/windows/handle_zip.go
pmengelbert: I removed this in the latest commit since I found something better.

frontend/windows/handle_zip.go
pmengelbert: done

frontend/windows/handle_zip.go
pmengelbert: done

frontend/windows/handler.go
```diff
@@ -0,0 +1,23 @@
+package windows
+
+import (
+	"github.com/Azure/dalec/frontend"
+	"github.com/moby/buildkit/frontend/subrequests/targets"
+)
+
+const (
+	targetKey = "windows"
```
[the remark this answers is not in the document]
pmengelbert: With the router changes, can we denote the cross-compilation with another path component? The output of this target is still a windows container, so I'd prefer to have it just named `windows`.

I was thinking `windows/cross/container` `windows/cross/zip`. The router will make these code paths pretty easy to follow. WDYT? Am I missing anything here?

test/fixtures/moby-containerd.yml
pmengelbert: done

frontend/windows/handle_zip.go
pmengelbert: - [ ] (Peter) Fix this

frontend/windows/handler.go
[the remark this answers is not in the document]
cpuguy83: I don't think this will work because an actual windows build would be totally different, but this is used in the spec directly:

```yaml
targets:
  windows:
    # windows specific stuff
```

The main thing there is build dependencies will be different.
Looking through what we have, that may be the only thing that's particularly problematic in a world where the targets are shared.

Perhaps the target key could be the path? In which case I'd do something like `jammy/cross/windows` or something to denote it's the jammy toolchain.
I'm not sure if this works or not so maybe worth testing?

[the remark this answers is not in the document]
pmengelbert: I guess this has always been ambiguous up until now. Does the first component of the path denote the build distro/OS or the output distro/OS?

We haven't had to decide this yes since the build & output distros/OSes have always been the same.

frontend/windows/handle_zip.go
pmengelbert: - [ ] Remove sorting here

frontend/windows/handler.go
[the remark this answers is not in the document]
cpuguy83: It is indeed a combination of build toolchain and output.

frontend/windows/handle_zip.go
pmengelbert: done

frontend/windows/handle_zip.go
pmengelbert: done

frontend/windows/handler.go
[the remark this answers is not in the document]
pmengelbert: As we discussed, I'm changing the target name to "windowscross". If there is a routing solution that disambiguates build toolchain from output, we'll include those changes in #172 (the routing PR)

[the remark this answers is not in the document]
pmengelbert: this is done

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,213 @@
...
+	sources, err := specToSourcesLLB(spec, sOpt)
+	if err != nil {
+		return llb.Scratch(), err
+	}
+
+	patchedMap := dalec.PatchSources(worker, spec, sources)
+
+	patched := mapToArraySortedByKeys(patchedMap)
+
+	combined := dalec.MergeAtPath(llb.Scratch(), patched, "/")
```
cpuguy83, line 122: Instead of merging states here, can we mount each one directly into the build?

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,213 @@
...
+	script := generateInvocationScript(binaries)
+
+	buildDeps := spec.GetBuildDeps(targetKey)
+
+	work := worker
+	if len(buildDeps) > 0 {
+		slices.Sort(buildDeps)
+
+		work = work.Run(
+			shArgs("apt-get update && apt-get install -y "+strings.Join(buildDeps, " ")),
```
cpuguy83, line 135: Can you make this a separate `llb.StateOption`?

Something like `work = worker.With(installBuildDeps)`
This can also handle the case of empty build deps.

image.go
```diff
@@ -47,9 +56,62 @@ type ImageConfig struct {
...
+
+type ConfigOpt func(*imageBuilderConfig)
+
+func WithPlatform(p ocispecs.Platform) ConfigOpt {
+	return func(c *imageBuilderConfig) {
+		c.platform = &p
+	}
+}
+
+func BuildImageConfig(ctx context.Context, client gwclient.Client, spec *Spec, targetKey string, dflt string, opts ...ConfigOpt) (*DockerImageSpec, error) {
```
cpuguy83, line 71: We should not be importing the gateway client in this package.
Instead of doing the `ResolveImageConfig` in here, can callers just pass an image config?

That also makes this more easily testable.

test/fixtures/moby-containerd.yml
```diff
@@ -0,0 +1,98 @@
+# syntax=ghcr.io/azure/dalec/frontend:latest
```
cpuguy83, line 1: Can you remove this file?

test/windows_test.go
```diff
@@ -0,0 +1,225 @@
...
+)
+
+func TestWindows(t *testing.T) {
+	t.Parallel()
+
+	ctx := startTestSpan(baseCtx, t)
+	testWindows(ctx, t, "windowscross/container")
+}
+
+func testWindows(ctx context.Context, t *testing.T, buildTarget string) {
```
cpuguy83, line 21: What's the difference between the existing test function and this one?

cpuguy83 (changes requested):

test/fixtures/moby-containerd.yml
pmengelbert: Yes, my mistake

test/windows_test.go
pmengelbert: It removes the TestSpecs from the third test, and I think there was one other change. Since we can't run commands in the final container, there's no way to run the tests from the spec. Is there a way around that?

image.go
pmengelbert: Sure

frontend/build.go
```diff
@@ -13,13 +14,28 @@ import (
...
+
 func loadSpec(ctx context.Context, client *dockerui.Client) (*dalec.Spec, error) {
 	src, err := client.ReadEntrypoint(ctx, "Dockerfile")
 	if err != nil {
 		return nil, fmt.Errorf("could not read spec file: %w", err)
 	}
 
-	spec, err := dalec.LoadSpec(bytes.TrimSpace(src.Data))
+	opts := []dalec.LoadOpt{}
+	if strings.HasPrefix(client.Target, windowsTargetPrefix) {
```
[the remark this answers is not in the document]
pmengelbert: Currently validation happens at load time. I will move this into the handler, it makes sense that each handler can do its own additional validations after the main validation is done.

test/windows_test.go
cpuguy83: Maybe we can re-use the other test function but have a spec generator with an option elide tests from the generated spec.

frontend/build.go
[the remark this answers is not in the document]
adamperlin: I think that is a good idea. From a delegation standpoint, it seems like `loadSpec` should check for validity at the Dalec spec level, and if individual handlers want to do further validation then they can

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,213 @@
...
+	return ref, &dalec.DockerImageSpec{}, err
+}
+
+func specToSourcesLLB(spec *dalec.Spec, sOpt dalec.SourceOpts, opts ...llb.ConstraintsOpt) (map[string]llb.State, error) {
+	out := make(map[string]llb.State, len(spec.Sources))
+	for k, src := range spec.Sources {
+		isDir := dalec.SourceIsDir(src)
+
+		s := ""
+		switch {
```
adamperlin, line 70: I think this switch could be moved to a method on `Source`, perhaps?

source.go
```diff
@@ -419,20 +420,30 @@ func (s Source) Doc(name string) (io.Reader, error) {
...
-		// on each iteration, mount source state to /src to run `patch`, and
-		// set the state under /src to be the source state for the next iteration
-		sourceState = worker.Run(
-			llb.AddMount("/patch", patchState, llb.Readonly, llb.SourcePath(p.Source)),
-			llb.Dir("src"),
-			shArgs(fmt.Sprintf("patch -p%d < /patch", *p.Strip)),
-			WithConstraints(opts...),
-		).AddMount("/src", sourceState)
-	}
+type patchFunc func(worker, sourceState llb.State, sourceToState map[string]llb.State, patchNames []PatchSpec, opts ...llb.ConstraintsOpt) llb.State
```
adamperlin, line 423: Not sure if we need the nested closure here? Maybe we could just add an extra `Source` or even `isDir` param to the original function?

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,213 @@
...
+	outName := filepath.Join(outputDir, name+".zip")
+	zipped := worker.Run(
+		shArgs("zip "+outName+" *"),
+		llb.Dir("/tmp/artifacts"),
+		llb.AddMount("/tmp/artifacts", artifacts),
+	).AddMount(outputDir, llb.Scratch())
+	return zipped
+}
+
+func generateInvocationScript(binaries []string) *strings.Builder {
```
adamperlin, line 162: any particular reason to do this as string script instead of building up LLB steps in this function?

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,213 @@
...
+			st = llb.Scratch().File(llb.Copy(st, "/", filepath.Join("/", k)))
+		}
+
+		out[k] = st
+	}
+
+	return out, nil
+}
+
+func mapToArraySortedByKeys[T any](m map[string]T) []T {
```
adamperlin, line 103: nit: maybe just `valuesSortedByKey`?

test/windows_test.go
pmengelbert: Oh, there is also the issue that some of the test specfiles have runtime dependencies, which windows outputs can't have. I will look at re-using the other test function.

test/fixtures/moby-containerd.yml
pmengelbert: done

frontend/build.go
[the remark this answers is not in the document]
pmengelbert: Done.

I added validations to the `targetWrapper` so that validations functions can be registered alongside handlers. These are automatically run before the call to `Build`.

frontend/handlers.go
```diff
@@ -103,8 +121,22 @@ var registeredHandlers = &handlerList{
...
 // Registered handlers may be overridden by [dalec.Spec.Targets].
-func RegisterHandler(group string, t bktargets.Target, build BuildFunc) {
-	registeredHandlers.Add(group, &targetWrapper{Target: t, Build: build})
+func RegisterHandler(group string, t bktargets.Target, build BuildFunc, opts ...handlerOpt) {
+	var cfg handlerOpts
+	for _, f := range opts {
+		f(&cfg)
+	}
+
+	validations := []validateFunc{}
```
cpuguy83, line 130: We can't do this here.
Target registration needs to be completely separate from everything else.

I think its fine to have targets do extra validation on the spec as needed.

frontend/windows/handle_zip.go
pmengelbert: I couldn't do everything as a mount. This is because some are files which have the same target directory, whereas others are directories and can successfully be mounted to a subdirectory of "/build".

This means sources that are single files are being merged into one state, mounted at "/build", then the directory sources are mounted at subdirectories of "/build".

If there is a better way, please let me know and I'll modify this change.

for now, this is done.

frontend/windows/handle_zip.go
pmengelbert: Removed this function

frontend/handlers.go
pmengelbert: > We can't do this here. Target registration needs to be completely separate from everything else.

Ok I'll change it.

Can you explain why? Helps me learn.

image.go
```diff
@@ -103,3 +165,53 @@ func MergeImageConfig(dst *DockerImageConfig, src *ImageConfig) error {
...
+	}
+
+	if i := spec.Targets[target].Image; i != nil {
+		if i.Entrypoint != "" {
+			cfg.Entrypoint = spec.Targets[target].Image.Entrypoint
+		}
+
+		if i.Cmd != "" {
+			cfg.Cmd = spec.Targets[target].Image.Cmd
+		}
```
DannyBrito, line 191: nit: keep consistent the the value setting for these
```suggestion
		if i.Entrypoint != "" {
			cfg.Entrypoint = i.Entrypoint
		}

		if i.Cmd != "" {
			cfg.Cmd = i.Cmd
		}
```

source.go
```diff
@@ -283,19 +284,19 @@ func WithCreateDestPath() llb.CopyOption {
...
+		return true
 	case src.HTTP != nil:
-		return false, nil
+		return false
 	case src.Inline != nil:
-		return src.Inline.Dir != nil, nil
+		return src.Inline.Dir != nil
 	default:
-		return false, fmt.Errorf("unsupported source type")
+		panic("unreachable")
```
DannyBrito, line 299: In the unlikely event that it did reach here, would it be a good idea to have a verbose error message?
```suggestion
		panic("unreachable: unsupported source type")
```

frontend/windows/handle_zip.go
```diff
@@ -0,0 +1,230 @@
...
+	buf := bytes.NewBuffer(nil)
+
+	fmt.Fprintln(buf, "#!/usr/bin/env sh")
+	fmt.Fprintln(buf, "set -x")
+
+	for i, step := range spec.Build.Steps {
+		fmt.Fprintln(buf, "(")
+
+		for k, v := range step.Env {
+			fmt.Fprintf(buf, "export %s=%s\n", k, v)
```
DannyBrito, line 214:
```suggestion
			fmt.Fprintf(buf, "export %s=\"%s\"\n", k, v)
```

frontend/windows/handle_zip.go
pmengelbert: The script is just complicated enough (it loops over the spec's `artifacts.binaries`) to warrant a subroutine.

frontend/handlers.go
cpuguy83: Its mixing concerns.
Part of the routing logic changes removes this spec loading part as well, so the build targets need to load the spec themselves.

source.go
pmengelbert: done

frontend/windows/handle_zip.go
pmengelbert: done

source.go
pmengelbert: Unless I am mistaken, I don't think the panic message gets propagated back to the client if the frontend process panics. @cpuguy83 is this right?

frontend/windows/handle_zip.go
pmengelbert: Good catch, thanks

pmengelbert: done

source.go
cpuguy83: Correct, its only in the daemon logs.
Though we could possibly recover panics and send them back as errors.

cpuguy83: I'm curious why remove the error?
I guess at this point it is a bug, but maybe, worth keeping the error and calling out it is a bug that should be filed.

cpuguy83: Having just hit this with the jammy PR, it is indeed really annoying to have this error.
I think its fine to panic.

pmengelbert: > Correct, its only in the daemon logs. Though we could possibly recover panics and send them back as errors.

I like this idea, because I've run into unexplained panics before and it's been a bit of a pain to troubleshoot.,

image.go
pmengelbert: done

frontend/handlers.go
pmengelbert: done

frontend/imgconfig.go
```diff
@@ -0,0 +1,117 @@
...
+	cfg := img.Config
+	if err := dalec.MergeImageConfig(&cfg, mergeSpecImage(spec, targetKey)); err != nil {
+		return nil, err
+	}
+
+	img.Config = cfg
+	return &img, nil
+}
+
+func getBaseOutputImage(spec *dalec.Spec, target, defaultBase string) string {
```
cpuguy83, line 69: Let's make this exported.
Then we can pass in the output of this as in input to `BuildImageConfig`, which removes a function paramerter for `BuildImageConfig`.

test/windows_test.go
```diff
@@ -0,0 +1,225 @@
...
+echo "${FOO}_1" > foo1.txt
+echo "$BAR" > bar.txt
+`,
+					},
+				},
+			},
+
+			Image: &dalec.ImageConfig{
+				Post: &dalec.PostInstall{
+					Symlinks: map[string]dalec.SymlinkTarget{
```
cpuguy83, line 190: Looks like we need to add support for symlinks.
Discussing this, I don't know how we could create an actual symlink for Windows _from_ a linux worker.
To get around this we could copy the binaries over.

cpuguy83 (changes requested):

frontend/imgconfig.go
pmengelbert: done

source.go
```diff
@@ -451,7 +473,13 @@ func PatchSources(worker llb.State, spec *Spec, sourceToState map[string]llb.Sta
 			continue
 		}
 		opts = append(opts, ProgressGroup("Patch spec source:"+sourceName))
-		states[sourceName] = patchSource(worker, sourceState, states, patches, withConstraints(opts))
+
+		var nestedPath string
+		if SourceIsDir(spec.Sources[sourceName]) {
```
cpuguy83, line 478: This is new behavior that is not mimicked in the rpm patcher.
Was there a particular reason to add it?

frontend/windows/handle_container.go
```diff
@@ -0,0 +1,124 @@
...
+
+func getTargetPlatform(bc *dockerui.Client) (ocispecs.Platform, error) {
+	platform := defaultPlatform
+
+	switch len(bc.TargetPlatforms) {
+	case 0:
+	case 1:
+		platform = bc.TargetPlatforms[0]
+	default:
+		return ocispecs.Platform{}, fmt.Errorf("multiple targets for a windows build, only amd64 is supported")
```
cpuguy83, line 112: This error mentions amd64 but doesn't validate it.
Also I'm not sure why it would matter since we aren't trying to run anything.

cpuguy83 (changes requested):

source.go
adamperlin: I think the reason it was added here (correct me if I'm wrong) is because `specToSourcesLLB` takes the extra step of nesting a source within a directory in the LLB itself. I.e., the LLB contains the directory as well as the contents. I think that nesting step could be moved out of `spectToSourcesLLB` (if indeed it is needed), so we don't have to handle that behavior here

Dockerfile
```diff
@@ -4,8 +4,8 @@ FROM go  AS frontend-build
 WORKDIR /build
 COPY . .
 ENV CGO_ENABLED=0 GOFLAGS=-trimpath
-ARG TARGETARCH TARGETOS
-ENV GOOS=${TARGETOS} GOARCH=${TARGETARCH}
+ARG TARGETARCH TARGETOS GOFLAGS
```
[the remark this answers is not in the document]
pmengelbert: The purpose of this is to make it easier to build the binaries with debug symbols and without optimizations. I find it helpful to compile a debug version, and then attach to the frontend process with a debugger. I had been manually editing the Dockerfile to add `-gcflags=all="-N -l"` in the go compilation commands. I kept accidentally checking in the changes. It also got annoying having to add it over and over.

frontend/windows/handle_container.go
pmengelbert: We are making sure there are not multiple platforms specified. Because only one platform is supported (`windows/amd64`), if there is more than one platform requested by the user, the request cannot be satisfied.

The `amd64` mention was an attempt at providing additional information. The idea was to tell the user, "try again, but only with amd64".

I improved the error message. LMK if you want something different.

source.go
pmengelbert: @adamperlin yes that's correct. I went back and changed `specToSourcesLLB` so that it doesn't nest the sources in the directories. Now that's happening in the `RunOpt` that mounts the sources. I restored the two patching functions (the public and private) to the original versions. This is much cleaner. Thanks for pointing this out @cpuguy83.

This is done

frontend/windows/handle_container.go
pmengelbert: done unless you want me to change it again.

source.go
pmengelbert: This is not done. Working on it now.

pmengelbert: Now this is done.

Dockerfile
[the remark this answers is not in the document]
cpuguy83: I think this would need to go above the `ENV` line with `GOFLAGS` with the default  value for the arg as `-trimpath`.

Dockerfile
```diff
@@ -4,8 +4,8 @@ FROM go  AS frontend-build
...
+ENV GOOS=${TARGETOS} GOARCH=${TARGETARCH} GOFLAGS=${GOFLAGS}
```
cpuguy83, line 8:
```suggestion
ENV CGO_ENABLED=0
ARG TARGETARCH TARGETOS GOFLAGS=-trimpath
```

pmengelbert: done

cpuguy83 (approved):

cpuguy83:
Great work, @pmengelbert.
Thanks for working through the constant back and forth on the review on this one.

PR #110, opened 2024-02-07
Add unit tests to validate source llb

cpuguy83:
These tests don't require buildkitd or *network access. They generate the LLB from source2llbgetter and validate that the protobufs are what we expect them to be.

Testing this way is *significantly* faster than trying to create integration tests for all possible permutations of sources.

*I can't guarentee that one of the LLB functions don't access the network, in fact I know `llb.Git` does, which is why I've created a stub listener. Also the `dockerfile2llb` needs a stub resolver so it doesn't reach out to a registry.


go.mod
```diff
@@ -25,15 +25,18 @@ require (
...
 	github.com/buger/jsonparser v1.1.1 // indirect
 	github.com/cenkalti/backoff/v4 v4.2.0 // indirect
 	github.com/containerd/console v1.0.3 // indirect
 	github.com/containerd/continuity v0.4.2 // indirect
 	github.com/containerd/ttrpc v1.2.2 // indirect
 	github.com/containerd/typeurl/v2 v2.1.1 // indirect
+	github.com/distribution/reference v0.5.0 // indirect
 	github.com/docker/distribution v2.8.2+incompatible // indirect
-	github.com/docker/docker v24.0.7+incompatible // indirect
+	github.com/docker/docker v25.0.2+incompatible // indirect
```
cpuguy83, line 38: This should have been updated by dependabot when it updated the buildkit lib...

source.go
```diff
@@ -184,25 +187,35 @@ func source2LLBGetter(s *Spec, src Source, name string, forMount bool) LLBGetter
 			opts = append(opts, llb.Filename(name))
 			return llb.HTTP(https.URL, opts...), nil
 		case src.Context != nil:
-			st, err := sOpt.GetContext(dockerui.DefaultLocalNameContext, localIncludeExcludeMerge(&src))
+			name := src.Context.Name
+			if name == "" {
```
adamperlin, line 191: I think this should be handled in `load.go` already?

source_test.go
```diff
@@ -0,0 +1,638 @@
...
+		Git: &SourceGit{
+			URL:    fmt.Sprintf("user@%s:test.git", addr),
+			Commit: t.Name(),
+		},
+	}
+
+	ops := getSourceOp(ctx, t, src)
+	checkGitOp(t, ops, &src)
+
+	t.Run("with subdir", func(t *testing.T) {
```
adamperlin, line 41: It looks like the subpath/include-exclude setup is duplicated across different tests. Maybe you could use an options type pattern to set a subdir and include/exclude on the source?

cpuguy83: I feel like this isn't really enough logic to make it worth abstracting.

test/fixtures/nested.yml
```diff
@@ -14,8 +14,8 @@ sources:
   src:
     build:
       source:
-        context:
-          name: test/fixtures/nested
+        path: test/fixtures/nested
+        context: {}
```
cpuguy83, line 18: Just to note, because of how things get unmarhalled, we need to include *something* in this context struct in order for it to be preserved (or as I've done here with the explicit `{}`

source.go
cpuguy83: Done.

adamperlin:
Can we add tests where we have source mounts, since that can change the filter operation behavior? I can also add that in as part of #117 if that works better. Just for documentation purposes, the particular cases are: mounting a source to a command where the source being mounted does NOT handle its own sub-path, and mounting a source to a command where the source being mounted DOES handle its own path.

adamperlin (approved):
I think this looks good. I'm really glad we can test the filter handling logic in particular in a methodical way. The only other question I still have has to do with source mount tests, but I can also add that as part of my current PR once this is in

cpuguy83:
CI failure looks legit, seems like something in the last change did something.

cpuguy83:
Ah it was because of the 2nd commit that fixes some formatting directives that `go vet` does not like.
```go
fmt.Fprintf(b, `%%build\n`)
```

Because of the backticks it was treating the `\n` as a literal.

cpuguy83:
Opened #129 to track building on top of this.
Since this change is fixing real bugs that were discovered I'm going to go ahead and merge and we can improve this later.

PR #113, opened 2024-02-07
[WIP] add arm64 runner for toolchain image

DannyBrito:
[WIP] - added `arm64` runner to build arm64 toolchain image, as this is build in two runners for the matrix strategy can't use `--push` from docker buildx. Decided to first implementation, to load the image result and push manually with arch tags and also create and push the manifest list. I would appreciate any feedback.


cpuguy83:
Looks like we are actually going to drop this image: #115

PR #115, opened 2024-02-08
Remove usage of mariner toolkit

cpuguy83:
This was originally a requirement from the mariner team to include our rpm's in mariner repos.
After discussions with that team, that is no longer a requirement.

This is now replaced with direct `rpmbuild` invocations.  This is significantly simpler as it does not to require us to setup chroots which the toolkit required.
It is also significantly faster since there is no intermediate step doing a (very slow) package graph calculation.

It also means we no longer need to maintain an extra image (which itself was fairly large) to support the toolkit.

In short:

* Less code
* Smaller build deps
* Faster build times
* More flexible build environment


cpuguy83:
e2e suite is running 3x faster 🎉

DannyBrito:
Question, does removing the usage of mariner toolkit mean the builds will have access to internet during build time now?

cpuguy83:
We shouldn't treat it like such.
Ideally if you have the src.rpm you can build the package.

frontend/mariner2/handle_rpm.go
```diff
@@ -2,63 +2,46 @@ package mariner2
...
 
-// marinerTdnfCacheWithPrefix returns a run option that sets up a persistent cache for tdnf.
-// The tdnf cache is mounted at `[prefix]/var/cache/tdnf`.
-//
-// This makes it so that when tdnf needs to download packages, repodata, etc it will use the cache dir.
-// Repeated builds will benefit from this as the cache will be reused.
-func marinerTdnfCacheWithPrefix(prefix string) llb.RunOption {
-	return llb.AddMount(filepath.Join(prefix, marinerTdnfCacheDir), llb.Scratch(), llb.AsPersistentCacheDir("mariner2-tdnf-cache", llb.CacheMountShared))
+func tdnfCacheMountWithPrefix(prefix string) llb.RunOption {
+	return llb.AddMount(filepath.Join(prefix, tdnfCacheDir), llb.Scratch(), llb.AsPersistentCacheDir(tdnfCacheName, llb.CacheMountLocked))
```
adamperlin, line 30: Could you explain why we are switching to `CacheMountLocked` here?

cpuguy83: tdnf is exiting due to locking issues with shared mode.

cpuguy83: To be clear, I'm not sure why it's not waiting for the lock, may be an issue with tdnf vs dnf.

frontend/rpm/handle_rpm.go
```diff
@@ -0,0 +1,35 @@
...
+	return workerImg.Run(
+		// some notes on these args:
+		//  - _topdir is the directory where rpmbuild will look for SOURCES and SPECS
+		//  - _srcrpmdir is the directory where rpmbuild will put the source RPM
+		//  - _rpmdir is the directory where rpmbuild will put the RPM
+		//  - --buildroot is the work directory where rpmbuild will build the RPM
+		//  - -ba tells rpmbuild to build the source and binary RPMs
+		//
+		// TODO(cpuguy83): specPath would ideally never change.
+		// We don't want to have to re-run this step just because the path changed, this should be based on inputs only (ie the content of the rpm spec, sources, etc)
```
adamperlin, line 26: How are you thinking of making this work without the `specPath` as a parameter? Pass the spec through inline? Or maybe if we are only ever building one SPEC at once (possible asterisk) maybe we could hardcode the SPEC path for this step and then copy the SPEC somewhere else later?

helpers.go
```diff
@@ -200,3 +201,18 @@ type runOptFunc func(*llb.ExecInfo)
 func (f runOptFunc) SetRunOption(ei *llb.ExecInfo) {
 	f(ei)
 }
+
+// ProgressGroup creates a progress group with the given name.
+// If a progress group is already set in the constraints the id is reused.
+// If no progress group is set a new id is generated.
+func ProgressGroup(name string, weak bool) llb.ConstraintsOpt {
```
adamperlin, line 208: Just for my own understanding can you possibly explain the reasoning for this change? Why is better to reuse the ID of a progress group if one already exists?

frontend/mariner2/handle_rpm.go
adamperlin: Ah got it. Maybe we could add a comment explaining that?

helpers.go
cpuguy83: It lets you use the same progress group but add a new message to it.

frontend/rpm/handle_rpm.go
cpuguy83: I'm not sure right now, which is why I added the todo.
In practice I doubt this will even be an issue.

helpers.go
adamperlin: I think I answered my own question. It looks like we're being more careful now about threading a progress group through to all the rpm building steps, and making sure everything is under the same progress group

adamperlin: Whoops didn't see your response

frontend/mariner2/handle_rpm.go
cpuguy83: Sure. Pushed that up, along with an extra commit to try and see if dnf is better here.
I'd like to use shared if we can.

adamperlin:
> Question, does removing the usage of mariner toolkit mean the builds will have access to internet during build time now?

I want to make sure I understand what @DannyBrito is saying here. What were the restrictions on internet access with the mariner toolkit?

frontend/mariner2/handle_rpm.go
cpuguy83: Yeah it just doesn't like being shared.
I've not seen these erorrs before now, but it seems like dnf appears to be somewhat racey in how it handles locks.

cpuguy83: I've removed that 2nd commit.

cpuguy83:
@adamperlin Technically there was network access, just no /etc/resolve.conf so no DNS resolution.
This is why in our go examples we do a `go mod download` as part of a source rather than just letting `go build` handle it.

Its also not abnormal to have network access restricted during build phases to help ensure you've captured everything in your source packages.

jeremyrickard (approved):
lgtm

PR #117, opened 2024-02-09
Source Refactor

adamperlin:
Refactors the source generation code in `sources.go.` In particular:
- Moves the logic for generating each source's LLB to a separate function for each source variant
- Changes source filtering logic to make filtering parameters as explicit as possible for each source
- Attempts to push source mounting logic to the caller as much as possible by allowing the caller to pass in a custom callback which wraps the source filtering function

Note: This could probably wait for #110, as those unit tests would probably be better able to catch any bugs in the filter handling logic introduced here.


test/source_test.go
```diff
@@ -65,3 +67,114 @@ func testCmdSource(t *testing.T) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		checkFile(ctx, t, filepath.Join(sourceName, "file.txt"), res, []byte("test contents\n"))
+		return gwclient.NewResult(), nil
+	})
+}
+
+func testSourceMountPathHandledByFilter(t *testing.T) {
```
adamperlin, line 131: I realize this test isn't ideal since it relies on the network and a particular git repo, but this test is temporary. This test case could probably be handled by a test that checks protobuf directly like in #110, or an inline source once those are introduced

cpuguy83: Yeah, let's get #110 in and do these thests there, I think.

source.go
```diff
@@ -8,11 +8,147 @@ import (
...
+			return llb.Scratch(), false, err
+		}
+
+		return st.With(filter(src.Path, src.Includes, src.Excludes)), true, nil
+	}
+
+	return llb.Scratch(), false, errNoSourceVariant
+}
+
+func (src *SourceContext) AsState(parent *Source, sOpt SourceOpts, opts ...llb.ConstraintsOpt) (llb.State, error) {
```
cpuguy83, line 71: Can we pass the includes/excludes directly ins5tead of the full parent source?

source.go
```diff
@@ -8,11 +8,147 @@ import (
...
+	}
+	gOpts = append(gOpts, withConstraints(opts))
+
+	st := llb.Git(ref.Remote, src.Commit, gOpts...)
+	return st, nil
+	// TODO: Pass git secrets
+}
+
+func (src *SourceDockerImage) AsState(name string, parent *Source, sOpt SourceOpts, opts ...llb.ConstraintsOpt) (llb.State, error) {
+	bklog.L.Printf("DockerImage: %v", src)
```
cpuguy83, line 102: Debugging left behind

source.go
```diff
@@ -8,11 +8,147 @@ import (
...
+		gOpts = append(gOpts, llb.KeepGitDir())
+	}
+	gOpts = append(gOpts, withConstraints(opts))
+
+	st := llb.Git(ref.Remote, src.Commit, gOpts...)
+	return st, nil
+	// TODO: Pass git secrets
+}
+
+func (src *SourceDockerImage) AsState(name string, parent *Source, sOpt SourceOpts, opts ...llb.ConstraintsOpt) (llb.State, error) {
```
cpuguy83, line 101: Can you pass just the path instead of `Source`?

source.go
```diff
@@ -8,11 +8,147 @@ import (
...
+func DefaultSourceFilter(extract string, includes, excludes []string, opts ...llb.ConstraintsOpt) llb.StateOption {
+	return filterState(extract, includes, excludes, opts...)
+}
+
+func TargetMountSourceFilter(_ string, includes, excludes []string, opts ...llb.ConstraintsOpt) llb.StateOption {
+	return filterState("/", includes, excludes, opts...)
+}
+
+func (src *SourceBuild) AsState(name string, _ *Source, filter FilterFunc, sOpt SourceOpts, opts ...llb.ConstraintsOpt) (llb.State, error) {
+	st, _, err := GetSource(src.Source, name, filter, sOpt, opts...)
```
cpuguy83, line 130: I'm not sure I'm following why we are passing around a filter function.
The filter should be understood by `GetSource`, and the filter from a parent should not affect the value of `src.Source` but rather the `llb.State` returned by this function.

source.go
```diff
@@ -8,11 +8,147 @@ import (
 	"strings"
 
 	"github.com/moby/buildkit/client/llb"
+	"github.com/moby/buildkit/util/bklog"
 	"github.com/moby/buildkit/util/gitutil"
 	"github.com/pkg/errors"
 )
 
-type LLBGetter func(sOpts SourceOpts, opts ...llb.ConstraintsOpt) (llb.State, error)
+type FilterFunc = func(string, []string, []string, ...llb.ConstraintsOpt) llb.StateOption
```
cpuguy83, line 16: In general I'm wondering if we need to be passing around a `FilterFunc`.
If we do, then it should be  passed as an `llb.StateOption` rather than a `func() -> llb.StateOption`

cpuguy83 (changes requested):

test/source_test.go
adamperlin: Currently working on adding these in with the new source tests

source.go
adamperlin: sure. we will also need to pass path in as well

source.go
adamperlin: Ah this is a bug. The filter passed in the recursive call should be no-op

source.go
adamperlin: I think I see what you're getting at now. There is no need for this to be passed in, it can be applied by the caller. I am changing to this approach

source.go
adamperlin: I'm changing things to not pass the filter through and have the caller apply the filter instead

source.go
```diff
@@ -83,161 +231,53 @@ func generateSourceFromImage(s *Spec, name string, st llb.State, cmd *Command, s
...
-	return p == "" || p == "/" || p == "."
-}
-
-func needsFilter(o *filterOpts) bool {
-	if !isRootPath(o.source.Path) && !o.forMount && !o.pathHandled {
-		return true
-	}
-	if o.includeExcludeHandled {
-		return false
+func Source2LLBGetter(_ *Spec, src Source, name string, sOpt SourceOpts, opts ...llb.ConstraintsOpt) (llb.State, error) {
```
cpuguy83, line 248: `Source2LLBGetter` is the only caller of `GetSource` and the only caller of `GetFilter`.
I don't think we'd ever want to call `GetSource` without `GetFilter`.

Probably that should be in `GetSource`.

source.go
```diff
@@ -12,6 +12,150 @@ import (
...
+var errNoSourceVariant = fmt.Errorf("no source variant found")
+
+func (src Source) handlesOwnPath() bool {
+	// docker images handle their own path extraction if they have an attached command,
+	// and this information is needed in the case of mounts when we can do path
+	// extraction at mount time
+	return src.DockerImage != nil && src.DockerImage.Cmd != nil
+}
+
+func GetFilter(src Source, forMount bool, opts ...llb.ConstraintsOpt) llb.StateOption {
```
cpuguy83, line 26: This should be uneported, and maybe inlined in `GetSource`?

source.go
```diff
@@ -12,6 +12,150 @@ import (
...
+		if src.DockerImage.Cmd != nil {
+			// if a docker image source has a command,
+			// the path extraction will be handled with a mount on the command
+			path = "/"
+		}
+
+		return filterState(path, src.Includes, src.Excludes)
+	}
+
+	return func(_ llb.State) llb.State { return llb.Scratch() }
```
cpuguy83, line 51: I think this should return the input state?

source.go
cpuguy83: I guess for code readability its nice to have some things split, but we also don't need to export those functions.

Maybe a `GetSource` (or maybe even `func(s *Source) AsState(opts ...llb.ConstraintsOpt) (llb.state, error)`) that calls `getSource().With(getFilter())`

source_test.go
```diff
@@ -176,6 +177,13 @@ func TestSourceHTTP(t *testing.T) {
 	}
 }
 
+func toImageRef(ref string) string {
+	return "docker-image://" + ref
+}
+
+var NoMountCheck = []expectMount{}
+var RootMount = expectMount{dest: "/", selector: "", typ: pb.MountType_BIND}
```
cpuguy83, line 185:
```suggestion
var (
	noMountCheck = []expectMount{}
	rootMount = expectMount{dest: "/", selector: "", typ: pb.MountType_BIND}
)
```

cpuguy83 (changes requested):

source.go
adamperlin: Yeah, I also like to have things split because of the mounting case. I.e., when we are mounting that should only affect the filter, so I wasn't a fan of passing the `forMount` flag into `getSource`. But I've wrapped `getSource` and `filter` into a single function.

source.go
adamperlin: I'd rather not inline it to avoid passing `forMount` through `GetSource`, if that's ok

cpuguy83 (approved):

cpuguy83:
This says there's conflicts.

cpuguy83:
It doesn't state the conflicts which is super weird.

image

PR #130, opened 2024-02-13
update test specs.

DannyBrito:


test/fixtures/cmd-src-ref.yml
```diff
@@ -21,8 +21,7 @@ sources:
         mounts:
           - dest: "/foo"
             spec:
-              context:
-                path: .
+              path: .
```
cpuguy83, line 24: For these, we need to not bother setting `path` and instead use an explicit `context: {}`

cpuguy83: *Not bother setting path when `path` is `.`, `./`,  or `/`

cpuguy83 (approved):

PR #136, opened 2024-02-15
Revert use of "weak" progress groups.

cpuguy83:
Using weak progress groups seems to hide all the output from the progress group.
Need to understand more of how to use weak groups, for now use non-weak ones for everything so we can get helpful output.


adamperlin (approved):
This lgtm, assuming the `weak` is truly the issue. Was wondering why we were getting much less build output.

jeremyrickard (approved):

PR #138, opened 2024-02-15
Minor fixes

cpuguy83:
1. Fix inline dir should be rooted at `/` not `/<sourcename>`
2. Untangle command/test/testspec 

On 2, this was just some incidental duplication.
The code is easier to reason about as separate implementations.


DannyBrito (approved):
lgtm, if we are trying to move away from using `,inline` there is another needed change https://github.com/Azure/dalec/blob/main/spec.go#L504 `CheckOutput` struct embedded in `FileCheckOutput`

adamperlin (approved):
This lgtm

PR #139, opened 2024-02-16
Migrate some e2e tests to integration tests

cpuguy83:
This moves everything that's not specifically dealing with a particular
source type into an integration test.

The test is designed to work with any target that has a container
output.
Though if/when we add support for building windows containers there will
need to be some changes since we are expecting linux specific outputs
currently.


cpuguy83:
This was merged in as 85ebba1b39ffda886ef02a4ab8a6f57de96435a0 in #141

PR #141, opened 2024-02-16
Add support for adding symlinks post-install

cpuguy83:
Not to be confused with post-install scripts.
This is handled directly in buildkit after the package is installed.
This allows us to customize things for the container without requiring
extra tooling in the image to perform those actions.

This is built on top of #139.
The actual changes are in the 2nd commit.


spec.go
```diff
@@ -175,6 +175,26 @@ type ImageConfig struct {
...
+type PostInstall struct {
+	// Symlinks is the list of symlinks to create in the container rootfs after the package(s) are installed.
+	// The key is the path the symlink should point to.
+	Symlinks map[string]SymlinkTarget `yaml:"symlinks,omitempty" json:"symlinks,omitempty"`
+}
+
+// SymlinkTarget specifies the properties of a symlink
+type SymlinkTarget struct {
+	// Path is the path where the symlink should be placed
+	Path string `yaml:"path" json:"path" jsonschema:"required"`
```
pmengelbert, line 197: What other options do you foresee here?

spec.go
```diff
@@ -175,6 +175,26 @@ type ImageConfig struct {
...
+	// Use this to perform actions that would otherwise require additional tooling inside the container that is not relavent to
+	// the resulting container and makes a post-install script as part of the package unnecessary.
+	Post *PostInstall `yaml:"post,omitempty" json:"post,omitempty"`
+}
+
+// PostInstall is the post install configuration for the image.
+type PostInstall struct {
+	// Symlinks is the list of symlinks to create in the container rootfs after the package(s) are installed.
+	// The key is the path the symlink should point to.
+	Symlinks map[string]SymlinkTarget `yaml:"symlinks,omitempty" json:"symlinks,omitempty"`
```
pmengelbert, line 191: Can this just be `map[string]string`?

pmengelbert:
LGTM, some questions but no reason to block.

pmengelbert (approved):

spec.go
cpuguy83: Did it as a struct because it doesn't prevent us from adding other options later.

spec.go
cpuguy83: https://man7.org/linux/man-pages/man1/ln.1.html

PR #146, opened 2024-02-20
feat: multiple specs in one file, with internal dependency resolution

pmengelbert:
Take a look at `test/fixtures/eraser-manager-build-dep.yml` and `test/fixtures/eraser-manager-runtime-dep.yml`

Try building different targets.

If there is more than one spec in a spec file, you can specify the target with (for the example kubernetes-eraser-manager):
`--target=kubernetes-eraser-manager/mariner2/rpm` or `--target=kubernetes-eraser-manager-compat/mariner2/container`, or similar.


graph.go
```diff
@@ -0,0 +1,293 @@
...
+	v    *vertex
+	w    *vertex
+	kind string
+}
+
+type cycle []*vertex
+type cycleList []cycle
+type orderedDeps []*Spec
+
+func (o orderedDeps) targetSlice(target ...string) []*Spec {
```
adamperlin, line 33: can you name this function something more specific or perhaps add a comment to make it a little clearer what it does?

graph.go
```diff
@@ -0,0 +1,293 @@
...
+		sb.WriteString(component.String())
+		if i+1 == len(cs) {
+			break
+		}
+		sb.WriteRune('\n')
+	}
+	return sb.String()
+}
+
+func minimum[T constraints.Ordered](x, y T) T {
```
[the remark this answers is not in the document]
pmengelbert: Nope, not until 1.21 with the min builtin.
https://stackoverflow.com/a/27516559

graph.go
```diff
@@ -0,0 +1,293 @@
...
+				if w == v {
+					break
+				}
+			}
+
+			w.onStack = false
+			output = append(output, component)
+		}
+
+		return nil
```
adamperlin, line 224: looks like this returns nil in all cases? do we need to return an error from this function?

graph.go
```diff
@@ -0,0 +1,293 @@
...
+	specs    map[string]*Spec
+	ordered  orderedDeps
+	indices  map[string]int
+	vertices []*vertex
+	edges    sets.Set[dependency]
+}
+
+type dependency struct {
+	v    *vertex
+	w    *vertex
```
adamperlin, line 25: slight nit but can we rename these to either `v1' and `v2' or `u` and 'v'? 'u' and 'v' are the most standard for notating an edge, in my experience

graph.go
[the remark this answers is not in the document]
adamperlin: if not can we rename to `min`?

graph.go
pmengelbert: done

graph.go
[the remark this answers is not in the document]
pmengelbert: I was wrong, renamed to min.

graph.go
pmengelbert: done

graph.go
```diff
@@ -0,0 +1,293 @@
...
+	kind string
```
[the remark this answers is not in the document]
pmengelbert: removed

graph.go
pmengelbert: done

cmd/tsort/main.go
```diff
@@ -0,0 +1,38 @@
+package main
```
adamperlin, line 1: just a reminder to remove this file from the final PR

graph.go
```diff
@@ -0,0 +1,288 @@
...
+	edges    sets.Set[dependency]
+}
+
+type dependency struct {
+	v1 *vertex
+	v2 *vertex
+}
+
+type cycle []*vertex
+type cycleList []cycle
```
adamperlin, line 29: does this need a type alias?

graph.go
```diff
@@ -0,0 +1,288 @@
...
+				})
+			}
+		}
+	}
+
+	return BuildGraph.topSort()
+}
+
+// https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm
+func (g *Graph) topSort() error {
```
adamperlin, line 162: can we split this function up? have one function that finds strongly connected components, and another that checks to make sure that components have only one vertex each? (i.e., L232 on in a separate function)

graph.go
pmengelbert: No

graph.go
pmengelbert: Sure

pmengelbert: done

cmd/tsort/main.go
pmengelbert: Done

graph.go
pmengelbert: done

graph.go
```diff
@@ -0,0 +1,293 @@
...
+type vertex struct {
+	name    string
+	index   *int
+	lowlink int
+	onStack bool
+}
+
+var (
+	graphLock  sync.Mutex
+	BuildGraph *Graph
```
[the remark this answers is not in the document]
adamperlin: Personally, I'm a fan of eliminating global state unless it's absolutely necessary, and I don't see a compelling need in this case. It's kind of necessary for registeredHandlers to be global, since the lifetime of that data is the entire run time of a dalec frontend instance. I'm not sure if we can assume that the lifetime of a build graph will be for the entire run of a Dalec frontend?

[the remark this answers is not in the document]
adamperlin: Now that I'm thinking about it, the build graph corresponds to a dalec (multi-)spec, and the frontend will be invoked individually on Dalec spec files, so I guess it's ok lifetime-wise. What I dislike about globals is that it makes the parameterization of functions implicit. I.e., now we have a bunch of handlers that are parameterized on `dalec.BuildGraph` which is data that is built up at runtime, and the behavior will change dramatically based on changes to that global, but the argument doesn't show up in the function signatures or in any of the data structures passed to those functions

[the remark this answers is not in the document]
pmengelbert: It will involve a lot of function/method signature changes for several reasons:
- There are no (public) methods on `Graph` that allow mutation of the data structure. Once it has been initialized, it is immutable. I suppose one could mutate the specs returned via pointers, and so we may want to store those as structs rather than pointers.
- Handlers can be generated via a spec (by providing a `frontend` image under a `target`), meaning we have to have the graph at the time the handlers are registered, in order to report accurately on which handlers are available.
- The specs that will be used need to be validated when they are read. In order to know which specs need to be validated, we need to know what specs will actually be used in the build sequence. Another option is to error out even if unused specs have errors.
- The final spec (the spec for the target, or before these changes, the only spec) is de facto global state. Once read and unmarshaled, the spec has the lifespan of the program as it drives all of the actions of the program and needs to be consulted throughout. It's therefore passed around as a pointed from start to finish. The graph technically doesn't need to live past the container generation stage (into the testing stage), but IMO that is splitting hairs.
- A value of `Mutex<T>` can be `const` (and therefore `static`, global, and mutable via mutex) in rust (without unsafe!), because its size and location are known at compile time, access to it is guarded by mutex, and as long as `T` can be initialized at compile time, the initial value is known as well.
- I should note that I agree with you about eliminating global state. I also agree with you on the implicit parameterization issue. I just see other concerns and I don't think the pattern is **that** bad, given the above.

These are reasons. Whether they are compelling or not is up to you :) I'm open to having my mind changed if you see ways around this and a benefit that outweighs the additional work. I also want explicit approval from @cpuguy83 before I change any public function signatures.

pmengelbert:
- [ ] Merge all the RPM states rather than looping through a spec's build deps. This will improve caching

pmengelbert:
- [ ] Remove the graph from global state.

graph.go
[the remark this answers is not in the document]
pmengelbert: Graph has been removed from global state

pmengelbert:
In addition to some other changes to the graph, I have made it (and the specs within it) immutable. Once initialized, it can't be changed.

PR #147, opened 2024-02-21
Allow fields starting with `x-` in the spec

cpuguy83:
This allows clients to store extra data in the spec without having to loosen up unmarshalling validations.

No "x-" field is explicitly added to the spec because frontends should not be interpreting this field.
Instead the fields are stripped from the input before unmarshalling to `Spec`.

### DIFF
--- a/load.go
+++ b/load.go
@@ -292,6 +292,12 @@ func (s *Spec) SubstituteArgs(env map[string]string) error {
 // LoadSpec loads a spec from the given data.
 func LoadSpec(dt []byte) (*Spec, error) {
 	var spec Spec
+
+	dt, err := stripXFields(dt)
+	if err != nil {
+		return nil, fmt.Errorf("error stripping x-fields: %w", err)
+	}
+
 	if err := yaml.UnmarshalWithOptions(dt, &spec, yaml.Strict()); err != nil {
 		return nil, fmt.Errorf("error unmarshalling spec: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


load.go
```diff
@@ -304,6 +310,21 @@ func LoadSpec(dt []byte) (*Spec, error) {
...
 }
 
+func stripXFields(dt []byte) ([]byte, error) {
+	var obj map[string]interface{}
+	if err := yaml.Unmarshal(dt, &obj); err != nil {
+		return nil, fmt.Errorf("error unmarshalling spec: %w", err)
+	}
+
+	for k := range obj {
+		if strings.HasPrefix(k, "x-") {
```
pmengelbert, line 320: allow uppercase X as well?

pmengelbert (approved):
LGTM, up to you if you want to allow `X-` in addition to `x-`.

load.go
```diff
@@ -304,6 +310,21 @@ func LoadSpec(dt []byte) (*Spec, error) {
 	return &spec, nil
 }
 
+func stripXFields(dt []byte) ([]byte, error) {
+	var obj map[string]interface{}
+	if err := yaml.Unmarshal(dt, &obj); err != nil {
+		return nil, fmt.Errorf("error unmarshalling spec: %w", err)
+	}
```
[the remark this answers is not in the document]
cpuguy83: Not as far as I can reason about.
Since we are unmarshalling to `map[string]interface{}` there shouldn't be anything dropped.

load.go
cpuguy83: Sure, that makes sense.

cpuguy83: Updated to include that.

PR #148, opened 2024-02-21
Remove GOROOT=/usr/lib/golang hack

### LINKED ISSUE #24: Fix issues with GOROOT in test fixtures
Once https://github.com/microsoft/CBL-Mariner/issues/6363 is fixed, we should stop explicitly setting GOROOT in our test fixtures.
e.g. https://github.com/Azure/dalec/commit/e78667c35ef1e7aa4e6cdded1074fa2d72b8a764#diff-71e44138c327abad97f620d808279e084836d17b70da19f02fb368bbb95bc3b9R50

```[tasklist]
### Tasks
- [ ] microsoft/cbl-mariner#6363
```

cpuguy83:
Supposedly this was fixed in the mariner go packages (though I've not verified that). In any case, I believe the main reason the bug in the packages was exposed was due to the way the mariner toolkit was executing the package builds.

This is no longer a problem for us, so removing these.

Closes #24

### DIFF
--- a/docs/examples/go-md2man-1.yml
+++ b/docs/examples/go-md2man-1.yml
@@ -19,7 +19,6 @@ dependencies:
 
 build:
   env:
-    GOROOT: /usr/lib/golang # Note: This is needed due to a bug in the golang package for mariner
     CGO_ENABLED: "0"
   steps:
     - command: |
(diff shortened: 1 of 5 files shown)


sozercan (approved):

pmengelbert (approved):
:+1:

PR #151, opened 2024-02-22
Add release workflow

cpuguy83:
When the repo is tagged (and that tag looks like a version number), the release workflow is triggered and will start a frontend build with the git tag (minus the leading `v`) as image tag.

Example (after a few tests) is here: https://github.com/cpuguy83/dalec-test-fork/actions/runs/8012326758
Which produced https://github.com/cpuguy83/dalec-test-fork/pkgs/container/dalec-test-fork%2Ffrontend/182474475?tag=0.1.5

image

### DIFF
--- a/.github/workflows/frontend-image.yml
+++ b/.github/workflows/frontend-image.yml
@@ -5,6 +5,11 @@ concurrency:
   cancel-in-progress: true
 
 on:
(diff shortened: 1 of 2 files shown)


sozercan (approved):
LGTM

PR #152, opened 2024-02-22
Remove some dependabot jobs

cpuguy83:
Remove some dependabot jobs
    
1. We don't have the image at frontend/mariner2 anymore, so that is not needed at all
2. For go deps, ignore major/minor version updates which can all be
    tricky for transative (or shared) upstream dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,12 +10,10 @@ updates:
     schedule:
       interval: weekly
 
-  - package-ecosystem: docker
-    directory: /frontend/mariner2
-    schedule:
-      interval: weekly
-
   - package-ecosystem: gomod
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major", "version-update:semver-minor"]


.github/dependabot.yml
```diff
@@ -9,13 +9,3 @@ updates:
...
-
-  - package-ecosystem: docker
-    directory: /frontend/mariner2
-    schedule:
-      interval: weekly
-
-  - package-ecosystem: gomod
-    directory: /
-    schedule:
-      interval: weekly
```
sozercan, line 21: for gomod, some things to consider:
- would changing the interval to monthly help?
- not a lot of value in merging these unless before release, can we leave the prs as is and use it as a signal especially for sec fixes as needed?
- we can also filter for patch releases if that helps

sozercan (approved):
Added some thoughts if it helps. I am not against removing if it's generating a lot of noise. Otherwise, LGTM

.github/dependabot.yml
cpuguy83: I've updated this, PTAL.

sozercan (approved):
LGTM

PR #161, opened 2024-02-24
Validate rpm fields

cpuguy83:
These fields are required to build an rpm.
When the fields are not there rpmbuild gives an error about the fields,
but those field names are different so its confusing to the user.

This checks those fields early so we can give a better error message.

### DIFF
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -659,7 +659,7 @@
 				},
 				"version": {
 					"type": "string",
-					"description": "Version setst he version of the package."
+					"description": "Version sets the version of the package."
 				},
 				"revision": {
 					"oneOf": [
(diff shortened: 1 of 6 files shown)


DannyBrito (approved):
lgtm

pmengelbert:
We could move this up the callstack and validate in `load.go`. I think most targets will need these fields, and I think it's just simpler to reject all specs that lack them regardless of the target.

pmengelbert:
> I don't think it is worth it, because several targets that we even have now don't need those fields. Just more hoops to mess around with things.
> 
> For instance, I don't need those fields to extract sources out, or for any of the debug targets.

Makes sense. IIRC the Jammy draft needs those fields too, so I'll make a comment there to ensure we get validation added.

pmengelbert (approved):

PR #164, opened 2024-02-26
feat: Implement Project struct

pmengelbert:
This is backward-compatible with the single-spec format.


cpuguy83:
Closing since this is no longer a priority and I think we plan to solve this in another way.

PR #165, opened 2024-02-27
Remove old "inline" field for build source

cpuguy83:
This can be completely replaced by using the inline source type.


pmengelbert (approved):
LGTM

cpuguy83:
As the test changes show, this is breaking a use-case. One that I think is useful (dockerfile is in a different state than the state for the build).

I plan on addressing that as a follow-up commit on this PR.

cpuguy83:
I changed my mind, just going to leave this PR as is and we can address the other use-case (having dockerfile as a separate context) elsewhere.

I've posted another branch with it on my fork but don't plan to open a PR for it at the moment just to keep the complexity down for now.

cpuguy83:
I did add an extra commit if you could give that a look @pmengelbert

source_build.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	if err := s.Source.validate("build subsource"); err != nil {
+		retErr = goerrors.Join(retErr, err)
+	}
+
+	return
+}
+
+func (src *SourceBuild) AsState(name string, _ *Source, sOpt SourceOpts, opts ...llb.ConstraintsOpt) (llb.State, error) {
```
adamperlin, line 49: I believe the `_ *Source` can be dropped here as there is no need for `AsState` to conform to a particular interface signature any more

adamperlin: This was an artifact of the refactor, I think

cpuguy83: Removed

adamperlin (approved):
Looks good to me as well

PR #170, opened 2024-03-08
Cue spec validation

### LINKED ISSUE #132: Experiment with Spec Validation Using Cue
We can model the Dalec Spec in Cue by creating a schema, and then use Cue to programmatically check to make sure that a user's spec YAML is correct. This might lead to better error messages and catch spec correctness issues earlier in the build process. 

This could also help with issues like #111, as Cue allows for regex specifications in Schemas. It might also reveal other parts of the spec that need more specification. I've started working on a proof of concept

adamperlin:
Currently, Dalec specs are validated using a mixture of strict YAML unmarshalling and manual validation of the unmarshaled spec in Go code.  This PR allows us to concretely specify a schema for Dalec specs using Cue, and then utilize Cue for almost all spec validation. This has a number of benefits:
1. It removes the need for some of the currently existing validation code within Dalec, within which it is hard to account for every error case.
2. It allows us to add new spec validations easily and declaratively, without changing Go validation code
 - As a proof of concept for this, the Cue spec included in this PR places additional restrictions on source names
 3. It gives us a model to use for generating templated YAML which is correct-by-construction. We can use the Cue spec included here to create templates which can be used for generating starter spec stubs for new Dalec users, as well as creating templates over subcomponents of the Dalec spec (i.e., reducing boilerplate for particular commonly used sources, and any number of other tasks). 

I am opening this as a draft mostly for discussion. All the current test cases in the repo are passing. I do have some concerns with user friendliness, as cue's error messages aren't always the friendliest and take a bit of getting used to. They do however nearly always point to the specific field (or lack of field) which is causing the issue.

Fixes #132 


test/fixtures/local-context.yml
```diff
@@ -60,7 +60,7 @@ build:
 
 changelog:
   -
-    date: 2023-10-10
+    date: 2023-10-10T00:00:00Z
```
adamperlin, line 63: It is a little annoying to have to specify dates this way. I'm going to look into whether this can be cleaner

test/fixtures/moby-runc.yml
```diff
@@ -67,7 +67,7 @@ sources:
 
 build:
   env:
-    CGO_ENABLED: 1
+    CGO_ENABLED: "1"
```
adamperlin, line 70: We do need to be explicit now about string vs. int values for env variables in the yaml itself

pmengelbert: That's a major gotcha in YAML. I think it's worth investigating if there's a way you can get CUE to interpret values in certain structures as strings. I'm not sure that it's possible but it may be.

cue/spec.cue
```diff
@@ -0,0 +1,258 @@
+package spec 
+
+import (
+    "time"
+)
+
+// TODO: consider adding defaults to more fields
```
pmengelbert, line 7: It is definitely worthwhile to capture all of the defaults in this PR. We actually have extensive defaults, mostly undocumented.

adamperlin:
Some further things to investigate:
- [ ] Scrape error messages from cue to provide more user-friendly and Dalec specific validation messages (even if just: `validation error in field x.y.z, expected values are: ...` I.e., we don't necessarily need to expose the user to cue-specific concepts like value unification

cpuguy83:
If we were to adopt this, I think we'd need to generate go types from the cue.

load.go
```diff
@@ -270,6 +277,25 @@ func (s *Spec) SubstituteArgs(env map[string]string) error {
...
+	v := cueCtx.CompileString(cueSpec)
+
+	if v.Err() != nil {
+		return v.Err()
+	}
+
+	_, err := pkgyaml.Validate(dt, v)
+	errs := (cueerrors.Errors(err))
+
+	var retErrs []error = make([]error, 0, len(errs))
```
MorrisLaw, line 291: nit:
```suggestion
	retErrs := make([]error, 0, len(errs))
```

load.go
```diff
@@ -270,6 +277,25 @@ func (s *Spec) SubstituteArgs(env map[string]string) error {
 	return nil
 }
 
+func CueValidate(dt []byte) error {
+	cueCtx := cuecontext.New()
+	v := cueCtx.CompileString(cueSpec)
```
MorrisLaw, line 282: nit:
```suggestion
	ctx := cuecontext.New()
	v := ctx.CompileString(cueSpec)
```

PR #172, opened 2024-03-13
Rewrite target routing logic

cpuguy83:
This breaks down routing logic into smaller pieces. Introduces a `BuildMux` upon which all the target routing is handled.

`BuildMux` implements a buildkit `BuildFunc` via its `Handle` method. With a `BuildMux` you register routes with `mux.Add("someKey", SomeHandler, optionalTargetInfo)`.

When a build request is made, `BuildMux` tries to match the requested build target to a registered handler via the registered route. The correct handler to use is determined with the following logic:

1. Build target is an exact match to a registered route
2. Build target is empty, check if a default handler is registered
3. Check if any of the registered handlers have a route that is a prefix match to the build target

`BuildMux` route handlers also must be buildkit `BuildFunc`s. As such a handler can itself also be a distinc `BuildMux` with its own set of routes.
This allows handlers to also do their own routing. All logic for *what* to route is handled outside of the `BuildMux`.

`BuildMux` and buildkit `BuildFunc` have a similar relationship as `http.ServeMux` and `http.Handler` (where `http.ServeMux` *is* an `http.Handler`).
In the same way `BuildMux` is a `BuildFunc` (or rather `BuildMux.Handle` is).
So `BuildMux` can be nested.

When `BuildMux` calls a handler, it modifies the client to chomp off the matched route prefix.
So a `BuildMux` with receiving a build target of `mariner2/container` will match on the registered handler for `mariner2` then call the handler with the build target changed to just `container`.

Finally, `BuildMux` sets an extra build option on the client `dalec.target=<matched prefix>`.
This is only done if the `dalec.target` option is not already set, so `dalec.target` is only modified once and then used in handlers to determine which target in `spec.Targets` applies to them.


pmengelbert:
Just noting here that we may want to account for cross-compilation in the structure of the routes themselves.

See this discussion on the Windows PR: https://github.com/Azure/dalec/pull/108#discussion_r1526625051

tldr: currently the first path component refers to both the build toolchain and the output (for example, `mariner2` is built on mariner and output as a mariner-based image). Going forward, we should account for situations where the two aren't the same. Brian has proposed the route structure `<build-toolchain>/cross/<output-image>`. In the case of cross-compiling from jammy to windows, this would by `jammy/cross/windows`.

frontend/build.go
```diff
@@ -88,75 +67,29 @@ func fillPlatformArgs(prefix string, args map[string]string, platform ocispecs.P
...
-
-	if err := registerSpecHandlers(ctx, spec, client); err != nil {
-		return nil, err
-	}
-
-	res, handled, err := bc.HandleSubrequest(ctx, makeRequestHandler(bc.Target))
-	if err != nil || handled {
-		return res, err
-	}
+type PlatformBuildFunc func(ctx context.Context, client gwclient.Client, platform *ocispecs.Platform, spec *dalec.Spec, targetKey string) (gwclient.Reference, *dalec.DockerImageSpec, error)
```
adamperlin, line 70: For some reason the name `targetKey` has been breaking my brain, maybe because I associate key so heavily with caching and/or table lookup. I wonder if we could change this to `targetRoute`, since this is so conceptually similar to an HTTP router?

frontend/mariner2/handle_container.go
```diff
@@ -98,20 +100,20 @@ func buildImageConfig(ctx context.Context, spec *dalec.Spec, target string, clie
...
 
 	if spec.Image != nil {
 		cfg = *spec.Image
 	}
 
-	if i := spec.Targets[target].Image; i != nil {
+	if i := spec.Targets[targetKey].Image; i != nil {
 		if i.Entrypoint != "" {
-			cfg.Entrypoint = spec.Targets[target].Image.Entrypoint
+			cfg.Entrypoint = spec.Targets[targetKey].Image.Entrypoint
```
adamperlin, line 112:
```suggestion
			cfg.Entrypoint = i.Entrypoint
```

frontend/mariner2/handle_container.go
```diff
@@ -98,20 +100,20 @@ func buildImageConfig(ctx context.Context, spec *dalec.Spec, target string, clie
...
 		}
 
 		if i.Cmd != "" {
-			cfg.Cmd = spec.Targets[target].Image.Cmd
+			cfg.Cmd = spec.Targets[targetKey].Image.Cmd
```
adamperlin, line 116:
```suggestion
			cfg.Cmd = i.Cmd
```

frontend/mux.go
```diff
@@ -0,0 +1,508 @@
...
+	// If this request was a request to list targets, we need to modify the response a bit
+	// Otherwise we can just return the result as is.
+	if opts[requestIDKey] == bktargets.SubrequestsTargetsDefinition.Name {
+		return m.fixupListResult(matched, res)
+	}
+	return res, nil
+}
+
+func (m *BuildMux) fixupListResult(matched string, res *gwclient.Result) (*gwclient.Result, error) {
+	// Update the targets to include the matched key in their path
```
adamperlin, line 319: Maybe move this into a comment before the function, and maybe add a little more detail on why the function is needed

frontend/build.go
cpuguy83: In this case "targetKey" is litterally referencing the map key in the `Targets` field on the spec.

frontend/mux.go
cpuguy83: Done.

frontend/mariner2/handle_container.go
cpuguy83: This looks to be resolved primarily by the Windows PR which moved these functions around.

frontend/build.go
```diff
@@ -88,75 +67,29 @@ func fillPlatformArgs(prefix string, args map[string]string, platform ocispecs.P
...
-		case dalecSubrequstForwardBuild:
-		case "":
-		default:
-			return nil, fmt.Errorf("unknown request id %q", requestID)
-		}
-	}
-
-	f, err := lookupHandler(bc.Target)
+// BuildWithPlatform is a helper function to build a spec with a given platform
+// It takes care of looping through each tarrget platform and executing the build with the platform args substituted in the spec.
```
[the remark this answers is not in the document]
cpuguy83: 🏴‍☠️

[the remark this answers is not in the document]
cpuguy83: Merged this suggestion in.

frontend/build.go
adamperlin: Yeah, I think I picked a poor place to make this comment, apologies -- I guess it's not really relevant in this particular spot.  I meant to refer to the use of `targetKey` as the name for routes in the `BuildMux`.  Would you be open to renaming this to something that reflects its treatment as a `route` or `path` in the `BuildMux` specifically?

frontend/mux.go
```diff
@@ -0,0 +1,515 @@
...
+}
+
+func maybeSetDalecTargetKey(client gwclient.Client, key string) gwclient.Client {
+	opts := client.BuildOpts()
+	if opts.Opts[keyTargetOpt] != "" {
+		// do nothing since this is already set
+		return client
+	}
+
+	// optimization to help prevent uneccessary grpc requests
```
adamperlin, line 150: maybe expand on this comment just a little bit to explain why it works?

frontend/mux.go
```diff
@@ -0,0 +1,515 @@
...
+
+// BuildMux implements a buildkit BuildFunc via its Handle method. With a
+// BuildMux you register routes with mux.Add("someKey", SomeHandler,
+// optionalTargetInfo).
+//
+// When a build request is made, BuildMux tries to match the requested build
+// target to a registered handler via the registered handler. The correct
+// handler to use is determined with the following logic:
+//
+// Build target is an exact match to a registered route Build target is empty,
```
pmengelbert, line 31: I think the formatting got mangled here. It's hard to understand what this is saying.

frontend/mux.go
```diff
@@ -0,0 +1,515 @@
...
+func (err *noSuchHandlerError) Error() string {
+	return fmt.Sprintf("no such handler for target %q: available targets: %s", err.Target, strings.Join(err.Available, ", "))
+}
+
+func (m *BuildMux) lookupTarget(ctx context.Context, target string) (matchedPattern string, _ *handler, _ error) {
+	// `target` is from `docker build --target=<target>`
+	// cases for `t` are as follows:
+	//    1. may have an exact match in the handlers (ideal)
+	//    2. may have a prefix match in the handlers, e.g. hander for `foo`, `target == "foo/bar"` (assume nested route)
+	// 	  3. No matching handler and `target == ""` and there is a default handler set (assume default handler)
```
pmengelbert, line 236: nit: formatting
nit2: match the order of the cases in this list and their order in the code

frontend/mux.go
```diff
@@ -0,0 +1,515 @@
...
+	matched, h, err := m.lookupTarget(ctx, t)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx = bklog.WithLogger(ctx, bklog.G(ctx).WithField("matched", matched))
+
+	// each call to `Handle` handles the next part of the target
+	// When we call the handler, we want to remove the part of the target that is being handled so the next handler can handle the next part
+	client = maybeSetDalecTargetKey(trimTargetOpt(client, matched), matched)
```
pmengelbert, line 302: nit: un-nest the call to `trimTargetOpt`. It's an important mutation of state and it's too easy to miss when it's nested.

frontend/windows/handler.go
```diff
@@ -1,24 +1,30 @@
...
 	"github.com/Azure/dalec/frontend"
-	"github.com/moby/buildkit/frontend/subrequests/targets"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+	bktargets "github.com/moby/buildkit/frontend/subrequests/targets"
 )
 
 const (
-	targetKey = "windowscross"
-	outputKey = "windows"
+	DefaultTargetKey = "windowscross"
```
pmengelbert, line 12: *note: no action needed at this time*

I am thinking ahead to when (or if?) we change the route to windows/cross/container (or possibly windows/container/cross). That is going to require some additional work in the `targets` part of the spec, because we will have one "dalec.target" for both cross-compiled and non-cross-compiled windows containers (both would have `windows` as the top-level path component).

There might be different requirements, like different build dependencies. What do we do about that?

frontend/mux.go
```diff
@@ -0,0 +1,515 @@
...
+
+		if len(spec.Targets) > 0 {
+			t, ok := spec.Targets[key]
+			if !ok {
+				bklog.G(ctx).WithField("spec targets", maps.Keys(spec.Targets)).WithField("targetKey", key).Info("Target not in the spec, skipping")
+				return nil
+			}
+
+			if t.Frontend != nil {
+				bklog.G(ctx).WithField("targetKey", key).Info("Target has custom frontend, skipping builtin-handler")
```
DannyBrito, line 508: there is the mention that has a custom frontend and is going to be skipped but only seems to be log it? does it need a return nil?

frontend/request.go
```diff
@@ -0,0 +1,118 @@
...
+
+		if spec != nil {
+			if req.FrontendOpt == nil {
+				req.FrontendOpt = make(map[string]string)
+			}
+			if spec.Target != "" {
+				req.FrontendOpt["target"] = spec.Target
+			}
+			for k, v := range spec.Args {
+				req.FrontendOpt["build-arg:"+k] = v
```
DannyBrito, line 108: is this `req.FrontendOpts` pointing to the build-args pass when calling `docker buildx build --build-arg "<key-vaule>"`? If so shouldn't this have greater precedence against the ones set on the spec `arg` block?
If that is the case we would need to do a conditional to prevent overwrite?

frontend/windows/handler.go
adamperlin: could we specify dependencies specifically for `windows/cross` in the targets portion of the spec?

pmengelbert: That would be difficult, because the client's `opts` is updated to have `dalec.target` set to the first path component. The rest of the target will have been gradually chomped away by the recursive calls to `mux.Handle`.

frontend/request.go
cpuguy83: At this point we've already loaded all the build args, including filling in the values from the client into the spec.
Additionally, these are args set on a build source, e.g.

```yaml
sources:
  someBuild:
    build:
      args:
        someKey: someval
```

frontend/build.go
cpuguy83: I changed this in the `Add` function, but everywhere else it seems like the right wording.

frontend/mux.go
cpuguy83: Fixed

frontend/mux.go
cpuguy83: Fixed.

frontend/mux.go
cpuguy83: Fixed. Good catch!

cpuguy83:
This should be all updated with requested items resolved.

frontend/windows/handler.go
adamperlin: Yeah, I see what you're saying. Trying to think about how this could be dealt with. One possible option is, in addition to `dalec.target`, we could keep an option that is updated with the current handler path. So, the handler registered at `windows/cross/container` would know the path that was taken to get to it. Then, when fetching dependencies for a target, we could first try the full handler path before falling back to `dalec.target`. Just an idea. It does kind of break the nested handler abstraction, though

frontend/gateway.go
```diff
@@ -237,51 +152,64 @@ func checkDiffMerge(client gwclient.Client) bool {
...
-		if _, ok := req.FrontendOpt[k]; ok {
-			// Do not overwrite existing opts
-			continue
-		}
-		req.FrontendOpt[k] = v
+		return nil
 	}
+}
+
+const keyTargetOpt = "dalec.target"
```
adamperlin, line 206: Can we change the name of this constant to `topLevelTarget` or something that really indicates it's supposed to be the top-level path component? I found it very easy to mix up `keyTargetOpt` and `keyTarget` when reading the code.

adamperlin (approved):
Just one more minor comment about the naming of `keyTarget*` variables

frontend/gateway.go
cpuguy83: That makes sense, done.

frontend/mux.go
```diff
@@ -0,0 +1,529 @@
...
+}
+
+type handler struct {
+	f gwclient.BuildFunc
+	t *bktargets.Target
+}
+
+// Add adds a handler for the given target
+// [targetKey] is the resource path to be handled
+func (m *BuildMux) Add(targePath string, bf gwclient.BuildFunc, info *bktargets.Target) {
```
adamperlin, line 70: nit: `targetPath`

adamperlin:
One tiny typo to fix

adamperlin (approved):
lgtm!

PR #175, opened 2024-03-14
add initial support for SOURCE_DATE_EPOCH

### LINKED ISSUE #54: Support SOURCE_DATE_EPOCH passthrough
SOURCE_DATE_EPOCH is a standard variable used for setting the timestamps of artifacts to a deterministic value.
This is used for reproducible builds.

Buildkit has some built-in support for it, e.g. layers and image creation dates are set to it.
Package build tooling (e.g. rpmbuild) also often supports passing this through.

1. We need to allow the build-arg to pass through our validation.
2. We should also automatically inject this as an env var to the package build step.

DannyBrito:
Adds initial support for `SOURCE_DATE_EPOCH`, technically this variable was already able to be injected in Dalec specs as one of the requirements to add this is to explicit require the `args/envs`.  It seems like `SOURCE_DATE_EPOCH` env var is auto injected by docker/buildx so no need to explicit `--build-arg`. The difference I could notice is that now with change applied the build will not fail as unknow arg for `SOURCE_DATE_EPOCH` var but it still requires the spec to set this variable to be used.
```yml
# syntax=ghcr.io/azure/dalec/frontend:latest
args:
  SOURCE_DATE_EPOCH: # [Required] Needs to explicit call for variable to be usable within spec

name: go-md2man
version: 2.0.3
packager: Dalec Example
...
sources:
...
build:
  env:
    SOURCE_DATE_EPOCH: ${SOURCE_DATE_EPOCH} # [Required] Needs to explicit set variable to be usable within build scope
  steps:
    - command: |
        set -x
        echo "${SOURCE_DATE_EPOCH} # If any of the required fields is missing value will be empty.
        exit 1
```

Fixes #54

### DIFF
--- a/load.go
+++ b/load.go
@@ -19,6 +19,8 @@ func knownArg(key string) bool {
 		return true
 	case "DALEC_DISABLE_DIFF_MERGE":
 		return true
+	case "SOURCE_DATE_EPOCH":
+		return true
 	}
 
 	return platformArg(key)


cpuguy83 (approved):
LGTM

AkihiroSuda:
> [Required] Needs to explicit call 

WDYT about defining `SOURCE_DATE_EPOCH` by default?
The `docker/dockerfile` frontend doesn't do that for compatibility with existing Dockerfiles though.

cpuguy83:
@AkihiroSuda I personally prefer the explicit declaration.

cpuguy83:
After giving this some more thought, I think it would be good to not require `SOURCE_DATE_EPOCH` to be explicitly set in the spec.

cpuguy83:
To clarify my reasoning here, if someone wants to build a spec, they shouldn't have to modify it to get the benefits of `SOURCE_DATE_EPOCH` in the build itself.

cpuguy83:
Opened https://github.com/Azure/dalec/issues/447

PR #176, opened 2024-03-16
Add CLI to build dalec specs directly for local dev

cpuguy83:
This gets around the problem of needing to have the default docker builder set with the current frontend already loaded.

The intent of this is for local iteration on dalec.

The CLI is very imperfect and only supports a subset of what buildx can do.
The one thing it does that buildx can't is it builds with the changes you currently have in your work tree (or at least current from when the binary was compiled).
So ideally use `go run ./cmd/localdev build <opts>` to make sure it is always current.

[asciicast](https://asciinema.org/a/t1D4oHSSpT163RnnHp3MHr0TF)


adamperlin:
@cpuguy83 it looks like this PR needs to be updated to account for the new router changes?

PR #187, opened 2024-03-21
Add documentation for sources

cpuguy83:


docs/sources.md
DannyBrito, line 1: I think it could be a good idea for the yamls snippets of each example. It would be good to add all the possible fields even if options and also, it's default values if omitted and give that as context.
Something like example for the git source

```yaml
sources:
  someSource1:
    git:
      # This uses an SSH style git URL.
      url: git@github.com:myOrg/myRepo.git
      commit: 1234567890abcdef
     keepGitDir: false # [Optional] - default value: true
  someSource2:
    git:
      # This uses an HTTPS style git URL.
      url: https://github.com/myOrg/myRepo.git
      commit: 1234567890abcdef
      keepGitDir: true # [Optional] - default value: true
```
wdyt?

docs/sources.md
```diff
@@ -0,0 +1,246 @@
...
+Ideally the content of a source is platform agnostic (e.g. no platform specific binaries).
+These sources are used to create source packages for the target platform, such as an srpm or a debian dsc.
+
+## Top-level source configuration
+
+For all source types, you can specify the following top-level configuration:
+
+- `path`: The path to extract from the source type
+- `includes`: A list of glob patterns to include from the source
+- `excludes`: A list of glob patterns to exclude from the source
```
DannyBrito, line 21: Could we maybe have some examples using these options or adding to existent examples?

docs/sources.md
```diff
@@ -0,0 +1,246 @@
...
+This is the equivelent of the following:
+
+```yaml
+sources:
+  someSource:
+    context:
+      name: "context"
+```
+
+Where `name: "context"`, not to be confused with the source type `context`, is named by convention by the docker CLI.
```
DannyBrito, line 98: This about context is a bit of confused maybe we could add a link to more info to contrast to build context in docker docs https://docs.docker.com/build/building/context/?

docs/sources.md
```diff
@@ -0,0 +1,246 @@
...
+Additionally contexts can be passed in from the docker cli: `docker build --build-context <name>=<path>`.
+The `<name>` would be the name to use in your yaml to access it.
```
DannyBrito, line 100: for this referring to the name of the context. Could we maybe add an example of how it could be used in the yaml?

docs/sources.md
```diff
@@ -0,0 +1,246 @@
...
+
+```yaml
+sources:
+  someDockerImage:
+    image:
+      ref: docker.io/library/alpine:3.14
+```
+
+You can also run commands in the image before fetching the contents.
+This is especially useful for generating content.
```
DannyBrito, line 160: Not sure I understand this. are you running the commands inside the image you are fetching? and what content are you retrieving?

docs/sources.md
```diff
@@ -0,0 +1,246 @@
...
+
+You can use the docker image source to produce any kind of content for your build.
+
+Docker image sources are considered to be "directory" sources.
+
+### Build
+
+Build sources allow you to build a dockerfile and use the resulting image as a source.
+It takes as input another source which must include the dockerfile to build.
+The default dockerfile path is `Dockerfile` just like a normal docker build.
```
DannyBrito, line 207: does this build source have a field to change the path to the dockerfile? In case these is not the default name `Dockerfile`?

docs/sources.md
DannyBrito: maybe, show a possible real use scenario?

DannyBrito (commented):
Maybe we could also point some of the fixture templates in the examples if relevant/applicable.

docs/sources.md
cpuguy83: Good idea, the examples I gave are not actually useful in the context of the spec.

docs/sources.md
cpuguy83: Done

docs/sources.md
cpuguy83: Done.

docs/sources.md
cpuguy83: Added an example.

docs/sources.md
cpuguy83: Well, there is no contrast with build context.
This is the build context.

cpuguy83: I did add an extra example and some extra wording here.

docs/sources.md
cpuguy83: The example is using a name, I included one that does not include the name.

cpuguy83:
> Maybe we could also point some of the fixture templates in the examples if relevant/applicable.

Do you mean anchors?
We could maybe add an example, but I'm not too keen on keeping yaml-specific docs here.

sozercan:
@cpuguy83 please add to sidebars too https://github.com/Azure/dalec/blob/main/website/sidebars.ts#L22

website/docs/intro.md
```diff
@@ -322,3 +322,4 @@ To re-itterate: the `x-` fields are ignored by the dalec parser and are only for
 
 * Details on editor support in [editor-support.md](editor-support.md)
 * More in-depth documentation for testing can be found in [testing.md](testing.md).
+* Explaination of sources and how to use them can be found in [sources.md](sources.md).
```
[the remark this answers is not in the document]
cpuguy83: Fixed

cpuguy83:
Added to sidebar

sozercan (approved):
LGTM

cpuguy83:
> > Maybe we could also point some of the fixture templates in the examples if relevant/applicable.
> 
> Do you mean anchors? We could maybe add an example, but I'm not too keen on keeping yaml-specific docs here.

@DannyBrito I was thinking we could add a tips/tricks section in the docs for more advanced use cases. WDYT?

website/docs/sources.md
```diff
@@ -0,0 +1,289 @@
...
+    git:
+      # This uses an SSH style git URL.
+      url: git@github.com:myOrg/myRepo.git
+      commit: 1234567890abcdef
+  someSource2:
+    git:
+      # This uses an HTTPS style git URL.
+      url: https://github.com/myOrg/myRepo.git
+      commit: 1234567890abcdef
+      # keepGitDir: true # Keep the .git directory when fetching the git source
```
DannyBrito, line 64: Something like this maybe so it gets picked up by syntax highlighter? 
```suggestion
    git:
      # This uses an HTTPS style git URL.
      url: https://github.com/myOrg/myRepo.git
      commit: 1234567890abcdef
      keepGitDir: true # [Optional] Keep the .git directory when fetching the git source. Default value: false
```

website/docs/sources.md
```diff
@@ -0,0 +1,289 @@
...
+      # No Digest verification
+      url: https://example.com/someFile.txt
+```
+
+The HTTP source type is considered to be a "file" source.
+
+### Build context
+
+Clients provide a build context to Dalec.
+As an  example, here is how the Docker client provides a build context to Dalec:
```
DannyBrito, line 91:
```suggestion
As an example, here is how the Docker client provides a build context to Dalec:
```

website/docs/sources.md
```diff
@@ -0,0 +1,289 @@
...
+The above example will fetch the git repo and build the dockerfile at the root of the repo.
+
+Here's another example using an inline source as the build source:
+
+```yaml
+  someBuild:
+    path: /hello.txt
+    build:
+      # dockerfile_path: other.Dockerfile # Change the default dockerfile path from `Dockerfile`
+      source:
```
DannyBrito, line 270: same here about syntax highlighter
```suggestion
    build:
      dockerfile_path: other.Dockerfile # [Optional] Change dockerfile path. Default value: "Dockerfile"
      source:
```

website/docs/sources.md
```diff
@@ -0,0 +1,289 @@
...
+      ref: docker.io/library/alpine:3.14
+      cmd:
+        dir: / # Default path that command steps are executed in
+        cache_dirs: null # map of cache mounts
+          # /foo: {
+          #   mode: shared # `locked` or `private` are the other options
+          #   key: myCacheKey
+          #   include_distro_key: false # Add the target key from the target being built into the cache key
+          #   include_arch_key: false # add the architecture of the image to run the command in into the cache key
+          # }
```
DannyBrito, line 206: It's a bit hard to see this, also it might just be me but not quite sure what this is used for. So maybe we could add a bit more context on its use or function?
```suggestion
        cache_dirs: # Map of cache mounts. Default value: `null`
          /foo: {
             mode: shared # The other options are `locked` or `private` 
             key: myCacheKey
             include_distro_key: false # Add the target key from the target being built into the cache key
             include_arch_key: false # add the architecture of the image to run the command in into the cache key
          }
```

DannyBrito:
> > > Maybe we could also point some of the fixture templates in the examples if relevant/applicable.
> > 
> > 
> > Do you mean anchors? We could maybe add an example, but I'm not too keen on keeping yaml-specific docs here.

No, I actually was referring to point out when applicable to the tests specs that live under `test/fixtures` example: `local-context.yml`.  Because sometimes these could be used as part and/or more examples, when we are talking about a specific source type or specific field option in these sources.



> > > Maybe we could also point some of the fixture templates in the examples if relevant/applicable.
> > 
> > 
> > Do you mean anchors? We could maybe add an example, but I'm not too keen on keeping yaml-specific docs here.
> 
> @DannyBrito I was thinking we could add a tips/tricks section in the docs for more advanced use cases. WDYT?

Yeah that sounds good, more examples and advance use cases are always good 😄

website/docs/sources.md
cpuguy83: Fixed.

website/docs/sources.md
cpuguy83: Fixed

website/docs/sources.md
cpuguy83: Fixed.

website/docs/sources.md
cpuguy83: Fixed.

cpuguy83:
@DannyBrito I've added a section linking to the test fixtures (with some warnings on usage).

DannyBrito (approved):
lgtm

Issue #185, opened 2024-03-21
[REQ] Allow patches to be a directory of patch files


cpuguy83:
### What kind of request is this?

Improvement of existing experience

### What is your request or suggestion?

If you have a bunch of patches, today you need to create a source for each individual patch file, and then reference that in the patches section of the spec.
This is a fairly poor experience.

We should be able to detect if the source is a directory or a file and then treat that accordingly.
Patch files would need to be lexicographically sorted in order to use the right order.

Alternatively we could do something similar to debbuild and look for a file called `series` with the sorted list of patch files.

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

PR #191, opened 2024-03-25
Implement FS.ReadDirFs to properly determine type of context source

### LINKED ISSUE #188: [BUG] Source paths re not checked to determine if the path is a dir or file
### Expected Behavior

We have some code that behaves differently if a source is expected to be a directory (e.g. like a git repo) or a file (e.g. something fetched from http).

However sources can be defined with subpaths, and those subpaths are never checked.
So the source could actually be just a file.

### Actual Behavior

A context or a git source is never treated as a file even if the path in the source that we are getting is a file.

### Steps To Reproduce

_No response_

DannyBrito:

The issue is that when trying to use a context source as a patch, it is currently treated as directory, so the file processed to a tar file, and when trying to apply the patch is not found under the sources as it is not handled.

This doesn't fix the context issue to be able to handle different for a file or directory.

These simple aims to allow context source to be used a patch.

I would appreciate so guidance/feedback on how to properly go about this issue.

Fixes #

Fixes #188


frontend/rpm/template.go
```diff
@@ -201,13 +201,17 @@ func (w *specWrapper) PrepareSources() (fmt.Stringer, error) {
 	for _, name := range keys {
 		src := w.Spec.Sources[name]
 		err := func(name string, src dalec.Source) error {
-			if patches[name] {
+			isDir := dalec.SourceIsDir(src)
```
cpuguy83, line 204: Instead of handing the "maybe it is a file in a dir" down below, it would be nice if `dalec.SourceIsDir` returned the actual truth.

To do this, we'll need to modify the `SourceOpt` type to include a field for interacting with the filesystem on an `llb.State`.
After we spoke about this last week I was even thinking we could provide an implementation of https://pkg.go.dev/io/fs#ReadDirFS for use in `SourceIsDir` and potentially other places.

docs/examples/kubernetes-patch.yml
```diff
@@ -0,0 +1,52 @@
+# syntax=local/dalec/frontend:latest
```
cpuguy83, line 1: These patch tests should use inline types and go into https://github.com/Azure/dalec/blob/ef6ac1e3168a3b52730f845d627cf068182d2795/test/mariner2_test.go#L22 and https://github.com/Azure/dalec/blob/ef6ac1e3168a3b52730f845d627cf068182d2795/test/windows_test.go#L21

cpuguy83 (changes requested):

frontend/rpm/template.go
adamperlin: I really like the `ReadDirFS` idea, and I opened an issue: #192

DannyBrito:
Thanks to @adamperlin, helping to implement ReadDirFs for Sources, we play a bit with the concept of using that to Figure out if a source should be treated as dir or not. However, it seems like this is way more complex than expected. So, I would like to gather feedback of how these should be handled.

While playing with how this should be handle is a bit tricky due to the flexibility to allow a source to include, exclude, and define a path of the source:
Here are some scenarios that I imagine we could possible encounter:
**PATH has greater presendece over includes and possible excludes?**
   Ni -> No included, Ai -> Any included
   Ne -> No excluded, Ae -> Any excluded
- IF path isn't empty and points to file [Ni,Ne]-> it should be a file
- IF path isn't empty and points to dir [Ni,Ne] -> it should be a dir
- IF path isn't empty and points to file [Ai,Ae] -> it should be a file it should have precedence? seems that if they include block is present it requires an exact match to the path to not break.
- IF path isn't empty and points to dir [Single Include to file, Ne] -> it should be a file if path is parent path of the include file, if include file is nested, it probably should be a dir.
- IF path isn't empty and points to dir [Single Include to dir, Ne] -> it should be a dir
- IF path isn't empty and points to dir [Multiple Includes, NE] -> it should be a dir
- IF path isn't empty and points to dir [Ni,Excludes all dirs] only one file remains in path as parent directory it should be a file, if remained file is in a nested path it should be a dir?
- IF path is empty [Multiple includes, Ne] it should be a dir
- IF path is empty [Single include not nested path, Ne] it should be a file
- IF path is empty [Single include nested path, Ne] it should be a dir

Should `path: ./`  field work and include everything, this might be broken.

Maybe, I am overthinking this, please any feedback would be great!

DannyBrito:
Other question is as we are using the FS implementation this needs to have a state. However, there is an issue we encounter with it as calling AsState for the source will also be calling to get the sources and these sources are getting filter accordingly. This affects to the original paths. Example if source has path `./to/the/source.txt` after filter and function the source would present at state with the `source.txt` file but this would be found at the `/` root level. So not encounter this problem of no longer valid paths, it was to invoke the `IsDir` right after getting sources and just before the filter within the AsState function and change returns to include a bool if isDIr. Would that be an ok approach? or any other suggestion would be apricated

cpuguy83:
So we are looking at this 1 source at a time.
If there are nested sources, we still only care about the source we are currently working with.

With filter/includes, this is implying a directory with one or more things in it.
I think we should treat that as a directory.

I think all we really need to address is if the `Path` is pointing at a file or a directory.

adamperlin:
I think the example we were thinking of was something like this (wanting to include a single patch file from build context). And the question is, what would be the ideal way to do this, and have the source be treated as just a file?
```yaml
 mypatch:
    path: ./patches/mypatch.patch
    includes:
       - ./patches/mypatch.patch
    context:
      name: context
```
In this case, we are using includes and excludes, but we really only want to include a single file. So I think regardless of whether there are includes or excludes, if the path points to a file in the original source state, we should treat it as a file? 

Slightly related (thought it doesn't solve the issue as a whole), for the patching case we could modify the patching code to always apply patches as a directory to support #185 with one or more files, and ensure any patches are pre-nested if they aren't already.

cpuguy83:
Correct, whatever `path` is pointing to should be the source of truth there.
The include/exclude is just optimization.

DannyBrito:
The CI is failing in two sections:
1. The Unit Test is failing on test: https://github.com/Azure/dalec/blob/main/source_test.go#L327 due to sOpts not really having available the GetFS function - that is used to determine if sources is a dir/file. The Issue here is that we cannot directly call to GetFS because worked implementation relies. on having the buildkit client, which in this case for the tests is not available. What would be the best to go about this issue?
2. The integration test: this is related to the windows tests and fails when gathering the sources and copy these to the windows base image, it is not really clear to me why this is failing, the local logs point out the base image path is not present https://github.com/Azure/dalec/blob/main/frontend/windows/handle_container.go#L60, for `windowsSystemDir`, when the copy is taking place. Note* This test is failing for me locally as well against `main` branch but not in the CI against main. I would appreciate some assistance debugging this.

cpuguy83:
> The Issue here is that we cannot directly call to GetFS

Can we provide a custom implementation for the test?
This is what we do for some other tests (that use other client functions).

source_test.go
```diff
@@ -749,10 +754,48 @@ func stubListener(t *testing.T) net.Addr {
 	return l.Addr()
 }
 
+type FakeFS struct {
+	entries []fs.DirEntry
+}
+
+type fakeRefEntry struct {
```
adamperlin, line 761: maybe `mockDirEntry`?

source_test.go
```diff
@@ -749,10 +754,48 @@ func stubListener(t *testing.T) net.Addr {
 	return l.Addr()
 }
 
+type FakeFS struct {
```
adamperlin, line 757: nit: I'd rename this to `MockReadDirFS`

source_test.go
```diff
@@ -778,6 +821,16 @@ func getSourceOp(ctx context.Context, t *testing.T, src Source) []*pb.Op {
 		return &st, nil
 	}
 
+	if len(fakeEntries) != 0 {
+		sOpt.GetFS = func(st llb.State) fs.ReadDirFS {
```
adamperlin, line 825: I think this should always be set, even if there are no entries. If there are no entries, `fs.entries` can just be an empty array

source_test.go
```diff
@@ -749,10 +754,48 @@ func stubListener(t *testing.T) net.Addr {
...
+
+func (st *FakeFS) ReadDir(name string) ([]fs.DirEntry, error) {
+	return st.entries, nil
+}
+
+func (st *FakeFS) Open(name string) (fs.File, error) {
+	return nil, nil
+}
+
+func NewFakeFS() *FakeFS {
```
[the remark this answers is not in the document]
DannyBrito: resolved

source_test.go
DannyBrito: yeah thanks forgot to rename

source_test.go
DannyBrito: yeah thanks forgot to rename

source_test.go
DannyBrito: sounds good

source_test.go
```diff
@@ -749,10 +754,48 @@ func stubListener(t *testing.T) net.Addr {
...
+func (fk *fakeRefEntry) Info() (fs.FileInfo, error) {
+	info := &fsutil.StatInfo{}
+	return info, nil
+}
+
+func (st *FakeFS) ReadDir(name string) ([]fs.DirEntry, error) {
+	return st.entries, nil
+}
+
+func (st *FakeFS) Open(name string) (fs.File, error) {
```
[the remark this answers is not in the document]
DannyBrito: done, Would it make sense to also do this for all other functions that we are not calling example for the mockDirEntry we have `Info`, `Type`, there are placeholders as of now

docs/examples/kubernetes-patch.yml
DannyBrito: > These patch tests should use inline types and go into
> 
> https://github.com/Azure/dalec/blob/ef6ac1e3168a3b52730f845d627cf068182d2795/test/mariner2_test.go#L22
> 
> and
> https://github.com/Azure/dalec/blob/ef6ac1e3168a3b52730f845d627cf068182d2795/test/windows_test.go#L21

I could move this template to there but to be honest more than a test fixture I was thinking of an example of how-to possible patch with dalec

source_test.go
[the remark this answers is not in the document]
adamperlin: You definitely could! You could also just fill in random defaults as well. I'd just avoid returning nil with no error

DannyBrito:
The current State on how we determine if a context source is a dir or file. Is after getting the source and pass it down the filter, we will check the FS on `/` in the result state. If we have a single entry in the `/` level and is a file, then it will be renamed to match source name and treated as a file. Otherwise is treated as a dir and not renamed

test/fs_test.go
```diff
@@ -0,0 +1,397 @@
...
+			assert.Equal(t, want.isDir, info.IsDir())
+		}
+
+		return rfs.Res()
+	})
+
+}
+
+func TestStateWrapper_ReadDir_HTTPSource(t *testing.T) {
+	url := "https://patch-diff.githubusercontent.com/raw/kubernetes/kubernetes/pull/120134.patch"
```
adamperlin, line 120: I'd prefer to not rely on links/git urls in these unit tests if we can avoid it. Are the http and git tests testing different behavior in the FS itself? Because if not, I'm not sure if we need these, since we already have some integration tests where this kind of code would be exercised

source.go
```diff
@@ -288,12 +323,31 @@ func WithCreateDestPath() llb.CopyOption {
...
+func contextSourceIsDir(srcState llb.State, sOpt SourceOpts) bool {
+	srcFS := sOpt.GetFS(srcState)
+	entries, err := srcFS.ReadDir("/")
+	if err != nil {
+		panic(err)
+	}
+	if len(entries) == 1 {
+		return entries[0].IsDir()
+	} else if len(entries) == 0 {
+		panic("shouldn't be the case") // Maybe this could raise a warning instead
```
adamperlin, line 339: I don't think we should panic here. If someone includes an empty context, it's probably an accident, but we should not assume anything. We could definitely give the user a warning if they include empty sources generally, but I'm not sure this is the place for it (perhaps we could add another piece of validation after we've generated the llb for all sources)

fs.go
```diff
@@ -0,0 +1,197 @@
...
+	s       llb.State
+	ctx     context.Context
+	opts    []llb.ConstraintsOpt
+	client  gwclient.Client
+	res     *gwclient.Result
+	o       sync.Once
+	initErr error
+}
+
+func NewStateRefFS(s llb.State, ctx context.Context, client gwclient.Client) *StateRefFS {
```
cpuguy83, line 31: Can you move this into the `frontend` package/
In general I try to keep everything that requires the gateway client and other such types out of the main dalec package.

fs.go
```diff
@@ -0,0 +1,197 @@
...
+		res, err := fs.fetchRef()
+		fs.res = res
+		fs.initErr = err
+	})
+
+	return fs.res, fs.initErr
+}
+
+func (fs *StateRefFS) Ref() (gwclient.Reference, error) {
+	res, err := fs.Res()
```
cpuguy83, line 66: We should be able to cache this.

fs.go
```diff
@@ -0,0 +1,197 @@
...
+		Definition: def.ToPB(),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return res, nil
+}
+
+func (fs *StateRefFS) Res() (*gwclient.Result, error) {
```
cpuguy83, line 55: Doesn't seem like this should be exported.

fs.go
```diff
@@ -0,0 +1,197 @@
...
+	fs.o.Do(func() {
+		res, err := fs.fetchRef()
+		fs.res = res
+		fs.initErr = err
+	})
+
+	return fs.res, fs.initErr
+}
+
+func (fs *StateRefFS) Ref() (gwclient.Reference, error) {
```
cpuguy83, line 65: I don't think we should need to export this function.

fs.go
```diff
@@ -0,0 +1,197 @@
...
+	stat   *types.Stat
+	offset int64
+}
+
+// close is a no-op
+func (s *stateRefFile) Close() error {
+	return nil
+}
+
+func (s *stateRefFile) Read(b []byte) (int, error) {
```
cpuguy83, line 139: `stateRefFile` should be able to implement `io.ReaderAt`, and then `Read` can be implemented on top of that.

source.go
```diff
@@ -288,12 +323,31 @@ func WithCreateDestPath() llb.CopyOption {
...
+	}
+	return true
+}
+
+func SourceIsDir(src Source, srcState llb.State, sOpt SourceOpts) bool {
 	switch {
+	case src.Context != nil:
+		return contextSourceIsDir(srcState, sOpt)
```
cpuguy83, line 347: I don't think we should treat context as special.
Also we should only need to inspect the state if `src.Path` is set to anything other than `""` or `"/"`.

Inspecting the state is a pretty expensive call.

test/fs_test.go
```diff
@@ -0,0 +1,397 @@
...
+	"github.com/Azure/dalec/frontend"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/moby/buildkit/client/llb"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+)
+
+func TestStateWrapper_OpenInvalidPath(t *testing.T) {
+	st := llb.Scratch().File(llb.Mkfile("/bar", 0644, []byte("hello world")))
+	testEnv.RunTest(context.Background(), t, func(ctx context.Context, gwc gwclient.Client) (*gwclient.Result, error) {
```
cpuguy83, line 22: I may have messed up some tests, but the should use `baseCtx` rather than` context.Background()`

cpuguy83 (changes requested):

fs.go
DannyBrito: Res seems like is used on the `testEnv.RunTest` that expects a  `func(ctx context.Context, c gwclient.Client) (*gwclient.Result, error)`

fs.go
DannyBrito: resolved

source.go
cpuguy83: I'd say if `len(entries) == 0` then this is a directory.

fs.go
adamperlin: it should be cached already. The initialization is in a sync.Once, so all future calls after the first one should simply return `fs.res` without doing the initialization

source.go
DannyBrito: ok, it is adapted now to only check if the path Is not root and same for the rename.

fs.go
adamperlin: This is done!

test/fixtures/local-context.yml
```diff
@@ -31,7 +31,13 @@ sources:
       - go.*
     excludes:
       - go.sum
-
+  src6:
```
adamperlin, line 34: Are these sources actually being used?

DannyBrito: thanks, not checking this as part of the build stage

source.go
```diff
@@ -303,12 +338,28 @@ func WithCreateDestPath() llb.CopyOption {
...
 
-func SourceIsDir(src Source) bool {
+func (s *Source) IsDir(srcState llb.State, sOpt SourceOpts) bool {
+	return SourceIsDir(*s, srcState, sOpt)
+}
+
+func SourceIsDir(src Source, srcState llb.State, sOpt SourceOpts) bool {
 	switch {
+	case src.Context != nil:
+		if !isRoot(src.Path) {
```
cpuguy83, line 348: All this logic needs to apply for *all* source types (which are normally considered a dir).

source.go
```diff
@@ -303,12 +338,28 @@ func WithCreateDestPath() llb.CopyOption {
...
+			srcFS := sOpt.GetFS(srcState)
+			entries, err := srcFS.ReadDir("/")
```
cpuguy83, line 350: The number of entries is not what matters here, only if `src.Path` is pointing to a file or a dir.

cpuguy83 (changes requested):

source.go
DannyBrito: the state here already represents the merged state so the `src.Path` is not longer valid and we wanted to check the `/` root

cpuguy83: In that case its not really a valid test because `Path` can point to a directory with a single file in it.
So we need to check the origin state before its fully filtered down to a path.

PR #203, opened 2024-04-09
Fix mounting file into image source

cpuguy83:
The old behavior was odd in this case where it could mount the file into a subdirectory of the path we requested (not desirable) and it was named after the parent source.
E.g. if you have an image source "foo" and you specify a file-based source at mount point '/some_file', it would mount this file as "/some_file/foo".

Related to the comment in this gist: https://gist.github.com/cpuguy83/c570f3ea97af215af63264061cac94af#file-spec-yaml-L114

### DIFF
--- a/source.go
+++ b/source.go
@@ -116,7 +116,7 @@ func (src *SourceDockerImage) AsState(name string, path string, sOpt SourceOpts,
 		return st, nil
 	}
 
-	st, err := generateSourceFromImage(name, st, src.Cmd, sOpt, path, opts...)
+	st, err := generateSourceFromImage(st, src.Cmd, sOpt, path, opts...)
 	if err != nil {
 		return llb.Scratch(), err
 	}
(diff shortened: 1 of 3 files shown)


source.go
```diff
@@ -195,7 +195,7 @@ func generateSourceFromImage(name string, st llb.State, cmd *Command, sOpts Sour
 	baseRunOpts := []llb.RunOption{CacheDirsToRunOpt(cmd.CacheDirs, "", "")}
 
 	for _, src := range cmd.Mounts {
-		srcSt, err := src.Spec.AsMount(name, sOpts, opts...)
```
adamperlin, line 198: I think this may have been a bug from the refactor

adamperlin (approved):
This looks good to me. Fairly straightforward fix

source.go
cpuguy83: 🤷 It happens, now we have a test case for it 😆

PR #204, opened 2024-04-09
Add support for validating HTTP by digest

### LINKED ISSUE #64: HTTP sources bust the cache
When a source uses http(s) the cache appears to be busted every time.
My guess here is due to the file getting the current timestamp (since we can't really cache an http source like we can a git ref).

Assuming this is the case, we should probably set some default value for timestamps on these sources.
We should also implement #54 so that this timestamp can be customized.

Additionally, buildkit does let you pass through a digest for http sources, we should support this. I do not know if buildkit will check its own cache for this digest before hitting the http url but if so this would also help with cache busts while also being more secure.

cpuguy83:
This is needed to ensure the integrity of http sources.

Closes #64 (or at least to the best we can possibly handle in dalec).

### DIFF
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -547,7 +547,12 @@
 		"SourceHTTP": {
 			"properties": {
 				"url": {
-					"type": "string"
+					"type": "string",
+					"description": "URL is the URL to download the file from."
+				},
+				"digest": {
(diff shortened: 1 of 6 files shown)


adamperlin (approved):
This lgtm

PR #205, opened 2024-04-10
Add support for generating go module deps

### LINKED ISSUE #180: [REQ] Built-in Support for golang sources

New feature

## Background

Today all the source types we support have direct `llb` equivalents.
Given that we support generating sources from a container build we can effectively support any kind of source.

However, we've found that while it is possible to support every kind of source, it can be tedious, error prone, and cause a lot of duplication for when using go modules.
Truly this is going to be the case for any language that has its own package management for resolving dependencies.
I am specifically targeting go modules because we have real-world usage of these right now and understand the pain points well.

## Problem

The crux of the issue for go modules is the need to download module dependencies as another source, since we disable network access during the build phase.
Today that means we end up with specs like this:

```yaml
sources:
  myProject:
    git: # the actual go module
      url: https://github.com/cpuguy83/myproject.git
      commit: deadbeef
  gomods: # Fetch the go module dependencies
    path: /build/cache
    image:
      ref: golang:1.21
      cmd:
        dir: /build/src
        mounts:
          - dest: /buildsrc
            spec:
              git:
                url: https://github.com/cpuguy83/myproject.git
                commit: deadbeef
        steps:
          - command: go mod download
            env:
              GOMODCACHE: /build/cache

build:
  steps:
    - command: |
        export GOMODCACHE="$(pwd)/gomods"

        # Build the project
```

Some of the duplication can be cleaned up by using yaml anchors, but that's only really taking care of the smallest chunk of code in that block.
Also most people would have to look up the right syntax to use a yaml anchor every time they make one, which is not a great experience to force on folks (nice that it exists but we should be able to do better).

If you have multiple go modules in a build, now you need to manage all of those dependencies, and the module directory for that.

## Proposal

Add a source to for go modules.
The go struct would look something like this:

```go
type SourceGomod struct {
	// From is how to fetch the origin source
	// This is optional and you may opt instead to provide just the module name
	From *Source `json:"from,omitempty" yaml:"from,omitempty"`

	// Name is the name of the full name of the module that go would use to fetch the module.
	// Either `name` OR `FROM` must be specified, but not both.
	// e.g. "github.com/Azure/daelc@v0.1.0"
	Name string `json:"name,omitempty" yaml:"name,omitempty"`
}
```

In order to fetch go modules, a target handler (eg. `mariner2`) needs to provide an image with `go` (and minimally `git`) in `$PATH`
Dalec will manage the go module directory.
If multiple gomod sources are provided, a single go modules directory will be mounted into each fetch so there is only a single go module.

During the build phase we will inject the modules into `/go/pkg/mod`.
Presumably the target handler can inject a custom path in as needed.

The user will still need to specify go as a build dependency, or otherwise make sure go is available during the build phase.

## Known issues not resolved

One thing not accounted for in this proposal (and also not handled by any other means), is sometimes someone would want to patch `go.mod` before fetching modules.
Since we patch during the build phase (or at least this is what we would do for typical packaging builds) this presents a problem.
I'm not sure we need to solve that problem with this change since its not a new problem.

Some thing we've thrown around as an idea to solving this is to apply patches twice.
First we patch while pulling modules, but then discard the diff.
Then patch again during build.
Alternatively we could just allow people to specify patches that get applied before build and annotate that in packaging specs (where applicable).

cpuguy83:
This adds a new `Generate` field on `Source` which is a list of
generator configs.
Today the only type of generator we support is for go modules but we'll
likely need to add support for other things, e.g. Rust cargo deps.

Multiple geneator configs are allowed in the case of a source with
multiple go modules defined in it.

Closes #180

### DIFF
--- a/frontend/rpm/template.go
+++ b/frontend/rpm/template.go
@@ -7,10 +7,13 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 	"text/template"
 
 	"github.com/Azure/dalec"
 )
+
+const gomodsName = "__gomods"
 
 var specTmpl = template.Must(template.New("spec").Parse(strings.TrimSpace(`
 Summary: {{.Description}}
@@ -169,6 +172,11 @@ func (w *specWrapper) Sources() (fmt.Stringer, error) {
 		}
 		fmt.Fprintf(b, "Source%d: %s\n", idx, ref)
 	}
+
+	if w.Spec.HasGomods() {
+		fmt.Fprintf(b, "Source%d: %s.tar.gz\n", len(keys), gomodsName)
+	}
+
 	return b, nil
 }
 
@@ -198,8 +206,17 @@ func (w *specWrapper) PrepareSources() (fmt.Stringer, error) {
 	// Sort keys for consistent output
 	keys := dalec.SortMapKeys(w.Spec.Sources)
 
+	prepareGomods := sync.OnceFunc(func() {
+		if !w.Spec.HasGomods() {
+			return
+		}
+		fmt.Fprintf(b, "mkdir -p \"%%{_builddir}/%s\"\n", gomodsName)
+		fmt.Fprintf(b, "tar -C \"%%{_builddir}/%s\" -xzf \"%%{_sourcedir}/%s.tar.gz\"\n", gomodsName, gomodsName)
+	})
+
 	for _, name := range keys {
 		src := w.Spec.Sources[name]
+
 		err := func(name string, src dalec.Source) error {
 			if patches[name] {
 				// This source is a patch so we don't need to set anything up
@@ -219,6 +236,8 @@ func (w *specWrapper) PrepareSources() (fmt.Stringer, error) {
 			for _, patch := range w.Spec.Patches[name] {
 				fmt.Fprintf(b, "patch -d %q -p%d -s < \"%%{_sourcedir}/%s\"\n", name, *patch.Strip, patch.Source)
 			}
+
+			prepareGomods()
 			return nil
 		}(name, src)
 		if err != nil {
@@ -251,6 +270,10 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
 	fmt.Fprintf(b, "%%build\n")
 
 	fmt.Fprintln(b, "set -e")
+
+	if w.Spec.HasGomods() {
+		fmt.Fprintln(b, "export GOMODCACHE=\"$(pwd)/"+gomodsName+"\"")
+	}
 
 	envKeys := dalec.SortMapKeys(t.Env)
 	for _, k := range envKeys {
--- a/generator_gomod.go
+++ b/generator_gomod.go
@@ -1,0 +1,93 @@
+package dalec
+
+import (
+	"github.com/moby/buildkit/client/llb"
+	"github.com/pkg/errors"
+)
+
+const (
+	gomodCacheDir = "/go/pkg/mod"
+)
+
+func (s *Source) isGomod() bool {
+	for _, gen := range s.Generate {
+		if gen.Gomod != nil {
+			return true
+		}
+	}
+	return false
+}
+
+// HasGomods returns true if any of the sources in the spec are a go module.
+func (s *Spec) HasGomods() bool {
+	for _, src := range s.Sources {
+		if src.isGomod() {
+			return true
+		}
+	}
+	return false
+}
+
+func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.ConstraintsOpt) func(llb.State) llb.State {
+	return func(in llb.State) llb.State {
+		const workDir = "/work/src"
+		var srcMount llb.RunOption
+		if g.Subpath != "" {
+			srcMount = llb.AddMount(workDir, srcSt, llb.SourcePath(g.Subpath))
+		} else {
+			srcMount = llb.AddMount(workDir, srcSt)
+		}
+		return worker.Run(
+			shArgs("go mod download"),
+			llb.AddEnv("GOMODCACHE", gomodCacheDir),
+			llb.Dir(workDir),
+			srcMount,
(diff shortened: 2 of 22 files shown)


frontend/rpm/template.go
```diff
@@ -169,6 +177,12 @@ func (w *specWrapper) Sources() (fmt.Stringer, error) {
 		}
 		fmt.Fprintf(b, "Source%d: %s\n", idx, ref)
 	}
+
```
adamperlin, line 180: shouldn't `maxID` always be `len(keys)-1`?

cpuguy83: Yep, seems so.
Silly me.

frontend/rpm/template.go
```diff
@@ -252,6 +277,10 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
 
 	fmt.Fprintln(b, "set -e")
 
+	if w.Spec.HasGomods() {
```
adamperlin, line 280: I'm not sure if this needs to be addressed in this PR since there is only one generator right now, but maybe if we add more generators in the future, it would be worth thinking about an interface for the common actions all generators have to take.  I.e., we could add hooks to allow generators to register handlers for verifying the spec to make sure the generator can be applied, generating extra LLB for the worker state, modifying the RPM spec as it's being built, etc. Just an idea

cpuguy83: Yep, definitely on my radar.

frontend/rpm/template.go
cpuguy83: Fixed.

cpuguy83:
I've updated this with a slightly different implementation that is more extensible and easier to discover.
Also marked this ready for review.

generator_gomod.go
```diff
@@ -0,0 +1,102 @@
+package dalec
+
+import (
+	"github.com/moby/buildkit/client/llb"
+	"github.com/pkg/errors"
+)
+
+const (
+	gomodCacheDir  = "/go/pkg/mod"
+	gomodGenerator = "gomod"
```
DannyBrito, line 10: not sure if I missed it but is this const `gomodGenerator` used?

test/fixtures/frontend.yml
```diff
@@ -37,10 +25,9 @@ build:
     CGO_ENABLED: "0"
     GOGC: off
     GOFLAGS: -trimpath
-    GOPATH: /go
+    # GOPATH: /go
```
DannyBrito, line 28: Is this intended to be a comment?

generator_gomod.go
```diff
@@ -0,0 +1,102 @@
...
+	sorted := SortMapKeys(s.Sources)
+	for _, key := range sorted {
+		src := s.Sources[key]
+		if !src.isGomod() {
+			continue
+		}
+
+		opts := append(opts, ProgressGroup("Fetch go module dependencies for source: "+key))
+		hasGomod = true
+		withMods, err := WithModuleDeps(worker, sOpt, &src, opts...)
```
DannyBrito, line 86: I am a bit confused on how this works, so this whole PR would allow for any specified source to have in this case a gomod generator. So for a scenario that we have at least 2 sources that have gomod gen. When we retrieve these gomods do these get merged into a single directory and state for all the gomods for the different sources? or each gomod would have it's own state? The other relevant section for this question is when preparing the sources in the rpm template: `frontend/rpm/template.go`
```go
	prepareGomods := sync.OnceFunc(func() {
		if !w.Spec.HasGomods() {
			return
		}
		fmt.Fprintf(b, "mkdir -p \"%%{_builddir}/%s\"\n", gomodsName)
		fmt.Fprintf(b, "tar -C \"%%{_builddir}/%s\" -xzf \"%%{_sourcedir}/%s.tar.gz\"\n", gomodsName, gomodsName)
	})
```
the `gomodsName` is a constant so it doesn't make sense to do this more than once if the gomod modules are merged into one for all sources.

test/fixtures/frontend.yml
cpuguy83: Hmm, I think I changed this for debugging purposes and never changed it back.

generator_gomod.go
cpuguy83: > When we retrieve these gomods do these get merged into a single directory and state for all the gomods for the different sources?

Yes, that is what this function is doing.
The modules are stored in the state referenced by the `deps` variable, which is getting overwritten below (L90) with `deps = deps.With(withMods)`, where `withMods` is an `llb.StateOption` that modifies the state to include the deps for an individual source.

The trick there is in the `withGomod` function where it is using return value of `llb.Run(...).AddMount(gomodCachedir, in)`.
Here the return value of `AddMount` is the the content of the mount point after the `Run` has executed.

test/fixtures/frontend.yml
cpuguy83: Fixed this.

generator_gomod.go
cpuguy83: Done.

test/helpers_test.go
```diff
@@ -161,7 +161,7 @@ func (d dirStatAsStringer) String() string {
 type srOpt func(*gwclient.SolveRequest)
 
 func newSolveRequest(opts ...srOpt) gwclient.SolveRequest {
-	sr := gwclient.SolveRequest{}
+	sr := gwclient.SolveRequest{Evaluate: true}
```
adamperlin, line 164: Why was the `Evaluate` flag changed here?

cpuguy83: It really needs to be default because it can easily be missed and cause a test to pass unexpectedly.
`Evaluate` unlazy's the request, basically.

If we do a solve and don't interact with the returned state or export it somewhere, etc, it doesn't actually do the operations.

adamperlin: Ah got it, that makes sense. Thanks for clarifying

adamperlin (approved):
I think this looks good! It would be interesting to see how this could possibly be slotted into an interface as described in #213

PR #206, opened 2024-04-12
Fix subrouting when parent path has match

cpuguy83:
Before this change, when the BuildMux's prefix route matching is used it would itterate the map of routes and use the first prefix match it finds.
Map ordering is non-deterministic so there was only a random chance that it would pick the closest match.

When itterating the map it now keeps a list of all possible candidates, then sorts those to select the best possible prefix match. The best candidate is the one with the longest set of matching characters.

As an example if we have two routes:

- `a`
- `a/other`

And a requested route of `a/other/1`

It would have a random chance of hitting `a/other` vs `a`, even though we *always* want `a/other` in this case.

### DIFF
--- a/frontend/mux_test.go
+++ b/frontend/mux_test.go
@@ -81,7 +81,7 @@ func TestBuildMux(t *testing.T) {
 	}
 	expectedSubrouteACount++
 
-	// Could should not have changed from above
+	// Count should not have changed from above
 	if count := realCount(); count != 2 {
 		t.Errorf("expected real handler to be called twice, got %d", count)
 	}
--- a/frontend/mux_test.go
+++ b/frontend/mux_test.go
@@ -1,0 +1,164 @@
+package frontend
+
+import (
+	"context"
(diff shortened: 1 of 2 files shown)


frontend/mux_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	client = newStubClient(withStubOptTarget("real/subroute/a"))
+	_, err = mux.Handle(ctx, client)
+	if err != nil {
+		t.Fatal(err)
+	}
+	expectedSubrouteACount++
+
+	// Could should not have changed from above
```
adamperlin, line 84: nit: `count`

adamperlin (approved):
This looks good to me. I'm glad it is a straightforward fix

frontend/mux_test.go
cpuguy83:
```suggestion
	// Count should not have changed from above
```

PR #214, opened 2024-04-19
Implement Package Signing

pmengelbert:

Fixes #


cmd/signer/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+ENV GOOS=${TARGETOS} GOARCH=${TARGETARCH} GOFLAGS=${GOFLAGS}
+RUN \
+    --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    go build -o /signer ./cmd/signer
+
+FROM mcr.microsoft.com/azure-cli:cbl-mariner2.0 as az-base
+WORKDIR /root
+COPY ./tmp/xsignextension-0.46-py3-none-any.whl .
```
pmengelbert, line 19: Obviously this won't work. We'll have to get the extension from a keyvault using secrets for this build..

test/fixtures/phony.go
```diff
@@ -30,7 +30,7 @@ func PhonyFrontend(ctx context.Context, gwc gwclient.Client) (*gwclient.Result,
 
 	st := llb.Image("golang:1.21", llb.WithMetaResolver(gwc)).
 		Run(
-			llb.Args([]string{"go", "build", "-o=/build/out", "./test/fixtures/phony"}),
+			llb.Args([]string{"go", "build", "-gcflags=all=-N -l", "-o=/build/out", "./test/fixtures/phony"}),
```
cpuguy83, line 33: Why the flags change?

test/testenv/buildx.go
```diff
@@ -189,6 +190,7 @@ func (b *BuildxEnv) bootstrap(ctx context.Context) (retErr error) {
 		}
 
 		c, err := client.New(ctx, "", client.WithContextDialer(func(ctx context.Context, addr string) (net.Conn, error) {
+			conn2.SetDeadline(time.Now().Add(time.Hour * 24))
```
cpuguy83, line 193: Why?

test/testenv/builld.go
```diff
@@ -46,8 +46,10 @@ func buildBaseFrontend(ctx context.Context, c gwclient.Client) (*gwclient.Result
 
 	defPB := def.ToPB()
 	return c.Solve(ctx, gwclient.SolveRequest{
-		Frontend:    "dockerfile.v0",
-		FrontendOpt: map[string]string{},
+		Frontend: "dockerfile.v0",
+		FrontendOpt: map[string]string{
+			"build-arg:GOFLAGS": "\"-gcflags=all=-N -l\"",
```
cpuguy83, line 51: Why?

frontend/mariner2/handle_rpm.go
```diff
@@ -48,6 +50,16 @@ func handleRPM(ctx context.Context, client gwclient.Client) (*gwclient.Result, e
 			return nil, nil, err
 		}
 
+		t := spec.Targets[targetKey]
+		if frontend.HasSigner(&t) {
+			signed, err := frontend.ForwardToSigner(ctx, client, platform, t.PackageConfig.Signer, st, "*.rpm")
```
cpuguy83, line 55: I don't think we can put "*.rpm" in here generically.
Also NOt sure this is explicitrly correct since the state should be a directory tree with `RPMS` and `SRPMS` in it.

cpuguy83: Instead this should probably be passed as part of the frontend config in the cmd line field.

frontend/request.go
```diff
@@ -116,3 +119,74 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
...
+		m[k] = def.ToPB()
+	}
+	req.FrontendInputs = m
+
+	stateDef, err := s.Marshal(ctx)
+	if err != nil {
+		return llb.Scratch(), err
+	}
+
+	req.FrontendOpt[contextKey] = compound(inputKey, id)
```
cpuguy83, line 168: This should be the context.

frontend/request.go
```diff
@@ -116,3 +119,74 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
...
+	req.FrontendInputs[id] = stateDef.ToPB()
+	req.FrontendOpt["dalec.target"] = opts["dalec.target"]
```
cpuguy83, line 170: The signer shouldn't need to be aware of `dalec.target`
Whatever arguments need to be passed should be done in the frontend config in the spec.

frontend/request.go
```diff
@@ -116,3 +119,74 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
...
+	req.FrontendOpt["find.pattern"] = filePattern
```
cpuguy83, line 171: Same here, this should be in the spec as part of the arguments to the frontend.

cpuguy83 (changes requested):

cmd/signer/Dockerfile
jeremyrickard: the extension itself doesn't need to be in a key vault does it? We could fetch this from storage?

test/fixtures/phony.go
pmengelbert: Left over from troubleshooting.

test/testenv/builld.go
pmengelbert: ditto

frontend/request.go
pmengelbert: It is, isn't it? `contextKey`  is `"context"`

frontend/mariner2/handle_rpm.go
pmengelbert: Do we not sign the SRPMs?

cmd/signer/main.go
```diff
@@ -0,0 +1,165 @@
...
+	ToolName         string        `json:"toolName"`
+	ToolVersion      string        `json:"toolVersion"`
+}
+
+type ParameterKV struct {
+	ParameterName  string `json:"parameterName"`
+	ParameterValue string `json:"parameterValue"`
+}
+
+func main() {
```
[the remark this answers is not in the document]
pmengelbert: Separate repo?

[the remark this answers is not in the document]
cpuguy83: Yes.

frontend/mariner2/handle_rpm.go
cpuguy83: We do need to sign SRPM's.
What I'm getting at with some of the suggestions is these should be passed as configuration (or by convention) to the signer.
What I mean by "by convention' here is the signer understands implicitly what its inputs are going to be and may not work with other inputs.

frontend/request.go
cpuguy83: Sorry, I think I flipped this in my head.
Looks like this is `opts["context"] = "input:<id>"`
However I think `id` should litterally be `context` by convention.

cmd/signer/Dockerfile
pmengelbert: Yeah, I meant "get the credentials from a keyvault to access the file from storage"

frontend/request.go
cpuguy83: Maybe it is not bad to have `dalec.target`, and the signer can abort if it doesn't know how to deal with it.

frontend/request.go
cpuguy83: If we pass `dalec.target`, then the signer should know how to deal with the input implicitly and we should not need to pass a pattern to it.

test/testenv/builld.go
pmengelbert: done

test/testenv/buildx.go
pmengelbert: done

cmd/signer/Dockerfile
pmengelbert: Irrelevant since this will be moved elsewhere.

cmd/signer/main.go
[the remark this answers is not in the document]
pmengelbert: done

frontend/mariner2/handle_rpm.go
pmengelbert: Now irrelevant for this PR since this code has been moved elsewhere.

test/fixtures/signer/main.go
```diff
@@ -0,0 +1,138 @@
...
+					OperationSetCode: signOp,
+					Parameters:       params,
+					ToolName:         "sign",
+					ToolVersion:      "1.0",
+				},
+			},
+			HashType: "sha256",
+		}
+
+		cc, ok := c.(frontend.CurrentFrontend)
```
DannyBrito, line 98: nit: a bit hard to see and differentiate between `c` and `cc`

docs/signing.md
DannyBrito, line 1: It seems liked we moved the docs to under `website/docs` maybe this file should be moved too?

frontend/mariner2/handle_rpm.go
cpuguy83: I don't think this is irrelevant.
This is the argument we are passing into every signer, but the signer should already be aware of what is signing based on the target key.

pmengelbert: Mistook this for another issue. Definitely relevant. It's been addressed on the frontend side but forgot to make the corresponding change here. Thanks.

test/fixtures/signer/main.go
pmengelbert: Done, thanks.

frontend/request.go
pmengelbert: done.

frontend/mariner2/handle_rpm.go
pmengelbert: Done.

docs/signing.md
DannyBrito: thanks! I think for adding on the website it needs an entry in here https://github.com/Azure/dalec/blob/main/website/sidebars.ts#L22. So, could you please add it?

pmengelbert: Done, thanks.

test/fixtures/signer/main.go
```diff
@@ -0,0 +1,138 @@
...
+			params = append(params, ParameterKV{
+				ParameterName:  "CoseFlags",
+				ParameterValue: "chainunprotected",
+			})
+
+		default:
+			signOp = linuxSignOp
+		}
+
+		config := Config{
```
cpuguy83, line 78: Can we simplify this down to something minimal for the test fixture?

test/signer_test.go
```diff
@@ -0,0 +1,196 @@
...
+// TestHandlerTargetForwarding tests that targets are forwarded to the correct frontend.
+// We do this by registering a phony frontend and then forwarding a target to it and checking the outputs.
+func TestSignerForwarding(t *testing.T) {
+	runTest := func(t *testing.T, f gwclient.BuildFunc) {
+		t.Helper()
+		ctx := startTestSpan(baseCtx, t)
+		testEnv.RunTest(ctx, t, f)
+	}
+
+	t.Run("test mariner2 signing", func(t *testing.T) {
```
cpuguy83, line 23: I think all this should move to the distro-level testing since its really making sure the distro implments signing correctly.

cpuguy83: Seems like we'd need to support `PackageConfig` at the root of the spec like we do other target fields.

cpuguy83 (changes requested):
This is looking really close!

test/fixtures/signer/main.go
pmengelbert: this is done.

test/signer_test.go
pmengelbert: > I think all this should move to the distro-level testing since its really making sure the distro implments signing correctly.

this part is done. Still need to support `PackageConfig` at spec root.

pmengelbert: Set up PackageConfig at the root level, so both parts of this are done now.

test/mariner2_test.go
```diff
@@ -304,6 +306,70 @@ echo "$BAR" > bar.txt
...
+					},
+				},
+				Artifacts: dalec.Artifacts{
+					Binaries: map[string]dalec.ArtifactConfig{
+						"foo": {},
+					},
+				},
+			}
+
+			sr := newSolveRequest(withSpec(ctx, t, &spec), withBuildTarget("mariner2/rpm"))
```
cpuguy83, line 353: This is specific to mariner, but we should treat this as a more generic test so that we can add other distros (azlinux3, jammy, etc) and get these tests for free.
Maybe we need to add an extra parameter to the test function to specify the signing target.

cpuguy83 (changes requested):

cpuguy83 (approved):

PR #215, opened 2024-04-19
mariner2: Use empty rootfs for default container

### LINKED ISSUE #51: Distroless build is a bit backwards
The way distroless containers are currently built is not ideal.
Right now (for mariner since that's all we have) we take a rootfs from either the default mariner ref (distroless base) or the ref set in the build spec.

The problem with this is distroless images don't have package db's (or any package manager).
This means when we install the new package it will overwrite whatever dependencies were already installed anyway.
The mariner ones do have manifest files that we could possibly re-hydrate a package db from... which is a possibility (maybe, I haven't looked into this).

This is a bit more problematic for debian based distros since the tooling isn't as forgiving as dnf for actually building up a rootfs (ordering of package install needs to be done manually)... but disregarding that, we're still going to replace existing packages already installed in the rootfs.

cpuguy83:
Before this we were using the distroless image's rootfs as the base for the output container.
With this change an empty rootfs is used unless the spec has an image specified.

Closes #51

### DIFF
--- a/frontend/mariner2/handle_container.go
+++ b/frontend/mariner2/handle_container.go
@@ -11,6 +11,7 @@ import (
 	"github.com/Azure/dalec/frontend"
 	"github.com/moby/buildkit/client/llb"
 	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+	"github.com/moby/buildkit/util/bklog"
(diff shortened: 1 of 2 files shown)


sozercan (approved):
LGTM

PR #240, opened 2024-05-09
Add ability to specify shared state and config directories

### LINKED ISSUE #239: [REQ] Support config and shared state directories

New feature

I would like to be able to specify config ( %{_sysconfdir}) and shared state (%{_sharedstatedir}) directories that need to be created when the RPM is installed. These should also show up in the files section of the RPM.

jeremyrickard:

This PR adds the ability to specify directories that should be created for %{_sharedstatedir} and  %{_sysconfdir}

This will ensure they are created during the install phase and that they show up in the files section of the resulting RPM spec.

In the spec, this would look like:

```
artifacts:
  createDirectories:
    config:
      kubernetes: 
        mode: 0755
      kubernetes/manifests: {}
    state:
      kubelet: {}
```

This would result in:

```
%install
mkdir -m 755 -p "%{buildroot}/%{_sysconfdir}/kubernetes"
mkdir -p "%{buildroot}/%{_sysconfdir}/kubernetes/manifests"
mkdir -p "%{buildroot}/%{_sharedstatedir}/kubelet"

%files
%dir %{_sysconfdir}/kubernetes
%dir %{_sysconfdir}/kubernetes/manifests
%dir %{_sharedstatedir}/kubelet
```

Fixes #239

### DIFF
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -57,6 +57,20 @@
 					},
 					"type": "object",
 					"description": "Manpages is the list of manpages to include in the package."
+				},
+				"configDirectories": {
+					"items": {
+						"type": "string"
+					},
+					"type": "array",
+					"description": "ConfigDirectories is a list of directories the RPM should under %{_sysconfigdir}\t."
+				},
+				"sharedStateDirectories": {
+					"items": {
+						"type": "string"
+					},
+					"type": "array",
+					"description": "SharedStateDirectories is a list of directories the RPM should under %{%_sharedstatedir}\t."
 				}
 			},
 			"additionalProperties": false,
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -63,14 +63,14 @@
 						"type": "string"
 					},
 					"type": "array",
-					"description": "ConfigDirectories is a list of directories the RPM should under %{_sysconfigdir}\t."
+					"description": "ConfigDirectories is a list of directories the RPM should place under %{_sysconfigdir}\t."
 				},
 				"sharedStateDirectories": {
 					"items": {
 						"type": "string"
 					},
 					"type": "array",
-					"description": "SharedStateDirectories is a list of directories the RPM should under %{%_sharedstatedir}\t."
+					"description": "SharedStateDirectories is a list of directories the RPM should place under %{%_sharedstatedir}\t."
 				}
 			},
 			"additionalProperties": false,
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -42,6 +42,17 @@
 			"type": "object",
 			"description": "ArtifactConfig is the configuration for a given artifact type."
 		},
+		"ArtifactDirConfig": {
+			"properties": {
+				"mode": {
+					"type": "integer",
+					"description": "Mode is used to set the file permission bits of the final created directory to the specified mode.\nMode is the octal permissions to set on the dir."
+				}
+			},
+			"additionalProperties": false,
+			"type": "object",
+			"description": "ArtifactDirConfig contains information about the directory to be created"
+		},
 		"Artifacts": {
 			"properties": {
 				"binaries": {
@@ -58,19 +69,9 @@
 					"type": "object",
 					"description": "Manpages is the list of manpages to include in the package."
 				},
-				"configDirectories": {
-					"items": {
-						"type": "string"
-					},
-					"type": "array",
-					"description": "ConfigDirectories is a list of directories the RPM should place under %{_sysconfigdir}\t."
-				},
-				"sharedStateDirectories": {
-					"items": {
-						"type": "string"
-					},
-					"type": "array",
-					"description": "SharedStateDirectories is a list of directories the RPM should place under %{%_sharedstatedir}\t."
+				"createDirectories": {
+					"$ref": "#/$defs/CreateArtifactDirectories",
+					"description": "Directories is a list of various directories that should be created by the RPM."
 				}
 			},
 			"additionalProperties": false,
@@ -237,6 +238,31 @@
 				"steps"
 			],
 			"description": "Command is used to execute a command to generate a source from a docker image."
+		},
+		"CreateArtifactDirectories": {
+			"properties": {
+				"Config": {
+					"additionalProperties": {
+						"$ref": "#/$defs/ArtifactDirConfig"
+					},
+					"type": "object",
+					"description": "Config is a list of directories the RPM should place under %{_sysconfigdir} (i.e. /etc)"
+				},
+				"State": {
+					"additionalProperties": {
+						"$ref": "#/$defs/ArtifactDirConfig"
+					},
+					"type": "object",
+					"description": "State is a list of directories the RPM should place under %{%_sharedstatedir} (i.e. /var/lib)."
+				}
+			},
+			"additionalProperties": false,
+			"type": "object",
+			"required": [
+				"Config",
+				"State"
+			],
+			"description": "CreateArtifactDirectories describes various directories that should be created on install."
 		},
 		"FileCheckOutput": {
 			"properties": {
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -246,14 +246,14 @@
 						"$ref": "#/$defs/ArtifactDirConfig"
 					},
 					"type": "object",
-					"description": "Config is a list of directories the RPM should place under %{_sysconfigdir} (i.e. /etc)"
+					"description": "Config is a list of directories the RPM should place under the system config directory (i.e. /etc)"
 				},
 				"State": {
 					"additionalProperties": {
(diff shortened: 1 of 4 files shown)


docs/spec.schema.json
```diff
@@ -57,6 +57,20 @@
 					},
 					"type": "object",
 					"description": "Manpages is the list of manpages to include in the package."
+				},
+				"configDirectories": {
+					"items": {
+						"type": "string"
+					},
+					"type": "array",
+					"description": "ConfigDirectories is a list of directories the RPM should under %{_sysconfigdir}\t."
```
adamperlin, line 66: nit: "the RPM should place under"

docs/spec.schema.json
```diff
@@ -57,6 +57,20 @@
...
+				},
+				"sharedStateDirectories": {
+					"items": {
+						"type": "string"
+					},
+					"type": "array",
+					"description": "SharedStateDirectories is a list of directories the RPM should under %{%_sharedstatedir}\t."
```
adamperlin, line 73: nit: "the RPM should place under"

adamperlin:
My only concern really is whether or not it's confusing to have `configDirectories` and `sharedStateDirectories` as artifacts, when they're things that are being created and not just extracted from the build. What do we think about this? Having the paths as lists instead of map keys does differentiate these sections of the `artifacts` from, say, the map keys.

docs/spec.schema.json
jeremyrickard: thank you!

docs/spec.schema.json
jeremyrickard: thank you!

jeremyrickard:
> My only concern really is whether or not it's confusing to have `configDirectories` and `sharedStateDirectories` as artifacts, when they're things that are being created and not just extracted from the build. What do we think about this? Having the paths as lists instead of map keys does differentiate these new sections of the `artifacts` from, say, the `binaries` section.

I think the `artifacts` section is really where we intend to put all the things that land in the artifact (i.e. the RPM) though, would it be more confusing to have extra pieces in different part of the spec file configuring / modifying that?

spec.go
```diff
@@ -127,7 +127,12 @@ type Artifacts struct {
 	Binaries map[string]ArtifactConfig `yaml:"binaries,omitempty" json:"binaries,omitempty"`
 	// Manpages is the list of manpages to include in the package.
 	Manpages map[string]ArtifactConfig `yaml:"manpages,omitempty" json:"manpages,omitempty"`
+	// ConfigDirectories is a list of directories the RPM should place under %{_sysconfigdir}	.
+	ConfigDirectories []string `yaml:"configDirectories,omitempty" json:"configDirectories,omitempty"`
```
cpuguy83, line 131: Maybe we can do something like `createDirectories`

which is a type like so:

```go
type CreateArtifactDirectories struct {
    Config: map[string]ArtifactDirConfig // These go in /etc
    State: map[string]ArtifactDirConfig // These go in /var/lib
}
```

frontend/rpm/template.go
```diff
@@ -326,6 +326,17 @@ func (w *specWrapper) Install() fmt.Stringer {
 		copyArtifact(`%{buildroot}/%{_mandir}`, p, cfg)
 	}
 
+	sort.Strings(w.Spec.Artifacts.ConfigDirectories)
+	for _, p := range w.Spec.Artifacts.ConfigDirectories {
+		dir := filepath.Join(`%{buildroot}/%{_sysconfdir}`, p)
+		fmt.Fprintln(b, "mkdir -p", dir)
```
cpuguy83, line 332: Need to make sure the dirs are quoted for the shell.

```suggestion
		fmt.Fprintf(b, "mkdir -p %q\n", dir)
```

frontend/rpm/template.go
```diff
@@ -326,6 +326,17 @@ func (w *specWrapper) Install() fmt.Stringer {
...
+	}
+
+	sort.Strings(w.Spec.Artifacts.SharedStateDirectories)
+	for _, p := range w.Spec.Artifacts.SharedStateDirectories {
+		dir := filepath.Join(`%{buildroot}/%{_sharedstatedir}`, p)
+		fmt.Fprintln(b, "mkdir -p", dir)
```
cpuguy83, line 338: Same, dir needs quoting.

cpuguy83 (changes requested):

spec.go
jeremyrickard: @cpuguy83  what would you see modeled in map form?

```
artifacts:
     createDirectories: 
          config: 
              <what's the map look like here, how do we want to use subpath and name?>
         state:
               <what's the map look like here, how do we want to use subpath and name?>
    ```

Comment (cpuguy83) [reply spec.go:131]: Map key would be the dir name to create.

File: frontend/rpm/template.go (hunk 4 above)
Comment (jeremyrickard) [reply frontend/rpm/template.go:332]: good catch thanks!

File: spec.go (hunk 3 above)
Comment (jeremyrickard) [reply spec.go:131]: What would the other ArtifactConfig things be used for? or would we just pretty much never include them?  

```
createDirectories: 
          config: 
               kubernets: {}
               kubernetes/manifests: {}
```

or 

```
createDirectories: 
          config: 
               kubernets: 
                   subpath: manifests
```

Comment (jeremyrickard) [reply spec.go:131]: oh sorry, you said `ArtifactDirConfig` which could be other things my bad, sorry!

Comment (jeremyrickard) [reply spec.go:131]: maybe for artifact dir config we want to optionally specify mode (i.e. mkdir -m <specified>) ?

Comment (cpuguy83) [reply spec.go:131]: Yep.

**Comment by adamperlin** [comment]
> > My only concern really is whether or not it's confusing to have `configDirectories` and `sharedStateDirectories` as artifacts, when they're things that are being created and not just extracted from the build. What do we think about this? Having the paths as lists instead of map keys does differentiate these new sections of the `artifacts` from, say, the `binaries` section.
> 
> I think the `artifacts` section is really where we intend to put all the things that land in the artifact (i.e. the RPM) though, would it be more confusing to have extra pieces in different part of the spec file configuring / modifying that?

Yeah, that makes sense. It would likely be much more confusing to have extra pieces in other parts of the spec. The `createDirectories` idea seems good since it makes the creation piece explicit

**Comment by jeremyrickard** [comment]
OK, I have refactored this to use the suggestion above to have a `createDirectories` block in the YAML and corresponding struct in the spec.go. 

Now the user's YAML spec  file would look like:

```
artifacts:
  binaries:
    kubernetes/_output/bin/kubelet: {}
  createDirectories:
    config:
      kubernetes: 
        mode: 0755
      kubernetes/manifests: {}
    state:
      kubelet: {}
```

if `mode` isn't provided, the `-m` arg will be committed from the mkdir. Path is now also quoted.

File: spec.go [hunk 6]
```diff
@@ -127,9 +127,27 @@ type Artifacts struct {
...
 	Manpages map[string]ArtifactConfig `yaml:"manpages,omitempty" json:"manpages,omitempty"`
+	// Directories is a list of various directories that should be created by the RPM.
+	Directories *CreateArtifactDirectories `yaml:"createDirectories,omitempty" json:"createDirectories,omitempty"`
 	// TODO: other types of artifacts (systtemd units, libexec, configs, etc)
 }
 
+// CreateArtifactDirectories describes various directories that should be created on install.
+// CreateArtifactDirectories represents different directory paths that are common to RPM systems.
+type CreateArtifactDirectories struct {
+	// Config is a list of directories the RPM should place under %{_sysconfigdir} (i.e. /etc)
```
cpuguy83, line 138:
```suggestion
	// Config is a list of directories the RPM should place under the system config directory (e.g. `/etc`)
```

spec.go
```diff
@@ -127,9 +127,27 @@ type Artifacts struct {
...
+	Config map[string]ArtifactDirConfig
+	// State is a list of directories the RPM should place under %{%_sharedstatedir} (i.e. /var/lib).
```
cpuguy83, line 140:
```suggestion
	// State is a list of directories the RPM should place under the common directory for shared (e.g. `/var/lib`).
```

cpuguy83 (changes requested):
Can you add a test case in https://github.com/Azure/dalec/blob/565c033351b7d5d104436fdafe56e887ecc00ae8/test/mariner2_test.go#L21?

jeremyrickard:
@cpuguy83  I added a test. I noticed that we removed the previous tests that checked permissions, in this case I think we should probably check them, although I am not sure if this is portable (paths too maybe?).

I did a first version of this that just ensured it ran successfully and that was fine but I thought maybe we would want to test the permissions because it's an option that can be specified in the config for the directories. LMK what you think.

cpuguy83 (approved):

cpuguy83:
>  in this case I think we should probably check them, although I am not sure if this is portable (paths too maybe?)

We'll see when I implement this in jammy 😅

PR #247, opened 2024-05-13
Add support to spec for config files from sources

### LINKED ISSUE #246: [REQ] Provide ability to define config files from sources

New feature

I would like to be able to use a source to populate a config file into the artifact generated by DALEC. I would also like to leverage the %config(noreplace) macro in the resulting RPM 

jeremyrickard:

This PR adds the ability to define config files for the resulting artifact and the relevant %config entries in the %files section, It also supports the optional inclusion of the %config(noreplace) with a boolean in the config.

This also updates the spec to add this (and directories) to the IsEmpty method for the Artifact.

Fixes #246 

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -127,10 +127,10 @@ type Artifacts struct {
 	Binaries map[string]ArtifactConfig `yaml:"binaries,omitempty" json:"binaries,omitempty"`
 	// Manpages is the list of manpages to include in the package.
 	Manpages map[string]ArtifactConfig `yaml:"manpages,omitempty" json:"manpages,omitempty"`
-	// Directories is a list of various directories that should be created by the RPM.
+	// Directories is a list of various directories that should be created by the package.
 	Directories *CreateArtifactDirectories `yaml:"createDirectories,omitempty" json:"createDirectories,omitempty"`
-	// ConfigFiles is a list of files that should be marked as config files in the RPM.
-	ConfigFiles map[string]ArtifactConfigFileConfig `yaml:"configFiles,omitempty" json:"configFiles,omitempty"`
+	// ConfigFiles is a list of files that should be marked as config files in the package.
+	ConfigFiles map[string]ArtifactConfig `yaml:"configFiles,omitempty" json:"configFiles,omitempty"`
 	// TODO: other types of artifacts (systtemd units, libexec, etc)
 }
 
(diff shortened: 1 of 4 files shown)


spec.go
```diff
@@ -129,7 +129,9 @@ type Artifacts struct {
 	Manpages map[string]ArtifactConfig `yaml:"manpages,omitempty" json:"manpages,omitempty"`
 	// Directories is a list of various directories that should be created by the RPM.
 	Directories *CreateArtifactDirectories `yaml:"createDirectories,omitempty" json:"createDirectories,omitempty"`
-	// TODO: other types of artifacts (systtemd units, libexec, configs, etc)
+	// ConfigFiles is a list of files that should be marked as config files in the RPM.
```
cpuguy83, line 132: Missed the `in the RPM` part in the last PR.
Should be `in the package`

spec.go
```diff
@@ -161,6 +163,13 @@ type ArtifactConfig struct {
 	Name string `yaml:"name,omitempty" json:"name,omitempty"`
 }
 
+// ArtifactConfigFileConfig is the configuration give for Artifact Config Files.
+// This is used to customize the settings for a config entry in the resulting artifact.
+type ArtifactConfigFileConfig struct {
+	ArtifactConfig
+	ReplaceOnUpdate bool `yaml:"replace,omitempty" json:"replace,omitempty"`
```
cpuguy83, line 170: Can you add a doc string for this?

spec.go
```diff
@@ -161,6 +163,13 @@ type ArtifactConfig struct {
 	Name string `yaml:"name,omitempty" json:"name,omitempty"`
 }
 
+// ArtifactConfigFileConfig is the configuration give for Artifact Config Files.
+// This is used to customize the settings for a config entry in the resulting artifact.
+type ArtifactConfigFileConfig struct {
+	ArtifactConfig
```
cpuguy83, line 169:
```suggestion
	ArtifactConfig `yaml:",inline"`
```

cpuguy83 (changes requested):

spec.go
cpuguy83: Actually reading up on `%config(noreplace)`, I think we should just assume this and use a normal `ArtifactConfig`.
If we need to tweak in the future we can add an option but I'm thinking `%config(noreplace)` is the normal correct behavior.

spec.go
jeremyrickard: I've removed this entire struct now in favor of just Artifactconfig.

spec.go
jeremyrickard: sounds good to me, made this change.

spec.go
jeremyrickard: *doh* updated this and the directories string.

cpuguy83 (approved):

PR #248, opened 2024-05-13
tests: Allow test suite to run on mac

cpuguy83:
1. Removes custom buildx config inpsection for new `buildx dial-stdio` subcommand for creating a buildkit client. This not only allows *any* buildx instance to be used, in particular it unblocks docker desktop.
2. Sets test fixtures to explicitly set the platform to the worker's platform. This is needed because the LLB is generated on the client and was assuming the client's platform (e.g. darwin/arm64 on mac).
3. Marks windows tests to use explictlty set amd64 as the target platform since windows/arm64 is not supported. This allows the test suite to run against an arm64 worker (e.g. docker for mac).

### DIFF
--- a/frontend/windows/handle_container.go
+++ b/frontend/windows/handle_container.go
@@ -20,7 +20,11 @@ const (
 
 var (
 	defaultPlatform = ocispecs.Platform{
-		OS:           outputKey,
+		OS: outputKey,
+		// NOTE: Windows is (currently) only supported on amd64.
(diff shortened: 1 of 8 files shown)


cpuguy83:
@jeremyrickard LGTY?

jeremyrickard:
@cpuguy83  I was able to run the tests locally on my Mac with your branch, so I think this is good to go.

jeremyrickard (approved):

PR #250, opened 2024-05-14
docs: update signing details

DannyBrito:
Updating docs in reference to singing,

This is a bit contradicting to Original PR in [dalec-build-defs](https://github.com/Azure/dalec-build-defs/pull/66#issue-2275877981), were shown example uses `mariner2/container` target and leverages signing!

Not sure what is the correct, if the example is wrong or we are missing the the signing frontend call during `mariner2/container` call.

### DIFF
--- a/website/docs/signing.md
+++ b/website/docs/signing.md
@@ -1,5 +1,7 @@
 # Signing Packages
 
+***Note:*** Avalible on dalec release `v0.3.0+`
+
 Packages can be automatically signed using Dalec. To do this, you will
 need to provide a signing frontend. There is an example in the test
 code `test/signer/main.go`. Once that signing image has been built and
@@ -14,6 +16,9 @@ targets: # Distro specific build requirements
         image: "ref/to/signing:image"
         cmdline: "/signer"
 ```
+
+These are the targets that can leverage signing:
+`windowscross/container`, `windowscross/zip`, `mariner2/rpm`.
 
 This will send the artifacts (`.rpm`, `.deb`, or `.exe`) to the
 signing frontend as the build context.
--- a/website/docs/signing.md
+++ b/website/docs/signing.md
@@ -1,6 +1,8 @@
 # Signing Packages
 
-***Note:*** Avalible on dalec release `v0.3.0+`
+:::note
+Available with Dalec release `v0.3.0` and later.
+:::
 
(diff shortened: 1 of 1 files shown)


website/docs/signing.md
```diff
@@ -1,5 +1,7 @@
 # Signing Packages
 
+***Note:*** Avalible on dalec release `v0.3.0+`
```
sozercan, line 3:
```suggestion
***Note:*** Available with Dalec release `v0.3.0+`
```

sozercan: I'll open an issue for supporting versioned docs so we don't have to do this in the future

website/docs/signing.md
```diff
@@ -15,6 +17,9 @@ targets: # Distro specific build requirements
         cmdline: "/signer"
 ```
 
+These are the targets that can leverage signing:
+`windowscross/container`, `windowscross/zip`, `mariner2/rpm`.
```
sozercan, line 21: can we say something like "at this time"?

re: windowscross/container, do we need to clarify that this is not container signing but artifact signing inside the container?

jeremyrickard:
> Package signing should work on the mariner2/container target. I'll double check this later today, but that is the intent.

This would sign the packages inside but not the container itself, since that is a different thing correct?

cpuguy83:
Correct. This is all packaging signing.

cpuguy83:
Ok yeah this is a bug that mariner2/container does not use the signed version.
I'll post a patch.

cpuguy83:
https://github.com/Azure/dalec/pull/252 fixes this

sozercan (approved):
LGTM

website/docs/signing.md
```diff
@@ -1,5 +1,7 @@
 # Signing Packages
 
+***Note:*** Available with Dalec release `v0.3.0` and later.
```
sozercan, line 3:
```suggestion
:::note
Available with Dalec release `v0.3.0` and later.
:::
```

sozercan: This is a docusaurus specific syntax: https://docusaurus.io/docs/markdown-features/admonitions

DannyBrito: Oh, cool!

PR #252, opened 2024-05-14
fix: Ensure container targets use signed artifact

cpuguy83:
I don't have a good test case for this right now.
I'll open an issue because ideally we can ensure that this doesn't regress in the future.


DannyBrito (approved):
lgtm

PR #254, opened 2024-05-15
WIP Systemd Unit Artifacts

### LINKED ISSUE #225: [REQ] Support creating systemd units

New feature

I would like to be able to specify configuration in the DALEC spec to create systemd units for packages. 

adamperlin:
Adds support for specifying systemd units as artifacts in dalec. This is accomplished by adding a new services subsection to `artifacts` in a dalec spec that looks like the following:

```yaml
 systemdUnits:
    src/simple.service:

```
Additional options are supported to configure the unit
```yaml
src/simple.service:
     # alias name service should be installed under
     name: simple@.service
     # should service be enabled on install
     enable: true
```

Fixes #225


frontend/rpm/template.go
```diff
@@ -105,6 +108,11 @@ func (w *specWrapper) Requires() fmt.Stringer {
 		writeDep(b, "BuildRequires", name, constraints)
 	}
 
+	if len(w.Artifacts.Services) > 0 {
+		// We take advantage of the systemd-rpm-macros package to simplify the service file installation
+		writeDep(b, "BuildRequires", "systemd-rpm-macros", nil)
```
cpuguy83, line 113: This can't really be set here since it is distro specific (and probably not right to have in the spec itself).
I think this could be added to the base builder image directly.

test/fixtures/env-multiple-commands.yml
```diff
@@ -11,6 +11,7 @@ license: Apache 2.0
 
 targets: # Distro specific build requirements
   mariner2:
+    base: mcr.microsoft.com/cbl-mariner/base/core:2.0
```
cpuguy83, line 14: Why?

spec.go
```diff
@@ -132,6 +132,7 @@ type Artifacts struct {
 	// ConfigFiles is a list of files that should be marked as config files in the package.
 	ConfigFiles map[string]ArtifactConfig `yaml:"configFiles,omitempty" json:"configFiles,omitempty"`
 	// TODO: other types of artifacts (systtemd units, libexec, etc)
+	Services map[string]ServiceConfig `yaml:"services,omitempty" json:"services,omitempty"`
```
cpuguy83, line 135: How about we call this `systemdUnit` so its explicit as to what it is?
It also means we can add other types of units other than just services.

test/fixtures/systemd-unit.yml
```diff
@@ -0,0 +1,56 @@
...
+version: 0.0.1
+revision: 1
+vendor: Microsoft
+license: Apache 2.0
+packager: Microsoft <support@microsoft.com>
+
+targets:
+  mariner2:
+    image:
+      base: mcr.microsoft.com/cbl-mariner/base/core:2.0
```
cpuguy83, line 15: How come this needs an explicit base image?

frontend/rpm/template.go
```diff
@@ -355,6 +411,27 @@ func (w *specWrapper) Install() fmt.Stringer {
 		cfg := w.Spec.Artifacts.ConfigFiles[c]
 		copyArtifact(`%{buildroot}/%{_sysconfdir}`, c, cfg)
 	}
+	serviceKeys := dalec.SortMapKeys(w.Spec.Artifacts.Services)
+	for _, p := range serviceKeys {
+		cfg := w.Spec.Artifacts.Services[p]
+		// must include '.service' suffix in name
```
cpuguy83, line 417: We should support more than `.service` files

cpuguy83 (changes requested):

frontend/rpm/template.go
adamperlin: As in other file formats?

test/fixtures/systemd-unit.yml
adamperlin: It actually does not

test/fixtures/env-multiple-commands.yml
adamperlin: Fixed.

frontend/rpm/template.go
cpuguy83: There's other types of systemd units, e.g. `<foo>.mount` or `<foo>.socket`

adamperlin: Ok. Do we want to add declarative options for these, to be extra explicit
```yaml
systemdUnits:
    mounts:
    sockets:
         src/foo.socket:
    services:
        src/foo.service:
    ...
```
We will need to handle sockets differently. I.e in the above example we would want to enable `foo.socket`, but not `foo.service` (assuming foo.service is supposed to be started once there are connections to foo.socket).  One complicating thing is that `.socket` units can override what service they start using the `Service=` option, which would make it harder for us to figure out what to do, unless the user tells us explicitly what should be enabled and disabled.

adamperlin: If we're going to add sockets, perhaps we should flip the default to disabled and make the user write `enable: true` to be absolutely sure that we don't accidentally enable a service that is associated with, for example, a socket

jeremyrickard: For our k8s builds, we need services. IF we need to handle things differently, would it make sense to handle the service case first and then worry about other units?

cpuguy83: I don't think we need to have special cases for different types of units.
They are all just files that get copied into some specific location.
As for different defaults and things I don't think its needed, each unit its its own thing.
A socket does reference a service, but its not on us to make sure the unit file is correct.
We are just starting the unit or not.

adamperlin: Ok. I think the best path forward then is to just make users put `enable: true` as a subproperty of the unit config, otherwise it will default to disabled

cpuguy83: SGTM

frontend/rpm/template.go
adamperlin: This should be fixed now.

frontend/rpm/template.go
adamperlin: I've now implemented this.

test/fixtures/service/socket/main.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/python3
```
adamperlin, line 1: This file will be installed as a binary in one of the test systemd spec containers, but the server it starts won't actually be checked. I'm not sure we can really do a full e2e test in our current setup since systemd must already be running in a container for the server to be initialized

spec.go
```diff
@@ -131,7 +132,35 @@ type Artifacts struct {
...
+	// Nested paths are not supported. It is the user's responsibility
+	// to name the service with the appropriate extension, i.e. .service, .timer, etc.
+	Name string `yaml:"name" json:"name"`
+
+	// Enable is used to enable the systemd unit on install
+	// This determines what will be written to a systemd preset file
+	Enable bool `yaml:"enable" json:"enabled"`
+
+	// Should service be reloaded instead of restarted? (Mutually exclusive with `Restart`)
+	Reload bool
```
jeremyrickard, line 153: Since these are mutually exclusive, should we consider having an enum type thing for this?   Mode w/ some const values that make sense and a default to whatever? Don't know if that would be easier for end users than two mutually exclusive booleans?

cpuguy83: I'd be in favor of removing these options for now until after we get jammy in.

adamperlin: Switched to an enum

cpuguy83 (changes requested):

spec.go
```diff
@@ -135,9 +136,38 @@ type Artifacts struct {
...
+
+type SystemdUnitConfig struct {
+	// Name is the name systemd unit should be copied under.
+	// Nested paths are not supported. It is the user's responsibility
+	// to name the service with the appropriate extension, i.e. .service, .timer, etc.
+	Name string `yaml:"name" json:"name"`
+
+	// Enable is used to enable the systemd unit on install
+	// This determines what will be written to a systemd preset file
+	Enable bool `yaml:"enable" json:"enabled"`
```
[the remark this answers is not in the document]
adamperlin: Oh good catch

spec.go
```diff
@@ -135,9 +136,38 @@ type Artifacts struct {
...
+
+	// On upgrade, should the service be restarted, reloaded, or neither
+	UpgradeRefreshPolicy string `yaml:"upgradeRefreshPolicy,omitempty" json:"systemdRefreshPolicy,omitempty" jsonschema:"enum=none,enum=reload,enum=restart"`
```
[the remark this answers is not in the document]
adamperlin: No I don't think it's strictly needed, especially since I'm not even sure we have use cases where we're upgrading packages we're producing yet right?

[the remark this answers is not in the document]
adamperlin: I'll remove it for now

test/fixtures/systemd-socket-unit.yml
```diff
@@ -0,0 +1,43 @@
+# syntax=local/dalec/frontend
```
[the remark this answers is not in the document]
adamperlin: I've deleted the systemd-* test fixtures for now. In the future I think it would be nice to have things like that as integration tests, which I think would require us to actually start containers built by dalec and exec commands in them for testing. How do you feel about that?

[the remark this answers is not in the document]
cpuguy83: I agree it would be good to have some tests around this, I'm not sure what that looks like just yet.

There is some support for manually managing the lifecycle of a container with buildkit, but its fairly minimal functionality and, depending on what we end up needing, could require a lot of extra stuff to make it do what we need, including managing our own buildkit instance so we can have the right security entitlements for running systemd.

go.mod
```diff
@@ -19,12 +19,14 @@ require (
 	github.com/opencontainers/image-spec v1.1.0
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.9.3
+	github.com/stretchr/testify v1.8.4
```
cpuguy83, line 22: Can you run `go mod tidy` to clean this up?

test/fixtures/systemd-socket-unit.yml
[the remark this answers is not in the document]
cpuguy83: Looks like the files are still here as well.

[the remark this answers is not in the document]
adamperlin: Whoops forgot to delete the actual fixtures

test/azlinux_test.go
```diff
@@ -340,6 +340,191 @@ echo "$BAR" > bar.txt
...
+			Description: "Test systemd unit",
+			Website:     "https://www.github.com/Azure/dalec",
+			Version:     "0.0.1",
+			Revision:    "1",
+			Vendor:      "Microsoft",
+			License:     "Apache 2.0",
+			Packager:    "Microsoft <support@microsoft.com>",
+			Dependencies: &dalec.PackageDependencies{
+				Build: map[string][]string{
+					"msft-golang": {},
```
cpuguy83, line 357: Doesn't seem like we need go here.

test/azlinux_test.go
```diff
@@ -340,6 +340,191 @@ echo "$BAR" > bar.txt
...
```
cpuguy83, line 453: Doesn't seem like we need go here.

spec.go
```diff
@@ -135,9 +136,35 @@ type Artifacts struct {
 	Docs map[string]ArtifactConfig `yaml:"docs,omitempty" json:"docs,omitempty"`
 	// Licenses is a list of doc files included in the package
 	Licenses map[string]ArtifactConfig `yaml:"licenses,omitempty" json:"licenses,omitempty"`
+	// SystemdUnits is a list of systemd units to include in the package.
+	SystemdUnits map[string]SystemdUnitConfig `yaml:"systemdUnits,omitempty" json:"systemdUnits,omitempty"`
 	// TODO: other types of artifacts (systtemd units, libexec, etc)
 }
 
+const (
+	SystemdUnitUpgradePolicyNone    = "none"
```
cpuguy83, line 145: We can take these policies out

cpuguy83 (changes requested):

cpuguy83 (approved):

PR #255, opened 2024-05-17
Add method to easily spin up doc server

cpuguy83:
This uses dagger to build the website and serve on a local port. Example usage:

```console
$ go -C ./website run .
```

This will, by default, make the server available on port 3000.

I've set this up as a separate module so that dagger dependencies do not interfere with the main module dependencies.


DannyBrito:
could, we also add this as maybe reference page in the docs-website on how to run it :)

sozercan:
wow this is fancy

+1 to @DannyBrito's comment about adding to https://github.com/Azure/dalec/blob/main/website/README.md

sozercan (approved):
LGTM. Thanks!

cpuguy83:
Added the readme, also made some minor changes to make the startup a lot cleaner and better use of caching for node modules.

PR #256, opened 2024-05-17
Add handling for docs files and license files

### LINKED ISSUE #241: [REQ] Support providing doc and license files for the resulting RPM spec

New feature

I would like to include one or more license and doc files in my resulting package, and have them reflected in the files directive. 

jeremyrickard:

This PR adds the ability to add files to the package and have them represented by the `%doc` directive or the %license directive within the %files list.

There isn't a corresponding install entry, because RPM handles this already:

> When the package is installed, RPM creates a directory in the documentation directory named the same as the package and copies the file there

The same happens with license files. 

I thought that a slice of strings made more sense than using ArtifactConfig here, since we don't really need to provide any other real options. It seems like you can specify a full path with the %doc directive (%license as well), but I think we can go with this for now? It feels cleaner/more organized. 

fixes #241 

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -132,9 +132,9 @@ type Artifacts struct {
 	// ConfigFiles is a list of files that should be marked as config files in the package.
 	ConfigFiles map[string]ArtifactConfig `yaml:"configFiles,omitempty" json:"configFiles,omitempty"`
 	// DocFiles is a list of doc files included in the package
-	DocFiles []string `yaml:"docFiles,omitempty" json:"docFiles,omitempty"`
+	Docs map[string]ArtifactConfig `yaml:"docs,omitempty" json:"docs,omitempty"`
 	// LicenseFiles is a list of doc files included in the package
-	LicenseFiles []string `yaml:"licenseFiles,omitempty" json:"licenseFiles,omitempty"`
+	Licenses map[string]ArtifactConfig `yaml:"licenses,omitempty" json:"licenses,omitempty"`
 	// TODO: other types of artifacts (systtemd units, libexec, etc)
 }
 
@@ -181,10 +181,10 @@ func (a *Artifacts) IsEmpty() bool {
 	if len(a.ConfigFiles) > 0 {
 		return false
 	}
-	if len(a.DocFiles) > 0 {
+	if len(a.Docs) > 0 {
 		return false
 	}
-	if len(a.LicenseFiles) > 0 {
+	if len(a.Licenses) > 0 {
 		return false
 	}
 	return true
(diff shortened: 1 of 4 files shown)


spec.go
```diff
@@ -177,6 +181,9 @@ func (a *Artifacts) IsEmpty() bool {
 	if len(a.ConfigFiles) > 0 {
 		return false
 	}
+	if len(a.DocFiles) > 0 {
+		return false
+	}
```
DannyBrito, line 186: Does the `LicenseFiles` Field need to be check too if empty?
```suggestion
	}
	if len(a.LicenseFiles) > 0 {
		return false
	}
```

jeremyrickard: oh yes, thanks I forgot! let me update.

spec.go
```diff
@@ -131,6 +131,10 @@ type Artifacts struct {
 	Directories *CreateArtifactDirectories `yaml:"createDirectories,omitempty" json:"createDirectories,omitempty"`
 	// ConfigFiles is a list of files that should be marked as config files in the package.
 	ConfigFiles map[string]ArtifactConfig `yaml:"configFiles,omitempty" json:"configFiles,omitempty"`
+	// DocFiles is a list of doc files included in the package
+	DocFiles []string `yaml:"docFiles,omitempty" json:"docFiles,omitempty"`
```
cpuguy83, line 135: The reason I like to make these structs is so that it is not a breaking change to add additional options.
I also think `subdir` is a thing relavent for these files (and licenses).

jeremyrickard: Got it to work. I needed to handle the copy myself for the subpaths and emit the full path inside the %files section.

spec.go
```diff
@@ -131,6 +131,10 @@ type Artifacts struct {
 	Directories *CreateArtifactDirectories `yaml:"createDirectories,omitempty" json:"createDirectories,omitempty"`
 	// ConfigFiles is a list of files that should be marked as config files in the package.
 	ConfigFiles map[string]ArtifactConfig `yaml:"configFiles,omitempty" json:"configFiles,omitempty"`
+	// DocFiles is a list of doc files included in the package
```
cpuguy83, line 134: These doc strings are not correct anymore 🙈

cpuguy83:
```suggestion
	// Docs is a list of doc files included in the package
```

spec.go
```diff
@@ -131,6 +131,10 @@ type Artifacts struct {
...
+	Docs map[string]ArtifactConfig `yaml:"docs,omitempty" json:"docs,omitempty"`
+	// LicenseFiles is a list of doc files included in the package
```
cpuguy83, line 136:
```suggestion
	// Licenses is a list of doc files included in the package
```

spec.go
jeremyrickard: LOL sorry

cpuguy83 (approved):

PR #258, opened 2024-05-17
Splitting Implement FS.ReadDirFs

DannyBrito:
Splitting Implement FS.ReadDirFs away from https://github.com/Azure/dalec/pull/191

Fixes #

### DIFF
--- a/frontend/fs.go
+++ b/frontend/fs.go
@@ -1,11 +1,10 @@
-package dalec
+package frontend
 
 import (
 	"context"
 	"io"
 	"io/fs"
 	"path"
-	"strings"
 	"sync"
 
 	gwclient "github.com/moby/buildkit/frontend/gateway/client"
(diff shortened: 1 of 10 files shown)


frontend/fs.go
```diff
@@ -0,0 +1,211 @@
+package frontend
```
cpuguy83, line 1: Let's put this in a separate package.
`pkg/bkfs` or `frontend/pkg/bkfs`?

frontend/fs.go
```diff
@@ -0,0 +1,211 @@
...
+	s         llb.State
+	ctx       context.Context
+	opts      []llb.ConstraintsOpt
+	client    gwclient.Client
+	clientRes *gwclient.Result
+	o         sync.Once
+	initErr   error
+}
+
+func NewStateRefFS(s llb.State, ctx context.Context, client gwclient.Client) *StateRefFS {
```
cpuguy83, line 32: Maybe `FromRef(context.Context, gwclient.Reference)`

With that something like`FromState(ctx llb.State, client gwclient.Client)` can be a convenience wrapper/

With `FromState` I think we should do the solve up front rather than lazily.

frontend/gateway.go
```diff
@@ -119,6 +120,9 @@ func SourceOptFromClient(ctx context.Context, c gwclient.Client) (dalec.SourceOp
 			}
 			return st, nil
 		},
+		GetFS: func(st llb.State) fs.ReadDirFS {
```
cpuguy83, line 123: Can we leave these changes in the other PR?

test/fs_test.go
```diff
@@ -0,0 +1,363 @@
...
+			assert.Equal(t, want.perms.String(), info.Mode().String())
+			assert.Equal(t, want.isDir, info.IsDir())
+		}
+
+		return rfs.Res()
+	})
+
+}
+
+func TestStateWrapper_ReadDir_Context(t *testing.T) {
```
cpuguy83, line 152: This seems like it belongs in the other PR.

source_build_test.go
```diff
@@ -22,7 +22,7 @@ func TestSourceBuild(t *testing.T) {
 	}
 
 	ctx := context.Background()
-	ops := getSourceOp(ctx, t, src)
+	ops := getSourceOp(ctx, t, src, []MockDirEntry{})
```
cpuguy83, line 25: Same comment re: keeping this in the other PR

helpers.go
```diff
@@ -183,7 +183,7 @@ func (f localOptionFunc) SetLocalOption(li *llb.LocalInfo) {
 	f(li)
 }
 
-func localIncludeExcludeMerge(includes []string, excludes []string) localOptionFunc {
+func LocalIncludeExcludeMerge(includes []string, excludes []string) localOptionFunc {
```
cpuguy83, line 186: Same comment re: keeping this in the other PR

Also, let's look at if this needs to be exported.

cpuguy83 (changes requested):

frontend/fs.go
adamperlin: This is done, let us know if this is what you were thinking.

cpuguy83 (approved):

PR #266, opened 2024-05-24
Fix filepath handling for a few types

### LINKED ISSUE #265: [BUG] Incorrect paths in %files section for a few types
### Expected Behavior

I noticed a few unexpected results in the spec file and rpm after using the latest 0.4.0 tag.

1. If I don't specify a `name` for the artifact config for a binary, I just get the directory in the %files section.
2. The `name` property is ignored for config files and a few other newer types.

EXPECTED:

```
binaries:
    kubernetes/_output/bin/kubelet: {}
```

yields:

```
%files
%{_bindir}/kubelet
```

and:

```
configFiles:
    kubernetes-release/cmd/krel/templates/latest/kubelet/kubelet.env:
      name: kubelet
      subpath: sysconfig
```

yields:

```
%files
%config(noreplace) %{_sysconfdir}/sysconfig/kubelet
```

### Actual Behavior

```
binaries:
    kubernetes/_output/bin/kubelet: {}
```

yields:

```
%files
%{_bindir}
```

While:

```
configFiles:
    kubernetes-release/cmd/krel/templates/latest/kubelet/kubelet.env:
      name: kubelet
      subpath: sysconfig
```

yields:

```
%files
%config(noreplace) %{_sysconfdir}/sysconfig/kubelet.env
```

### Steps To Reproduce

_No response_

jeremyrickard:

This PR corrects a few unexpected results when generating a package 

Fixes #265 

55ff43e7480aadaafb09a3da77e49c47a97a2fae refactors a bit to create a new function that will use filepath.Base as default and replace it with cfg.Name if provided

507e531af5ff3c547b6c88e7643111cbbb81e634 changes how license and doc files are handled, because currently they need to be at the root of the buildroot unless a subpath was specified

### DIFF
--- a/frontend/rpm/template.go
+++ b/frontend/rpm/template.go
@@ -345,15 +345,11 @@ func (w *specWrapper) Install() fmt.Stringer {
 		fmt.Fprintln(b, "mkdir -p", targetDir)
 
 		var targetPath string
-		if cfg.Name != "" {
-			targetPath = filepath.Join(targetDir, cfg.Name)
+		file := resolveFileName(p, cfg)
+		if !strings.Contains(file, "*") {
+			targetPath = filepath.Join(targetDir, file)
 		} else {
-			baseName := filepath.Base(p)
-			if !strings.Contains(baseName, "*") {
-				targetPath = filepath.Join(targetDir, baseName)
-			} else {
-				targetPath = targetDir + "/"
-			}
+			targetPath = targetDir + "/"
 		}
 		fmt.Fprintln(b, "cp -r", p, targetPath)
 	}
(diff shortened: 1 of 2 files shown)


frontend/rpm/template.go
```diff
@@ -537,3 +519,11 @@ func WriteSpec(spec *dalec.Spec, target string, w io.Writer) error {
 	}
 	return nil
 }
+
+func resolveFileName(path string, cfg dalec.ArtifactConfig) string {
```
adamperlin, line 523: in the systemd PR I actually added a `ResolveName` method to `ArtifactConfig` that does exactly this, maybe you could use that instead?

jeremyrickard: Thanks for pointing that out!

frontend/rpm/template_test.go
```diff
@@ -209,18 +209,116 @@ func TestTemplateSources(t *testing.T) {
...
-	want := `%postun
+		got := w.PostUn().String()
+		want := `%postun
 %systemd_postun test.service
 `
-	assert.Equal(t, want, got)
+		assert.Equal(t, want, got)
+	})
+
+	t.Run("test doc templaing using artifact config", func(t *testing.T) {
```
adamperlin, line 229: nit for these tests: s/templaing/templating

cpuguy83:
Looks good after @adamperlin's comment is fixed.

frontend/rpm/template_test.go
jeremyrickard: doh...sorry!

cpuguy83 (approved):

PR #272, opened 2024-05-28
Support SystemD Drop Ins

### LINKED ISSUE #271: [REQ] Support systemd config dropping

New feature

I would like to be able to specify systemd config files for use with systemd dropins. 

I should be able to specify the unit name to store the config files under and optionally rename the file. 

jeremyrickard:

This PR adds support for specifying systemd dropin config files. As part of this, I refactored the existing systemd block to nest units and dropins under a `systemd` field. 

This PR adds tests both to both the unit tests for template.go and for the az_linux tests. 

Fixes #271

### DIFF
--- a/test/azlinux_test.go
+++ b/test/azlinux_test.go
@@ -378,9 +378,11 @@ WantedBy=multi-user.target
 				},
 			},
 			Artifacts: dalec.Artifacts{
-				SystemdUnits: map[string]dalec.SystemdUnitConfig{
-					"src/simple.service": {
-						Enable: true,
+				SystemdConfigurations: &dalec.SystemdConfiguration{
+					SystemdUnits: map[string]dalec.SystemdUnitConfig{
+						"src/simple.service": {
+							Enable: true,
+						},
 					},
 				},
 			},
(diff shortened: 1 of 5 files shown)


test/azlinux_test.go
```diff
@@ -478,16 +484,29 @@ ListenStream=127.0.0.1:8080
...
-						Enable: true,
+				Systemd: &dalec.SystemdConfiguration{
+					Units: map[string]dalec.SystemdUnitConfig{
+						"src/foo.service": {},
+						"src/foo.socket": {
+							Enable: true,
+						},
+					},
+					Dropins: map[string]dalec.SystemdDropinConfig{
+						"src/foo.conf": {
```
adamperlin, line 507: Can we add another test config file here that is specifically for the `foo.socket` unit?

adamperlin (approved):
There's one more test that I think it might be nice to have, but it otherwise looks good

test/azlinux_test.go
jeremyrickard: done.

cpuguy83 (approved):

Issue #263, opened 2024-05-21
[REQ] Allow passing a config directly to the signer


cpuguy83:
### What kind of request is this?

None

### What is your request or suggestion?

Signers will often need some configuration passed along from the client (or alternatively they'd need to be built into the signer image, which isn't ideal for images).

Today a client can use `--build-context` to pass the signer a config, but the signer will need to know a specific key to look for. Also, these deals strictly with `llb.State`'s, so the signer will also need to know which file(s) in the passed in data is what.
Again, this can be convention, but can also be difficult to work with due to having to craft a context to look exactly like the signer is expecting.

I think we should allow the spec to declare what the config should be.
We should be able to take advantage of the `Source` type to declare this in the spec.

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

PR #275, opened 2024-05-29
Add support for passing build args to signers

cpuguy83:
I considered extending this to allow this for any kind of frontend (and as such sticking it into the `dalec.Frontend` type), however there really shouldn't be different vars passed to target frontends. Hence why I am restricting this to just signers for now.

If we decide to later we can add this more generically, however it will be much more difficult to take it away if we decide it was a bad idea.

Related to #263

### DIFF
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -426,7 +426,7 @@
 		"PackageConfig": {
 			"properties": {
 				"signer": {
-					"$ref": "#/$defs/Frontend",
+					"$ref": "#/$defs/PackageSigner",
 					"description": "Signer is the configuration to use for signing packages"
 				}
 			},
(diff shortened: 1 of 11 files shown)


pmengelbert (changes requested):
This looks good. Just to make sure I understand how dalec will behave after this PR is merged, is the following accurate?

1. Values needed for the signing image can be provided via the `package_config` section, either on the spec level, or on the level of a specific target. (good for packages with both linux + windows targets).
2. These values can be supplied at build-invocation time by leveraging the spec's `args`.
3. The args will be read by the consumer (i.e. the signing image) via the filesystem, and will have to be initialized into the environment on that end.

The only actionable request I have is that this should be captured in an example somewhere (in the `docs/examples` directory perhaps). As we keep adding features like this, having an up-to-date example with the kitchen sink makes it easier for all the developers on the project to track spec changes.

cpuguy83:
Updated with doc updates.

pmengelbert:
I think this needs to be rebased, then it should pass the CI

pmengelbert (commented):
LGTM, will wait for rebase + CI passage to approve.

cpuguy83:
This was rebased, FYI.

pmengelbert (approved):
LGTM, thanks

PR #279, opened 2024-06-04
Add Target to Extract Binary Artifacts from RPMs

adamperlin:
There is a desire from some Dalec users to extract the binary artifacts produced as part of Dalec's rpm builds as standalone files.  This PR adds a target which produces a single zip file containing the binary artifacts specified in a Dalec spec. Only the binaries are included with no dependencies whatsoever, so it is the user's responsibility to ensure that any standalone binaries extracted with this target are fully statically linked. 

Fixes #


adamperlin:
Haven't determined the source of the CI failures yet, it appears to be some kind of timeout

cpuguy83:
Rebased to include #264

cpuguy83:
Rebased this on top of #280 and tests are passing now.

adamperlin:
> Rebased this on top of #280 and tests are passing now.

Thank you!!

frontend/azlinux/handle_bin.go
```diff
@@ -0,0 +1,121 @@
...
+	return sb.String()
+}
+
+func zip(worker llb.State, zipName string, outputDir string, artifacts llb.State) llb.State {
+	outName := filepath.Join(outputDir, zipName+".zip")
+	return worker.Run(
+		shArgs("zip "+outName+" *"),
+		llb.Dir("/tmp/artifacts"),
+		llb.AddMount("/tmp/artifacts", artifacts)).AddMount(outputDir, llb.Scratch())
+}
```
DannyBrito, line 60: Can't this `zip` function be deleted and use directly the `Zip` from `dalec` package in the `helpers.go` file? These seem to be the same, no?

adamperlin: Ah yes thank you, I think I meant to delete this

frontend/azlinux/handle_bin.go
```diff
@@ -0,0 +1,121 @@
...
+				llb.AddMount("/script", scriptState),
+				llb.AddMount("/package", rpmState),
+				llb.AddMount("/extracted", llb.Scratch()),
+			).AddMount("/out", llb.Scratch())
+
+			pg = dalec.ProgressGroup("Compressing artifacts: ")
+
+			zipWorker := w.Base(client, pg).Run(w.Install([]string{"zip"})).Root()
+
+			zipped := zip(zipWorker, "binaries", "/out", st)
```
adamperlin, line 99: Would we prefer `.zip` or `.tar.gz`?

test/bin_extract_test.go
```diff
@@ -0,0 +1,271 @@
...
+	"os"
+	"testing"
+
+	"github.com/Azure/dalec"
+	"github.com/Azure/dalec/frontend/pkg/bkfs"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBinExtract_Mariner2(t *testing.T) {
```
cpuguy83, line 19: We should be able to move these tests directly into the distro test.

cpuguy83: Reason being this makes sure when we add new targets we don't forget to add these extra tests.

test/bin_extract_test.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		t.Fatalf("unable to read zip: %s", err)
+	}
+
+	zr, err := zip.NewReader(bytes.NewReader(read), info.Size())
+	if err != nil {
+		t.Fatal("unable to create zip reader %w", err)
+	}
+
+	for _, file := range zr.File {
```
cpuguy83, line 52: All the assertions in here should include the file name.

test/bin_extract_test.go
```diff
@@ -0,0 +1,271 @@
...
+			Artifacts: dalec.Artifacts{
+				Binaries: map[string]dalec.ArtifactConfig{
+					"src/phony.sh": {},
+				},
+			},
+		}
+
+		testEnv.RunTest(ctx, t, func(ctx context.Context, client gwclient.Client) (*gwclient.Result, error) {
+			sr := newSolveRequest(withBuildTarget(buildTarget), withSpec(ctx, t, spec))
+			sr.Evaluate = true
```
cpuguy83, line 100: IIRC evaluate is being set to true by default now.

test/bin_extract_test.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		t.Fatalf("unable to stat zip: %s", err)
+	}
+
+	read, err := io.ReadAll(zf)
+	if err != nil {
+		t.Fatalf("unable to read zip: %s", err)
+	}
+
+	zr, err := zip.NewReader(bytes.NewReader(read), info.Size())
```
cpuguy83, line 47: Should be able to drop the `fs.File` into `zip.NewReader` directly.

test/bin_extract_test.go
```diff
@@ -0,0 +1,271 @@
...
+type expectFile struct {
+	contents    string
+	permissions os.FileMode
+}
+
+func assertZipContentsMatch(t *testing.T, zf fs.File, want map[string]expectFile) {
+	info, err := zf.Stat()
+	if err != nil {
+		t.Fatalf("unable to stat zip: %s", err)
+	}
```
cpuguy83, line 40:
```suggestion
	assert.NilError(t, err)
```

test/bin_extract_test.go
```diff
@@ -0,0 +1,271 @@
...
+
+	zr, err := zip.NewReader(bytes.NewReader(read), info.Size())
+	if err != nil {
+		t.Fatal("unable to create zip reader %w", err)
+	}
+
+	for _, file := range zr.File {
+		f, err := file.Open()
+		assert.Nil(t, err)
```
cpuguy83, line 54:
```suggestion
		assert.NilError(t, err)
```

test/bin_extract_test.go
```diff
@@ -0,0 +1,271 @@
...
+		contents, err := io.ReadAll(f)
+
+		assert.Nil(t, err)
```
cpuguy83, line 57:
```suggestion
		assert.NilError(t, err)
```

test/bin_extract_test.go
```diff
@@ -0,0 +1,271 @@
...
+
+	read, err := io.ReadAll(zf)
+	if err != nil {
+		t.Fatalf("unable to read zip: %s", err)
+	}
+
+	zr, err := zip.NewReader(bytes.NewReader(read), info.Size())
+	if err != nil {
+		t.Fatal("unable to create zip reader %w", err)
+	}
```
cpuguy83, line 50:
```suggestion
	assert.NilError(t, err)
```

frontend/azlinux/handle_bin.go
cpuguy83: I don't think we need to do anything with the binaries, we can just keep them in a raw llb.State.
Same for Windows.

cpuguy83 (changes requested):

frontend/azlinux/handler.go
```diff
@@ -33,6 +33,11 @@ func newHandler(w worker) gwclient.BuildFunc {
 	})
 	mux.Add("rpm/debug", handleDebug(w), nil)
 
+	mux.Add("rpm/bin", handleBin(w), &targets.Target{
```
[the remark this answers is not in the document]
adamperlin: Just to clarify are you suggesting renaming `windows/zip` to `windows/artifacts/binaries` then?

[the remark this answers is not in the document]
cpuguy83: I am suggesting to add a new target `windows/artifacts/binaries`

test/bin_extract_test.go
adamperlin: I don't think that method exists with the testify library

cpuguy83: Ah that explains it, we have 2 assertion libraries in the repo.

adamperlin: Oh interesting. Perhaps we should pare it down to one?

frontend/azlinux/handle_bin.go
```diff
@@ -0,0 +1,107 @@
...
+	sb.WriteString(`
+#!/bin/bash
+set -e
+declare -a RPMS=()
+export RPM_BINDIR=$(rpm --eval '%{_bindir}')
+`)
+
+	for _, rpm := range rpms {
+		sb.WriteString(fmt.Sprintf("RPMS+=(%q)\n", rpm))
+	}
```
pmengelbert, line 29: Bash has built-in ways of building up arrays from text. My personal choice is to use [readarray](https://helpmanual.io/builtin/readarray/) (an alias of `mapfile`). You have to be sure you're using bash and not some other shell, and you have to be sure that your bash version is >= 4 (almost all distros these days).

You could get the data into the container by
```go
rpmsData := []byte(strings.Join(rpms, "\x00")) // null separate the rpms
```

Then, you can take those bytes and use `llb.MkFile` to mount it into the container, for example at `/tmp/rpm_list`. In bash, you would do:

```bash
readarray -d '' RPMS < /tmp/rpm_list # reads the file and puts the results into the array variable "$RPMS"
```

This note is for educational purposes, what you have is fine.

frontend/azlinux/handle_bin.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func binCopyScript(rpms []string, binaries map[string]dalec.ArtifactConfig) string {
+	var sb strings.Builder
+
+	sb.WriteString(`
+#!/bin/bash
+set -e
+declare -a RPMS=()
+export RPM_BINDIR=$(rpm --eval '%{_bindir}')
```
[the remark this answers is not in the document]
adamperlin: So just to confirm, are we ok with having this just extract from `%{_bindir}` (and any subpaths) or do we want to look in any additional locations?

frontend/azlinux/handle_bin.go
adamperlin: Thanks for this, I am not super familiar with bash array building capabilities, so I appreciate it!

frontend/azlinux/handle_bin.go
```diff
@@ -0,0 +1,107 @@
...
+
+	for _, rpm := range rpms {
+		sb.WriteString(fmt.Sprintf("RPMS+=(%q)\n", rpm))
+	}
+
+	sb.WriteString("for rpm in $RPMS; do\n")
+	binaryPathList := make([]string, 0, len(binaries))
+	for path, bin := range binaries {
+		srcPath := bin.InstallPath(path)
+		binaryPathList = append(binaryPathList, filepath.Join(".$RPM_BINDIR", srcPath))
```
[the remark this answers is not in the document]
adamperlin: Changed this. I discovered that `filepath.Join` calls  `filepath.Clean` which will strip off a leading `./`, so the `./` must be appended after the call

frontend/azlinux/handle_bin.go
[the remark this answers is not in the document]
pmengelbert: > So just to confirm, are we ok with having this just extract from `%{_bindir}`

Yes, unless I've misunderstood.

pmengelbert (approved):
LGTM, great work on this

frontend/azlinux/handle_bin.go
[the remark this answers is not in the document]
cpuguy83: Not ignoring this just need some time to think about the edge cases.

adamperlin:
Thanks for the review and sign off @pmengelbert! After a bit of offline discussion today, I think we've decided to put this PR on hold for now as it may not actually be needed to serve the use case

cpuguy83:
We discussed this and are thinking this should be handled outside of dalec.
It is difficult to solve all use cases with this and do not want to encourage people to extract the binaries (which are typically going to be dynamically linked anyway).

adamperlin:
Closing this due to above discussion.

PR #281, opened 2024-06-05
Add spec load pass to check for undefined build args

### LINKED ISSUE #274: [BUG] Argument substitution should handle undefined args used for substitution
### Expected Behavior

Given some substituion like `$FOO`, if `$FOO` is not defined dalec should error out similar to how bash with `set -u` does for unknown vars.
This can help prevent user errors (e.g. mispells) from going unnoticed.

### Actual Behavior

Currently this will give you an empty string when there is no matching substitution.

### Steps To Reproduce

_No response_

adamperlin:
Currently, if a build arg is not specified in the `args` section of a Dalec spec file, its value is set to `""` when the spec is loaded. This is confusing behavior, as a user could easily forget to declare a build-arg or mistype a build-arg reference, leading to an empty string substitution which could be hard to track down.

This PR adds an additional pass upon loading the spec which tokenizes all the build args first, and checks to ensure that each referenced arg does appear in the `args` section of the spec, raising an error if not. 

Fixes #274 


cpuguy83:
Did you happen to look into using https://pkg.go.dev/github.com/moby/buildkit/frontend/dockerfile/shell?utm_source=godoc#Lex.ProcessWordWithMatches instead?
I believe that should give us the list of vars which weren't in our env list.

adamperlin:
Ah no, I was not aware that function existed. I will see if I can reimplement using that instead

adamperlin:
It looks like the version of this function you referenced exists in a newer version of buildkit (0.14.0) than the one we import in dalec (0.13.1). Are we ok to bump the buildkit dependency?

cpuguy83:
Oh nice, yes we can/should bump it.

load.go
```diff
@@ -38,67 +38,75 @@ func platformArg(key string) bool {
...
 const DefaultPatchStrip int = 1
 
-func (s *Source) substituteBuildArgs(args map[string]string) error {
+func expandArgs(lex *shell.Lex, s string, args map[string]string) (string, *ErrorList) {
+	result, err := lex.ProcessWordWithMatches(s, args)
+	if err != nil {
+		return "", AsList(err)
+	}
+
+	var errL ErrorList
```
cpuguy83, line 47: You can use `errors.Join` to do this without needing a separate type.

load.go
```diff
@@ -471,22 +490,22 @@ func (c *TestSpec) processBuildArgs(lex *shell.Lex, args map[string]string, name
...
-		step.Stderr = stderr
 
+		step.Stderr = stderr
 		c.Steps[i] = step
 	}
 
 	for name, f := range c.Files {
 		if err := f.processBuildArgs(lex, args); err != nil {
-			return errors.Wrap(err, name)
+			errL.Append(errors.Wrap(err, name))
```
cpuguy83, line 508: The error is never returned in this case now.

cpuguy83 (changes requested):

load.go
adamperlin: Thank you for catching this!

load.go
```diff
@@ -44,67 +44,96 @@ func platformArg(key string) bool {
...
+func expandArgs(lex *shell.Lex, s string, args map[string]string) (string, error) {
+	result, err := lex.ProcessWordWithMatches(s, args)
+	if err != nil {
+		return "", err
+	}
+
+	var errs []error
+	for m := range result.Unmatched {
+		if !knownArg(m) {
+			errs = append(errs, fmt.Errorf(`build arg "%s" not declared`, m))
```
cpuguy83, line 56: Nit, this can be a continue here to remove the `else if` to just an `if`

cpuguy83 (changes requested):
MInor readability nit, but otherwise LGTM

cpuguy83 (approved):

Issue #177, opened 2024-03-17
Optimization: Use pre-built mariner builder image


cpuguy83:
### What kind of request is this?

Improvement of existing experience

### What is your request or suggestion?

Currently on a fresh buildkit instance we spend a few minutes preparing the mariner2 build environment by taking the base mariner image and installing rpmbuild and some other tooling.
It would be nice if this was available as a pre-built image so that this is not required.

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

cpuguy83:
This is done! It does require pulling those worker images into your build context.
https://project-dalec.github.io/dalec/targets#worker-images

PR #286, opened 2024-06-10
[WIP] add prebuild work images

DannyBrito:
Initial work to add pre-build images for mariner, azlinux worker images.

Fixes # https://github.com/Azure/dalec/issues/177

Questions:
1. Would these work images need to be versioned? If so, would these follow same versioning of `dalec` tags?
2. Would we need to add a ci for these images?

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -61,6 +61,16 @@ variable "RUNC_REVISION" {
     default = "1"
 }
 
+target "dalec-workers" {
+    name = "dalec-${distro}-worker"
+    matrix = {
+        distro = ["mariner2", "azlinux3"]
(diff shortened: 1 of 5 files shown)


pmengelbert:
Can you update the PR description with some detail on what problem this is solving?

cpuguy83:
In https://github.com/Azure/dalec/pull/328 we added support for:

1. Supplying a custom worker image by providing a specifically named build context (name depends on the target)
2. Building/exporting just the worker image as a build target, e.g. `mariner2/worker`

With these we have these pre-cached worker images without having to maintain a custom set in the repo (though we still could do this, just using the new build endpoints).

PR #288, opened 2024-06-13
delete unused test function parameter

DannyBrito:

delete unused test function parameter


cpuguy83 (approved):
Thanks!

PR #290, opened 2024-06-17
rpm: Fix use of inccorrect field for runtime constraints

cpuguy83:
When building the rpm spec we were using the wrong set of constraints for runtime dependencies.
This was likely just a bad copy/paste.

Added tests for this.


adamperlin (approved):
LGTM, pretty straightforward fix

Issue #298, opened 2024-06-24
[BUG] minimal base is not used


sozercan:
### Expected Behavior

minimal base image with only the component we are building and/or supplied in runtime

### Actual Behavior

i expected a minimal base image but looking at the /usr/bin contents, there are a lot of files

https://oci.dag.dev/layers/sozercan/oras@sha256:0137bd96038e5a7edd4c3114e45997b27dcaed83b906230a3943b1c161b76d8b/usr/bin/

### Steps To Reproduce

```bash
docker build -t oras:latest -f  
https://gist.githubusercontent.com/sozercan/1f856f629ba85b9a223dfff1f5cb5f2a/raw/794e8ebf0fb0c9d5581050a51e8a70156c901340/oras.yaml
--load . --target mariner2/container --progress plain
```

### Are you willing to submit PRs to contribute to this bug fix?

- [ ] Yes, I am willing to implement it.

pmengelbert:
This bug was introduced with PR #254 . Specifically, it's this commit [https://github.com/adamperlin/dalec/commit/c91c9bae5812aacf34c8e2aedbef118e047658c3](c91c9bae5812aacf34c8e2aedbef118e047658c3), although I can't yet fathom why

pmengelbert:
What is happening is that the presence of the `%post`, `%preun`, or `%postun` causes `/bin/sh` to be baked into the dependencies of the rpm. This makes sense because a shell is needed to execute the postinstall scripts, and would be needed to run pre- or post- uninstall scripts.

_without %post_:
$ rpm -q --requires /tmp/out/RPMS/x86_64/oras-v1.2.0-1.cm2.x86_64.rpm
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsZstd) <= 5.4.18-1
```

_with %post_:
$ rpm -q --requires /tmp/out/RPMS/x86_64/oras-v1.2.0-1.cm2.x86_64.rpm
/bin/sh
/bin/sh
/bin/sh
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsZstd) <= 5.4.18-1
```

[...]

PR #299, opened 2024-06-26
fix: Don't emit rpm spec fields if empty in dalec spec

pmengelbert:
This PR prevents the fields `%post`, `%preun`, and `%postun` from being written to the rpm SPEC unless they are specified in the dalec spec.

This is a short-term solution to the problem specified in #298. Please see that issue for more details on the long-term solution. A short summary of the problem follows:

What is happening is that the presence of the `%post`, `%preun`, or `%postun` causes `/bin/sh` to be baked into the dependencies of the rpm. This makes sense because a shell is needed to execute the postinstall scripts, and would be needed to run pre- or post- uninstall scripts.

_without %post_:
```
$ rpm -q --requires /tmp/out/RPMS/x86_64/oras-v1.2.0-1.cm2.x86_64.rpm
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsZstd) <= 5.4.18-1
```

_with %post_:
```
$ rpm -q --requires /tmp/out/RPMS/x86_64/oras-v1.2.0-1.cm2.x86_64.rpm
/bin/sh
/bin/sh
/bin/sh
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsZstd) <= 5.4.18-1
```

The `bash` package supplies `/bin/sh`, and all of its dependencies are installed into the container as well. So the distroless minimal image is used, but it has a bunch of extra stuff installed.

Fixes #

### DIFF
--- a/frontend/rpm/template.go
+++ b/frontend/rpm/template.go
@@ -294,7 +294,7 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
 func (w *specWrapper) PreUn() fmt.Stringer {
 	b := &strings.Builder{}
 
-	if dalec.SystemdIsEmpty(w.Artifacts.Systemd) {
+	if w.Artifacts.Systemd.IsEmpty() {
 		return b
 	}
 
@@ -311,7 +311,7 @@ func (w *specWrapper) PreUn() fmt.Stringer {
 func (w *specWrapper) Post() fmt.Stringer {
 	b := &strings.Builder{}
 
-	if dalec.SystemdIsEmpty(w.Artifacts.Systemd) {
+	if w.Artifacts.Systemd.IsEmpty() {
 		return b
 	}
 
@@ -330,7 +330,7 @@ func (w *specWrapper) Post() fmt.Stringer {
 func (w *specWrapper) PostUn() fmt.Stringer {
 	b := &strings.Builder{}
 
-	if dalec.SystemdIsEmpty(w.Artifacts.Systemd) {
+	if w.Artifacts.Systemd.IsEmpty() {
 		return b
 	}
 
--- a/frontend/rpm/template.go
+++ b/frontend/rpm/template.go
@@ -293,14 +293,20 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
 
 func (w *specWrapper) PreUn() fmt.Stringer {
 	b := &strings.Builder{}
-	b.WriteString("%preun\n")
 
-	if w.Spec.Artifacts.Systemd != nil {
-		keys := dalec.SortMapKeys(w.Spec.Artifacts.Systemd.Units)
-		for _, servicePath := range keys {
-			serviceName := filepath.Base(servicePath)
-			fmt.Fprintf(b, "%%systemd_preun %s\n", serviceName)
-		}
+	if w.Spec.Artifacts.Systemd == nil {
+		return b
+	}
+
+	if len(w.Spec.Artifacts.Systemd.Units) == 0 {
+		return b
+	}
+
+	b.WriteString("%preun\n")
+	keys := dalec.SortMapKeys(w.Spec.Artifacts.Systemd.Units)
+	for _, servicePath := range keys {
+		serviceName := filepath.Base(servicePath)
+		fmt.Fprintf(b, "%%systemd_preun %s\n", serviceName)
 	}
 
 	return b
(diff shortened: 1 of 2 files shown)


frontend/rpm/template.go
```diff
@@ -293,45 +293,66 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
...
 	b := &strings.Builder{}
-	b.WriteString("%preun\n")
 
-	if w.Spec.Artifacts.Systemd != nil {
-		keys := dalec.SortMapKeys(w.Spec.Artifacts.Systemd.Units)
-		for _, servicePath := range keys {
-			serviceName := filepath.Base(servicePath)
-			fmt.Fprintf(b, "%%systemd_preun %s\n", serviceName)
-		}
+	if w.Spec.Artifacts.Systemd == nil {
```
cpuguy83, line 297: Nit: we can add a function like `w.Spec.Artifacts.Systemd.IsEmpty()`
That function can also do its own nil check.

cpuguy83 (changes requested):
Looks good, just a minor suggestion.

frontend/rpm/template.go
```diff
@@ -293,45 +294,78 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
...
+		return b
+	}
+
+	if len(w.Spec.Artifacts.Systemd.Units) == 0 {
+		return b
+	}
+
+	if strings.HasSuffix(w.Target, containerSuffix) {
```
adamperlin, line 306: I thought we'd decided we weren't going to have different behavior based on the target?

pmengelbert: I forgot to push my revert commit! Should be fixed now

frontend/rpm/template.go
pmengelbert: If `w.Spec.Artifacts.Systemd` is nil, the method `IsEmpty` can't be called on it, so it'll have to be a function and not a method.

pmengelbert: done

cpuguy83: You can call a method on a nil pointer:

https://go.dev/play/p/BoM5AGGb0Lt

sozercan:
after this merges, can we cherry pick and create a patch release?

spec.go
```diff
@@ -682,3 +682,15 @@ type CheckOutputError struct {
...
+	if s == nil {
+		return true
+	}
+
+	if len(s.Units) == 0 {
+		return true
+	}
+
+	return false
+}
```
cpuguy83, line 696:
```suggestion
func (s *SystemdConfiguration) IsEmpty() bool {
	if s == nil {
		return true
	}

	if len(s.Units) == 0 {
		return true
	}

	return false
}
```

frontend/rpm/template.go
```diff
@@ -293,45 +293,54 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
...
+		fmt.Fprintf(b, "%%systemd_post %s\n", unitConf.ResolveName(servicePath))
 	}
+
 	return b
 }
 
 func (w *specWrapper) PostUn() fmt.Stringer {
 	b := &strings.Builder{}
+
+	if dalec.SystemdIsEmpty(w.Artifacts.Systemd) {
```
cpuguy83, line 333:
```suggestion
	if w.Artifacts.Systemd.IsEmpty() {
```

frontend/rpm/template.go
```diff
@@ -293,45 +293,54 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
...
+		fmt.Fprintf(b, "%%systemd_preun %s\n", serviceName)
 	}
 
 	return b
 }
 
 func (w *specWrapper) Post() fmt.Stringer {
 	b := &strings.Builder{}
+
+	if dalec.SystemdIsEmpty(w.Artifacts.Systemd) {
```
cpuguy83, line 314:
```suggestion
	if w.Artifacts.Systemd.IsEmpty() {
```

frontend/rpm/template.go
```diff
@@ -293,45 +293,54 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
...
 	b := &strings.Builder{}
-	b.WriteString("%preun\n")
 
-	if w.Spec.Artifacts.Systemd != nil {
-		keys := dalec.SortMapKeys(w.Spec.Artifacts.Systemd.Units)
-		for _, servicePath := range keys {
-			serviceName := filepath.Base(servicePath)
-			fmt.Fprintf(b, "%%systemd_preun %s\n", serviceName)
-		}
+	if dalec.SystemdIsEmpty(w.Artifacts.Systemd) {
```
cpuguy83, line 297:
```suggestion
	if w.Artifacts.Systemd.IsEmpty() {
```

cpuguy83 (changes requested):

cpuguy83 (approved):

cpuguy83:
@sozercan 0.6.0 is released with this fix.

frontend/rpm/template.go
pmengelbert: > You can call a method on a nil pointer:
> 
> https://go.dev/play/p/BoM5AGGb0Lt

TIL, thanks!

PR #300, opened 2024-06-26
Fix cmd mounts

cpuguy83:
Adds additional validations to sources to prevent hitting cryptic errors, or worse unexpected data in a specified source.

### DIFF
--- a/source.go
+++ b/source.go
@@ -200,6 +200,13 @@ func generateSourceFromImage(st llb.State, cmd *Command, sOpts SourceOpts, subPa
 		return llb.Scratch(), fmt.Errorf("no steps defined for image source")
 	}
 
+	if subPath == "" {
+		// TODO: We should log a warning here since extracing an entire image while also running a command is
+		// probably not what the user really wanted to do here.
+		// The buildkit client provides functionality to do this we just need to wire it in.
+		subPath = "/"
+	}
+
 	for k, v := range cmd.Env {
 		st = st.AddEnv(k, v)
 	}
--- a/source.go
+++ b/source.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"strings"
 
 	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/identity"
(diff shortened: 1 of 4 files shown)


source.go
```diff
@@ -193,13 +194,33 @@ func (s *Source) AsMount(name string, sOpt SourceOpts, opts ...llb.ConstraintsOp
 	return s.asState(name, true, sOpt, opts...)
 }
 
+var errInvalidMountConfig = errors.New("invalid mount config")
+
+func (m *SourceMount) validate(root string) error {
+	if m.Dest == "/" {
+		return errors.Wrap(errInvalidMountConfig, "mount destination must not be \"/\"")
+	}
+	if root != "/" && filepath.HasPrefix(m.Dest, root) {
```
adamperlin, line 203: Need to change from `HasPrefix` or ignore linter it looks like

adamperlin, line 203: Checking the output of `filepath.Rel` is a possible alternative?

adamperlin (changes requested):
I think the `HasPrefix` usage should maybe be changed? Otherwise, it looks good

source.go
cpuguy83: `filepath.Rel` doesn't really work here because it will always give you the combined path regardless of if they are related.

I'ave added a more thorough implementation that deals with a few different things.

source_test.go
```diff
@@ -1010,3 +1041,42 @@ func (stubMetaResolver) ResolveImageConfig(ctx context.Context, ref string, opt
...
+		{"/foo/", "/foo", true},
+		{"/foo/", "/foo/", true},
+		{"//foo", "/foo", true},
+		{"//foo/", "/foo", true},
+		{"/foo/bar", "/foo", true},
+		{"/foo/bar/", "/foo", true},
+		{"/foo/bar/", "/foo/", true},
+		{"/foo/bar", "/foo/", true},
+		{"/foo/bar", "/bar", false},
+		{"/foo/bar", "/foo/bar/baz", false},
```
adamperlin, line 1063: maybe you could add:
```suggestion
		{"/foo/bar/baz", "/foo/bar", true},
```

adamperlin: meant to be an addition, not an edit

cpuguy83: Added

adamperlin (approved):

PR #304, opened 2024-07-01
Refactor a bunch of things

cpuguy83:
This is a bunch of refactoring I've pulled out of my jammy PR which will help to get in so that there aren't so many conflicts from upcoming changes.

### DIFF
--- a/frontend/rpm/handle_sources.go
+++ b/frontend/rpm/handle_sources.go
@@ -3,7 +3,6 @@ package rpm
 import (
 	"context"
 	"fmt"
-	"path/filepath"
 
 	"github.com/Azure/dalec"
 	"github.com/Azure/dalec/frontend"
@@ -12,35 +11,6 @@ import (
 	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 )
-
-// TarImageRef is the image used to create tarballs of sources
-// This is purposefully exported so it can be overridden at compile time if needed.
-// Currently this image needs /bin/sh and tar in $PATH
-var TarImageRef = "busybox:latest"
-
-func shArgs(cmd string) llb.RunOption {
-	return llb.Args([]string{"sh", "-c", cmd})
-}
-
-func tar(src llb.State, dest string, opts ...llb.ConstraintsOpt) llb.State {
-	tarImg := llb.Image(TarImageRef, dalec.WithConstraints(opts...))
-
-	// Put the output tar in a consistent location regardless of `dest`
-	// This way if `dest` changes we don't have to rebuild the tarball, which can be expensive.
-	outBase := "/tmp/out"
-	out := filepath.Join(outBase, filepath.Dir(dest))
-	worker := tarImg.Run(
-		llb.AddMount("/src", src, llb.Readonly),
-		shArgs("tar -C /src -cvzf /tmp/st ."),
-		dalec.WithConstraints(opts...),
-	).
-		Run(
-			shArgs("mkdir -p "+out+" && mv /tmp/st "+filepath.Join(out, filepath.Base(dest))),
-			dalec.WithConstraints(opts...),
-		)
-
-	return worker.AddMount(outBase, llb.Scratch())
-}
 
 func HandleSources(wf WorkerFunc) gwclient.BuildFunc {
 	return func(ctx context.Context, client gwclient.Client) (*gwclient.Result, error) {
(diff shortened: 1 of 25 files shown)


frontend/rpm/handle_sources.go
```diff
@@ -85,39 +55,35 @@ func HandleSources(wf WorkerFunc) gwclient.BuildFunc {
...
 	st, err := spec.GomodDeps(sOpt, worker, opts...)
 	if err != nil {
 		return nil, errors.Wrap(err, "error adding gomod sources")
 	}
 
-	if st != nil {
-		out = append(out, tar(*st, gomodsName+".tar.gz", opts...))
+	sorted := dalec.SortMapKeys(sources)
+	for _, k := range sorted {
+		st := sources[k]
```
DannyBrito, line 72: Maybe, consider keeping the `dalec.ProgressGroup("Add spec source: " + k)` for a bit more visibility/logging?

frontend/rpm/handle_sources.go
```diff
@@ -85,39 +55,35 @@ func HandleSources(wf WorkerFunc) gwclient.BuildFunc {
...
+		if dalec.SourceIsDir(spec.Sources[k]) {
+			st = st.With(sourceTar(worker, k, opts...))
+		}
+		out = append(out, st)
 	}
 
+	if st != nil {
+		out = append(out, st.With(sourceTar(worker, gomodsName, opts...)))
+	}
```
DannyBrito, line 81: I am assuming the order of this appending matters of this state. Because it is set before the loop and the append after the loop, is this correct? If order doesn't matter maybe, could you move the `st` value setting right before the `st` appending to the out?

helpers.go
```diff
@@ -316,28 +319,53 @@ func (s *Spec) GetTestDeps(targetKey string) []string {
...
 }
 
-func (s *Spec) GetSymlinks(target string) map[string]SymlinkTarget {
-	lm := make(map[string]SymlinkTarget)
-
-	if s.Image != nil && s.Image.Post != nil && s.Image.Post.Symlinks != nil {
-		for k, v := range s.Image.Post.Symlinks {
-			lm[k] = v
+func (s *Spec) GetImagePost(target string) *PostInstall {
+	img := s.Targets[target].Image
```
DannyBrito, line 323: is there any change that the arg `target` is not existing withing `s.Targets`? If so maybe we should check before using `.Image`?

source.go
```diff
@@ -574,3 +570,45 @@ func (s *Spec) getPatchedSources(sOpt SourceOpts, worker llb.State, filterFunc f
 
 	return PatchSources(worker, s, states, opts...), nil
 }
+
+// Tar creates a tar+gz from the provided state and puts it in the provided dest.
+// The provided work state is used to perform the necessary operations to produce the tarball and requires the tar and gzip binaries.
+func Tar(work, src llb.State, dest string, opts ...llb.ConstraintsOpt) llb.State {
```
DannyBrito, line 576: could you add the type for the `work` arg?

source.go
```diff
@@ -574,3 +570,45 @@ func (s *Spec) getPatchedSources(sOpt SourceOpts, worker llb.State, filterFunc f
...
+	// Put the output tar in a consistent location regardless of `dest`
+	// This way if `dest` changes we don't have to rebuild the tarball, which can be expensive.
+	outBase := "/tmp/out"
+	out := filepath.Join(outBase, filepath.Dir(dest))
+	worker := work.Run(
+		llb.AddMount("/src", src, llb.Readonly),
+		llb.Args([]string{"/bin/sh", "-c", "tar -C /src -cvzf /tmp/st ."}),
```
DannyBrito, line 583: q: could this use the helper function `ShArgs`? same question for `587`

test/azlinux_test.go
```diff
@@ -875,23 +882,19 @@ Environment="KUBELET_KUBECONFIG_ARGS=--bootstrap-kubeconfig=/etc/kubernetes/boot
...
 					Name: "Doc files should be created in correct place",
 					Files: map[string]dalec.FileCheckOutput{
-						"/usr/share/doc/test-docs-handled/src1":                      {},
-						"/usr/share/doc/test-docs-handled/subpath/src2":              {},
-						"/usr/share/licenses/test-docs-handled/src3":                 {},
-						"/usr/share/licenses/test-docs-handled/license-subpath/src4": {},
+						"/usr/share/doc/test-docs-handled/src1":                                         {},
+						"/usr/share/doc/test-docs-handled/subpath/src2":                                 {},
+						filepath.Join(testConfig.LicenseDir, "test-docs-handled/src3"):                  {},
+						filepath.Join(testConfig.LicenseDir, "/test-docs-handled/license-subpath/src4"): {},
```
DannyBrito, line 888: nit:
```suggestion
						filepath.Join(testConfig.LicenseDir, "test-docs-handled/src3"):                  {},
						filepath.Join(testConfig.LicenseDir, "test-docs-handled/license-subpath/src4"): {},
```

frontend/rpm/handle_sources.go
cpuguy83: The reason I am creating it before taring up all the sources is because it can error out and I'd rather error early rather than generate a bunch of other state unnecessarily.

helpers.go
cpuguy83: `s.Targets[target]` is never nil here.

source.go
cpuguy83: Any particular reason?

cpuguy83: I can, actually I assumed the linter would complain with that but it doesn't seem to care.

frontend/rpm/handle_sources.go
cpuguy83: Done (though slightly differently).

source.go
cpuguy83: Done.

test/azlinux_test.go
cpuguy83: Done.

source.go
DannyBrito: not strong reason just `nit` for style consistency

DannyBrito (approved):
lgtm

PR #305, opened 2024-07-02
Fix source file from dir

### LINKED ISSUE #188: [BUG] Source paths re not checked to determine if the path is a dir or file

cpuguy83:
Allow patch sources to be a dir-backed source with a subpath pointing to a file.

Closes #191
Closes #188


website/docs/sources.md
```diff
@@ -28,6 +28,12 @@ The below example uses a [`context`](#build-context) source type.
 The root of the source is the `path/in/source` directory.
 The source will include all `.txt` files within `path/in/source` except for `secret.txt`.
 
+For each source type, the source may be treated as a directory or a file, which
+changes the handling of the source.
+The `path` field is also used to determine that.
+If `path` points at a file in the source then the source is always considered a
+file regardless of source type.
```
DannyBrito, line 35: Could you please add something about determining if is a directory or file is really only based on the path `field` and the `include` and `exclude` are not taken into account?

frontend/rpm/template.go
```diff
@@ -226,7 +232,10 @@ func (w *specWrapper) PrepareSources() (fmt.Stringer, error) {
 				return nil
 			}
```
DannyBrito, line 233: would we still need to handle when the case is a patch and a directory when preparing sources?

cpuguy83: We just don't support that right now and it will give an error.
We can handle this in a follow-up.

website/docs/sources.md
cpuguy83: Added an explicit callout for include and exclude.

DannyBrito (approved):
lgtm

PR #306, opened 2024-07-02
Use strong type for package constraints

cpuguy83:
This is another commit I've pulled out for of the jammy.
It allows constraints to apply to specific architectures+version combos which is something I realized we were missing while going through the debian control file spec.

### DIFF
--- a/frontend/rpm/template_test.go
+++ b/frontend/rpm/template_test.go
@@ -290,6 +290,7 @@ fi
 %doc %{_docdir}/test-pkg/docs/README
 
 `
+
 		assert.Equal(t, want, got)
 	})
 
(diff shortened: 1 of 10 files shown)


adamperlin:
So, to clarify my understanding of the current behavior, it looks like if `arch` is speciifed, the listed version constraint only applies to specific platforms? My question is, what if for the same package we want to specify **different** version constraints for different architectures, i.e., for `my-package` we have a requirement of `>1.1` for x86 and `>1.2` for arm. Would this be supported right now / do we want to support this?

cpuguy83:
I considered this, and even implemented it, but decided its not worth the complexity given how unlikely it is to have different version constraints on the same package based on arch.
It would be more likely that it would be different packages for different arches.

frontend/rpm/template_test.go
```diff
@@ -426,24 +426,28 @@ func TestTemplate_Requires(t *testing.T) {
...
-			Runtime: map[string][]string{
+			},
+			Runtime: map[string]dalec.PackageConstraints{
 				"a-no-constraints": {},
 				"b-one-constraints": {
-					"< 2.0",
+					Version: []string{"< 2.0"},
 				},
 				"c-multiple-constraints": {
+					Version: []string{
```
adamperlin, line 447: can we add arch-specific constraints to one of these tests?

adamperlin (changes requested):
Can we add unit tests for constraints which only apply to specific architectures?

frontend/rpm/template_test.go
cpuguy83: huh, I'm confused.
I'm 100% positive I wrote more tests here.

cpuguy83: (but they are not here...)

cpuguy83: Added.

cpuguy83:
Ugh, messed up my rebase.

cpuguy83:
All cleaned up.

adamperlin (approved):

adamperlin:
Looks good to me!

PR #307, opened 2024-07-02
Make systemd unit enable optional

cpuguy83:
Deb based distros enable systemd units by default as opposed to mariner (and other rpm distros) which do not enable by default.

By making the systemd unit's enable field optional we can choose to use the default for a distro when the enable field is unset.

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -397,3 +397,24 @@ func SortedMapValues[T any](m map[string]T) []T {
 
 	return out
 }
+
+// OptionTrue returns a pointer to a bool with the value true.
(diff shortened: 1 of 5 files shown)


helpers.go
```diff
@@ -397,3 +397,24 @@ func SortedMapValues[T any](m map[string]T) []T {
 
 	return out
 }
+
```
adamperlin, line 400: At some point we might want to add something like 
```go
func OptionUnset() *bool {
   return nil
}
```

adamperlin (approved):
Looks good to me!

Issue #242, opened 2024-05-09
[REQ] Support Specifying Artifacts for Most Common RPM Path Macros


adamperlin:
### What kind of request is this?

None

### What is your request or suggestion?

The reference [here](https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/#macros_installation) shows the most commonly used RPM path macros, along with their defaults. Some of them we can map to already using the `binaries` and `manpages` sections of `artifacts`. It seems like we'd want to support most of the other ones listed on the page for completeness.
- [ ] `%{_sysconfdir}` (#240)
- [ ] `%{_includedir}`
- [ ] `%{_bindir}`
- [ ] `%{_libdir}` 
- [ ] `%{_libexecdir}` 
- [ ] `%{_datadir}` 
- [ ] `{_infodir}`
- [ ] `%{_mandir}`
- [ ] `%{_docdir}`
- [ ] `%{_localstatedir}`
- [ ] `%{_sharedstatedir}` (#240)

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

tuananh:
@adamperlin regarding `_infodir`, how does one detect whether its' GNU info file? file extension alone enough?

cpuguy83:
@tuananh We don't do any detection.
Its up to the spec writer to deal with that.

tuananh:
@cpuguy83 for _infodir, it's usually used to bundle files into -doc subpackage right?

should we just check for `.info` & `.info.tgz` and generate _infodir marco accordingly like this

```go
docKeys := dalec.SortMapKeys(w.Spec.Artifacts.Docs)
for _, d := range docKeys {
	cfg := w.Spec.Artifacts.Docs[d]

	var root string
	if strings.Contains(w.Name , ".info") || strings.Contains(w.Name, ".info.tgz") {
		root := filepath.Join(`%{buildroot}/%{_infodir}`, w.Name)
	} else {
		root := filepath.Join(`%{buildroot}/%{_docdir}`, w.Name)
	}
	
	copyArtifact(root, d, &cfg)
}
```

PR #308, opened 2024-07-03
feat: add support for datadir in rpm spec

pmengelbert:

Users have requested the ability to put read-only architecture-independent data files under `/usr/share`.

Related #242 (does not completely resolve)

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -129,7 +129,7 @@ type Artifacts struct {
 	// Manpages is the list of manpages to include in the package.
 	Manpages map[string]ArtifactConfig `yaml:"manpages,omitempty" json:"manpages,omitempty"`
 	// DataDirs is a list of read-only architecture-independent data files, to be placed in /usr/share/
-	DataDirs map[string]ArtifactConfig `yaml:"data_files,omitempty" json:"data_files,omitempty"`
+	DataDirs map[string]ArtifactConfig `yaml:"data_dirs,omitempty" json:"data_dirs,omitempty"`
 	// Directories is a list of various directories that should be created by the package.
 	Directories *CreateArtifactDirectories `yaml:"createDirectories,omitempty" json:"createDirectories,omitempty"`
 	// ConfigFiles is a list of files that should be marked as config files in the package.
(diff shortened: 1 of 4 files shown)


test/azlinux_test.go
```diff
@@ -760,6 +760,80 @@ Environment="KUBELET_KUBECONFIG_ARGS=--bootstrap-kubeconfig=/etc/kubernetes/boot
...
+					},
+				},
+			},
+			Build: dalec.ArtifactBuild{},
+			Artifacts: dalec.Artifacts{
+				Binaries: map[string]dalec.ArtifactConfig{
+					"bin": {},
+				},
+				DataFiles: map[string]dalec.ArtifactConfig{
+					"data_dir":  {},
```
adamperlin, line 810: It might be good to add a test with a subpath in the artifact config

adamperlin (approved):
I think this looks fine to me. One thing to consider is, it likely would not be good practice to install directly under `%{_datadir}` as in `%{_datadir}/my_file.data`, I believe there should pretty much always be at least one directory level of nesting to specify the package the data applies to. That being said, I'm not sure here is the place to validate that necessarily.

cpuguy83:
@adamperlin Agreed I was going to comment the same thing but yeah I don't think there's really anything we can do here other than provide guidance that it should be namespaced.

Looks like you need to `go generate ./...`

pmengelbert:
> @adamperlin Agreed I was going to comment the same thing but yeah I don't think there's really anything we can do here other than provide guidance that it should be namespaced.
> 
> Looks like you need to `go generate ./...`

@cpuguy83 @adamperlin 
We could require a subpath for this artifact type and drop the parent directory if the `DataFile` in question is a Directory. Happy to change it to that if y'all think it would be better.

test/azlinux_test.go
pmengelbert: done, thanks!

cpuguy83 (changes requested):
This can also use a squash.

spec.go
```diff
@@ -128,6 +128,8 @@ type Artifacts struct {
 	Binaries map[string]ArtifactConfig `yaml:"binaries,omitempty" json:"binaries,omitempty"`
 	// Manpages is the list of manpages to include in the package.
 	Manpages map[string]ArtifactConfig `yaml:"manpages,omitempty" json:"manpages,omitempty"`
+	// DataDirs is a list of read-only architecture-independent data files, to be placed in /usr/share/
+	DataDirs map[string]ArtifactConfig `yaml:"data_files,omitempty" json:"data_files,omitempty"`
```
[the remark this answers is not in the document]
pmengelbert: done

cpuguy83 (approved):

PR #310, opened 2024-07-03
Fix RPM Systemd Artifact Support

### LINKED ISSUE #301: [BUG] DALEC should not be using systemd preset files
### Expected Behavior

Reading up on systemd presets and it seems like the recommendation is to not ship preset files in packages.
The intention of these is for the distribution to have a central location of presets.

ref: https://www.freedesktop.org/software/systemd/man/latest/systemd.preset.html#Description

### Actual Behavior

This should probably just use a post install script.

### Steps To Reproduce

N/A

adamperlin:
Fixes #301. It is not good practice to include `.preset` files in an RPM package so this PR changes the `%post` section of dalec-generated RPM Specs to enable/disable systemd units manually. 

Additionally, I discovered in the course of changing this behavior that systemd was **not** actually marked as a `post`, `preun` and `postun` dependency, meaning that prost, preun, and postun scripts using the `systemd_*` macros were actually broken to begin with. This PR fixes that issue as well.

Fixes #


frontend/rpm/template.go
```diff
@@ -317,6 +325,22 @@ func (w *specWrapper) PreUn() fmt.Stringer {
 	return b
 }
 
+func systemdPostScript(unitName string, cfg dalec.SystemdUnitConfig) string {
+	verb := "disable"
```
cpuguy83, line 329: This should take into account the default value and do nothing.

cpuguy83: Actually thinking about it we should probably only ever enable and do nothing when false.
I wrote up that other PR to make it a `*bool` with the idea that we have these preset files but I'm thinking maybe we don't need that?

basically

```go

if !cfg.Enable {
  return ""
}

// systemctl enable
```

cpuguy83 (changes requested):

frontend/rpm/template.go
adamperlin: Yeah that seems reasonable. Because I believe `systemctl disable` is essentially a no-op unless the service is already enabled, right?

cpuguy83: And also potentially something we don't want to do willy-nilly.

adamperlin: So this should be fixed. It does actually complicate things a bit though, because if we aren't explicitly disabling units, then we don't necessarily need a `%post` section or `%post` systemd dependency if no enabled units have been specified in the spec. I've added some logic to account for this

cpuguy83:
Can you add a revert of https://github.com/Azure/dalec/pull/307 since given this change I don't think we need it?

adamperlin:
> Can you add a revert of #307 since given this change I don't think we need it?

This is done!

cpuguy83 (approved):

PR #318, opened 2024-07-09
Fix patching from dirs redux

### LINKED ISSUE #185: [REQ] Allow patches to be a directory of patch files

cpuguy83:
The original patch (https://github.com/Azure/dalec/commit/19f035d20a648a01e110f29516c387d2029673fd) did not work and is reverted in the first commit.

We did not know the original commit was broken because the test case for it was not running correctly, likely due to a bad rebase somewhere down the line. This is fixed in the 2nd commit.

The third commit simplifies the whole process a bit by adding a `Path` field to the patch spec which allows the patch spec itself specify a file.

Closes #185


load.go
```diff
@@ -405,6 +405,31 @@ func (s Spec) Validate() error {
 		}
 	}
 
+	var patchErr error
+	for src, patches := range s.Patches {
+		for _, patch := range patches {
+			if isRoot(patch.Path) {
```
DannyBrito, line 411: I think we should drop this if conditional to only check the patch if it `isRoot` and check for all of the patches.
As possible wrong use case could happen something like you specify a http patch source but then you also added the
`Path` in the patches block
```yml
sources:
  src:
    git:
      url: https://github.com/kubernetes/kubernetes.git
      commit: "fake"
      keepGitDir: true
  patch_1:
    http:
      url: https://.../fake.patch
patches:
 src:
 - source: patch_1
    path: fake/path
```
and then:
1. we would check that this source exists 
2.  if the source is a dir. handling `http` and `context` sources.
2.1 if the source is not a dir, then we don't expect the path field 
2.2 if it is a dir, we check that path is not root

frontend/rpm/template.go
```diff
@@ -274,37 +268,28 @@ func (w *specWrapper) PrepareSources() (fmt.Stringer, error) {
...
+		// This is a directory source so it needs to be untarred into the rpm build dir.
+		fmt.Fprintf(b, "mkdir -p \"%%{_builddir}/%s\"\n", key)
+		fmt.Fprintf(b, "tar -C \"%%{_builddir}/%s\" -xzf \"%%{_sourcedir}/%s.tar.gz\"\n", key, key)
+	}
+	prepareGomods()
+
+	// Apply patches to all suorces
+	// Note: These are applied based ont he key sorting algorithm (lexicographic).
+	//  Using one patch to patch another patch is not supported, except that it may
+	//  occur if they happen to be sorted lexicographically.
```
DannyBrito, line 288:
```suggestion
	// Apply patches to all sources
	// Note: These are applied based on the key sorting algorithm (lexicographic).
	// Using one patch to patch another patch is not supported, except that it may
	// occur if they happen to be sorted lexicographically.
```

load.go
adamperlin: Just to add on to what Danny is saying, I'm fairly certain the behavior here is not correct as is: validation of a patch source shouldn't be conditional on whether it has a root sub-path, it should always be done

adamperlin: Right now, we could patch a source with a non-existent one in the spec and there would be no error if the user also supplied a non-root subpath

cpuguy83: Yes, I think this was a mistake and I extended this if block around the whole thing.

frontend/rpm/template.go
cpuguy83: Fixed.

frontend/rpm/template.go
```diff
@@ -274,37 +268,28 @@ func (w *specWrapper) PrepareSources() (fmt.Stringer, error) {
...
+		// This is a directory source so it needs to be untarred into the rpm build dir.
+		fmt.Fprintf(b, "mkdir -p \"%%{_builddir}/%s\"\n", key)
+		fmt.Fprintf(b, "tar -C \"%%{_builddir}/%s\" -xzf \"%%{_sourcedir}/%s.tar.gz\"\n", key, key)
+	}
+	prepareGomods()
+
+	// Apply patches to all sources.
+	// Note: These are applied based on the key sorting algorithm (lexicographic).
+	//  Using one patch to patch another patch is not supported, except that it may
+	//  occur if they happen to be sorted lexicographically.
```
adamperlin, line 288: The possibility of patching another patch if they *happen* to be sorted lexicographically is a weird one which I think we should still disallow? Maybe not needed for this PR though

cpuguy83: Yeah, I considered explicitly disallowing it, but kudos to someone who needs to and finds a way to make it happen, I guess.
I think it would be super-duper weird for someone to actually even attempt to apply a patch to a patch file.

adamperlin (approved):

DannyBrito (commented):
could we add an example of the spec might look using both type of accepted patch sources, when it is a dir and when it is a file? In case for the http and context sources and how when it is a dir the path is required and exclusive to a dir to the docs?
Otherwise lgtm

cpuguy83:
@DannyBrito Thanks for the reminder, added docs!

website/docs/sources.md
```diff
@@ -320,10 +320,67 @@ sources:
...
+    - source: md2man-patch # The name of the source that contains the patch
+      path: "" # Path inside the patch source where the patch file is located.
+    # Add more patches to the list (After adding them to the sources section) if needed
+```
+
+Each patch in the list of patch sources MUST be pointing to a file. When the
+patch source is a file-based source, such as `http`, the `path` parameter in the
+patch spec must not be set. When the patch is a directory-based source, such as
+`git`, the `path` parameter in the patch spec must be set AND referecing a file
+in the source.
```
DannyBrito, line 356: Is this correct?
```
When the patch is a directory-based source, such as
`git`, the `path` parameter in the patch spec must be set AND referecing a file
in the source.
```
can you use a git source as the patch? or would that be a `context` source? if both are correct could you also mention `context` as the example below is using a context source?

website/docs/sources.md
```diff
@@ -320,10 +320,67 @@ sources:
...
+      # Each entry is a patch spec and points to a source listed in the `sources` section
+    - source: md2man-patch # The name of the source that contains the patch
+      path: "" # Path inside the patch source where the patch file is located.
+    # Add more patches to the list (After adding them to the sources section) if needed
+```
+
+Each patch in the list of patch sources MUST be pointing to a file. When the
+patch source is a file-based source, such as `http`, the `path` parameter in the
+patch spec must not be set. When the patch is a directory-based source, such as
+`git`, the `path` parameter in the patch spec must be set AND referecing a file
```
DannyBrito, line 355:
```suggestion
`git`, the `path` parameter in the patch spec must be set AND referencing a file
```

website/docs/sources.md
```diff
@@ -327,10 +320,67 @@ sources:
...
+sources:
+  md2man:
+    git:
+      url: https://github.com/cpuguy83/go-md2man.git
+      commit: v2.0.3
+  localPatches:
+    context: {}
+
+
+patches:
```
DannyBrito, line 371: If I recall correctly, not setting the path in the context source will use the `.` context right? This is technically but wouldn't this bloat this source with all the other context in `.`, maybe we could be more specific?
```suggestion
    context: {}
    path: patches

patches:
```

cpuguy83: Yes, but this isn't a doc about context.
I decided to keep it simple here (note my original version of this had an `includes` block).
I think using `path` in the source here makes what we are trying to demonstrate more confusing.

website/docs/sources.md
cpuguy83: Every directory-based source is supported (context, build, git, image, inline-dir).
Note that all the sources in this doc mention if they are directory or file based.

website/docs/sources.md
cpuguy83: I added a few more examples, let me know what you think!

DannyBrito (approved):
lgtm!

Issue #282, opened 2024-06-06
[REQ] add option to skip signing


DannyBrito:
### What kind of request is this?

Improvement of existing experience

### What is your request or suggestion?

When we specify the signing block for a target in a `dalec` spec, it defaults to calling the signer. This can be inconvenient when attempting to test builds in a local environment, especially when we are not aiming to test signing.

It would be beneficial to have a `build-arg` or an equivalent flag/option to bypass the signing call.

Maybe, we might want to consider moving the signing block to a `build-arg` or flag. From my perspective, having the dalec spec and the signing block together makes it harder to share the specs directly. We would need to inform the recipient that they need to adjust the signing block to use their own signer image or to remove it. Alternatively, we would need to modify the spec and delete the signing block before sharing it.

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

cpuguy83:
I think it would be fine to support a build-arg like `DALEC_SKIP_SIGNING=1`

cpuguy83:
I could also see supporting reading the signing spec from a named context.

pmengelbert:
- [ ] https://github.com/Azure/dalec/pull/320#discussion_r1674671889 Throw warning when signing is disabled

PR #320, opened 2024-07-11
feat: Add known build arg to disable signing

pmengelbert:
Allows the user to skip signing by adding `--build-arg DALEC_SKIP_SIGNING=1` to their `docker build` invocation. This is a partial resolution of #282. I will submit a separate PR that implements the remainder of #282.

### DIFF
--- a/frontend/azlinux/handle_rpm.go
+++ b/frontend/azlinux/handle_rpm.go
@@ -73,13 +73,19 @@ func specToRpmLLB(ctx context.Context, w worker, client gwclient.Client, spec *d
 	specPath := filepath.Join("SPECS", spec.Name, spec.Name+".spec")
 	st := rpm.Build(br, base, specPath, opts...)
 
-	if signer, ok := spec.GetSigner(targetKey); ok {
-		signed, err := frontend.ForwardToSigner(ctx, client, signer, st)
-		if err != nil {
-			return llb.Scratch(), err
-		}
-		st = signed
+	if frontend.SigningDisabled(client) {
+		return st, nil
 	}
 
-	return st, nil
+	signer := spec.GetSigner(targetKey)
+	if signer == nil {
+		return st, nil
+	}
+
+	signed, err := frontend.ForwardToSigner(ctx, client, signer, st)
+	if err != nil {
+		return llb.Scratch(), err
+	}
+
+	return signed, nil
 }
--- a/frontend/azlinux/handle_rpm.go
+++ b/frontend/azlinux/handle_rpm.go
@@ -73,19 +73,5 @@ func specToRpmLLB(ctx context.Context, w worker, client gwclient.Client, spec *d
 	specPath := filepath.Join("SPECS", spec.Name, spec.Name+".spec")
 	st := rpm.Build(br, base, specPath, opts...)
 
-	if frontend.SigningDisabled(client) {
-		return st, nil
-	}
-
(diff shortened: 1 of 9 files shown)


helpers.go
```diff
@@ -369,6 +370,12 @@ func InstallPostSymlinks(post *PostInstall, rootfsPath string) llb.RunOption {
 }
 
 func (s *Spec) GetSigner(targetKey string) (*PackageSigner, bool) {
+	if skipVal, ok := s.Args["DALEC_SKIP_SIGNING"]; ok {
```
cpuguy83, line 373: This is a field in the spec, I don't think we are looking at people adding this to the spec itself, just a built-in arg that people can specify at build time that we'll look for.

See https://github.com/Azure/dalec/blob/577b9ef6845074e178268dc25cd19c4be7334ffa/load.go#L16-L27 for adding this in to the list of known args so the build doesn't error out.

We'll need to check the for the option in the build args on the client.

helpers.go
```diff
@@ -369,6 +370,12 @@ func InstallPostSymlinks(post *PostInstall, rootfsPath string) llb.RunOption {
...
+		if shouldSkip, err := strconv.ParseBool(skipVal); err == nil && shouldSkip {
+			return nil, false
```
cpuguy83, line 375: I noticed these bools are only true when we pass a non-nil signer back and false for every other case. Can we drop the bool?

cpuguy83 (changes requested):

helpers.go
pmengelbert: Good catch. It's already been added to the list of known args

helpers.go
pmengelbert: Sure

helpers.go
pmengelbert: Fixed this. It's checking the client now.

frontend/azlinux/handle_rpm.go
```diff
@@ -73,13 +73,19 @@ func specToRpmLLB(ctx context.Context, w worker, client gwclient.Client, spec *d
...
+		return st, nil
+	}
+
+	signed, err := frontend.ForwardToSigner(ctx, client, signer, st)
+	if err != nil {
+		return llb.Scratch(), err
 	}
 
-	return st, nil
+	return signed, nil
```
DannyBrito, line 90: At this point, should we consider wrapping these steps in a function? So, it can be called in another handler when needed? In this `frontend/windows/handle_zip.go` could be reused?

pmengelbert: done

frontend/request.go
```diff
@@ -134,7 +135,40 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
 	return st.Marshal(ctx)
 }
 
-func ForwardToSigner(ctx context.Context, client gwclient.Client, cfg *dalec.PackageSigner, s llb.State) (llb.State, error) {
+func MaybeSign(ctx context.Context, client gwclient.Client, st llb.State, spec *dalec.Spec, targetKey string) (llb.State, error) {
+	if signingDisabled(client) {
+		return st, nil
```
cpuguy83, line 140: As a follow-up, we should look at throwing a warning back to the client when signing is disabled.

cpuguy83 (changes requested):
LGTM

cpuguy83 (approved):

PR #322, opened 2024-07-12
Send client warning when signing is disabled

cpuguy83:
When signing is explicitly disabled this surfaces a warning message on the client.
Example:

image

### DIFF
--- a/frontend/gateway.go
+++ b/frontend/gateway.go
@@ -10,6 +10,8 @@ import (
 	"github.com/moby/buildkit/frontend/dockerui"
 	gwclient "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/solver/pb"
+	"github.com/moby/buildkit/util/bklog"
+	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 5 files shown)


test/testenv/buildx.go
```diff
@@ -261,13 +304,14 @@ func (b *BuildxEnv) RunTest(ctx context.Context, t *testing.T, f TestFunc) {
 		f(ctx, gwc)
 		return gwclient.NewResult(), nil
 	}, ch)
-	if err != nil {
-		t.Fatal(err)
```
cpuguy83, line 265: This fatal call here could sometimes make the test runner yell because a buildkit status message could hit the test log after the fatal is called.
Hence why I moved the `<-done` after.

frontend/gateway.go
```diff
@@ -213,3 +213,17 @@ type BuildOpstGetter interface {
 func GetTargetKey(client BuildOpstGetter) string {
 	return client.BuildOpts().Opts[keyTopLevelTarget]
 }
+
+// Warn sends a warning tot he client for the provided state.
```
DannyBrito, line 217:
```suggestion
// Warn sends a warning to the client for the provided state.
```

frontend/request.go
```diff
@@ -137,6 +138,10 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
 
 func MaybeSign(ctx context.Context, client gwclient.Client, st llb.State, spec *dalec.Spec, targetKey string) (llb.State, error) {
 	if signingDisabled(client) {
+		err := Warn(ctx, client, st, "Signing disabled by build-arg "+keySkipSigningArg)
+		if err != nil {
+			bklog.G(ctx).WithError(err).Error("Error writing warning log")
```
DannyBrito, line 143: We could be a bit more specific about being an error creating the warning about the signing being disabled. Might help in the future if we add more warning?

frontend/request.go
```diff
@@ -153,9 +158,11 @@ func MaybeSign(ctx context.Context, client gwclient.Client, st llb.State, spec *
 	return signed, nil
 }
 
+const keySkipSigningArg = "DALEC_SKIP_SIGNING"
+
```
DannyBrito, line 162: nit: move this `const` to the top of the file?

cpuguy83: Done

frontend/request.go
cpuguy83: I'm not sure I 100% follow here, but I've updated this to move all the error handling in the `Warn` function and also made it more robust to such errors.

frontend/gateway.go
cpuguy83: Fixed

pmengelbert (approved):
LGTM

PR #323, opened 2024-07-13
Fix test platforms

cpuguy83:
This fixes 2 things:

1. Always return a non-nil image config for build artifacts regardless of if they are an image [1].
2. Tests: set platform on returned llb.State [2]

### [1] Stop returning nil image config
This can cause some issues in buildkit libs and also yields a cryptic
error on the docker CLI for these non-container targets when the build
is not outputing tot he local dir.

One specific case right now is the dockerui client which drives all the
build targets will panic on the windowscross/zip target when using
windows as a target platform.

In the past this has also caused panics in dockerd and buildkit. These
have been fixed upstream and were considered security issues due to a
frontend being able to panic dockerd, but I don't think we should
continue doing this.

Something to consider in the future is to either replace our usage of
dockerui to drive builds (not great because it is the canonical
implementation for interacting with docker) and also returning a custom
OCI artifact type for these non-container builds OR we could work
upstream to make dockerui support a more generic type.

### [2] Set platform on returned llb state
In my case I am trying to run tests from a Mac connecting to a
linux/amd64 machine over SSH.
The windows signing tests end up failing due to the platform missing
from the `llb.State` generated for the the unzipper intermediate state.
It tries to run a command in the unzip state but fails due to the
architecture not being compataible.

Before this change `reqToState` is returning an `llb.State` without a
platform set on it. Subsequent `.Run()` calls on that state will
default to the platform of the client (darwin/arm64).

This changes extracts the first (and should be only in the case of our
tests) platform from the returned result and adds that to the returned
`llb.State`.

### DIFF
--- a/frontend/azlinux/handle_rpm.go
+++ b/frontend/azlinux/handle_rpm.go
@@ -47,7 +47,7 @@ func handleRPM(w worker) gwclient.BuildFunc {
 			if err != nil {
 				return nil, nil, err
 			}
-			return ref, nil, nil
+			return ref, &dalec.DockerImageSpec{}, nil
 		})
 	}
 }
--- a/frontend/debug/handle_gomod.go
+++ b/frontend/debug/handle_gomod.go
@@ -55,6 +55,6 @@ func Gomods(ctx context.Context, client gwclient.Client) (*client.Result, error)
 		if err != nil {
 			return nil, nil, err
 		}
-		return ref, nil, nil
+		return ref, &dalec.DockerImageSpec{}, nil
 	})
 }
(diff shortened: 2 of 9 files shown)


test/helpers_test.go
```diff
@@ -266,3 +267,29 @@ func withBuildContext(ctx context.Context, t *testing.T, name string, st llb.Sta
...
+		t.Fatal(err)
+	}
+
+	st, err := ref.ToState()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	dt, ok := res.Metadata[exptypes.ExporterPlatformsKey]
+	if ok {
```
adamperlin, line 286: in what cases would there not be a platform in the metadata?

cpuguy83: Things like debug/sources wouldn't have a platform set.

cpuguy83: Oh that's not true, we are using it there... well anything that's not using `frontend.BuildWIthPlatform` would not have a platform set.
Could be some intermediate state that we marshal internally in a test, as an example.

adamperlin (approved):

PR #326, opened 2024-07-15
feat: provide signing config separately from the main spec

### LINKED ISSUE #282: [REQ] add option to skip signing

pmengelbert:
This PR allows you to specify the signing config separately from the main spec.

You can do this in one of three ways:
1. Provide a build context with the exact name `dalec_signing_config` which points to a directory containing a file with the exact name `dalec_signing_config.yml`
1. Provide a build context as above, but with a custom path for the config, by using the build arg `DALEC_SIGNING_CONFIG_PATH`. This path will be relative to the root of the supplied build context.
1. Provide a custom path for the config on your local filesystem, by using the build arg `DALEC_SIGNING_CONFIG_PATH`. This path can be absolute, or relative to the context location from the `docker build` invocation.

Fixes #282 


frontend/request.go
```diff
@@ -135,14 +141,97 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
 	return st.Marshal(ctx)
 }
 
+func signingContextPresent(bopts map[string]string) bool {
+	_, ok := bopts["context:"+dalecSigningContext]
+	return ok
+}
+
+func tryGetNamedContextSigning(ctx context.Context, client gwclient.Client, configPath string) (*dalec.PackageSigner, error) {
```
cpuguy83, line 149: nit: no need for `try` here.

frontend/request.go
```diff
@@ -135,14 +141,97 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
...
+	if err != nil {
+		return nil, err
+	}
+
+	ref, err := res.SingleRef()
+	if err != nil {
+		return nil, err
+	}
+
+	dt, err := ref.ReadFile(ctx, gwclient.ReadRequest{
```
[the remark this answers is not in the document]
cpuguy83: To simplify this down further, maybe there is _no_ implicit context name we will look for.
Instead the build-arg for the filename must always be provided, and if the client wants to use a custom context they need to tell us that in another build arg (e.g. `DALEC_SIGNING_CONTEXT_NAME`)

frontend/request.go
```diff
@@ -168,6 +267,16 @@ func signingDisabled(client gwclient.Client) bool {
 	return isDisabled
 }
 
+func getUserSignConfigPath(client gwclient.Client) string {
+	bopts := client.BuildOpts().Opts
+	v, ok := bopts["build-arg:DALEC_SIGNING_CONFIG_PATH"]
```
cpuguy83, line 272: Can you make the base-key a const?

cpuguy83 (changes requested):

frontend/request.go
```diff
@@ -135,14 +141,97 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
...
 
+func signingContextPresent(bopts map[string]string) bool {
+	_, ok := bopts["context:"+dalecSigningContext]
+	return ok
+}
+
+func tryGetNamedContextSigning(ctx context.Context, client gwclient.Client, configPath string) (*dalec.PackageSigner, error) {
+	bopts := client.BuildOpts().Opts
+
+	dc, err := dockerui.NewClient(client)
```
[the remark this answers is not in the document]
pmengelbert: > This is fairly expensive to create.

Surprising -- why is that?

frontend/request.go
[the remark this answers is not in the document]
pmengelbert: done

frontend/request.go
```diff
@@ -135,14 +141,97 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
...
+	if customSignCtxPresent {
+		sc, _, err = dc.NamedContext(ctx, dalecSigningContext, dockerui.ContextOpt{})
+	} else {
+		sc, err = dc.MainContext(ctx)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	// When the main context is used, we don't want to solve with the full
```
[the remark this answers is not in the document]
pmengelbert: done

frontend/request.go
pmengelbert: done

frontend/request.go
pmengelbert: Yes

pmengelbert: done

frontend/request.go
```diff
@@ -135,22 +140,72 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
...
+	switch cfgPath := getUserSignConfigPath(client); cfgPath {
+	case "": // no custom path provided, check the spec
+		cfg := spec.GetSigner(targetKey)
+		if cfg == nil {
+			// i.e. there's no signing config. not in the build context, not in the spec.
+			return st, nil
+		}
+
+		return forwardToSigner(ctx, client, cfg, st)
+	default: // no custom path provided, check the spec
```
cpuguy83, line 201: Comment looks like bad copy/psate since this is checking the context.

test/fixtures/signer/sign_config.yml
```diff
@@ -0,0 +1,3 @@
+signer:
```
cpuguy83, line 1: Doesn't look like we need this file?

test/signing_test.go
```diff
@@ -8,15 +8,420 @@ import (
...
 )
 
-func distroSigningTest(t *testing.T, spec *dalec.Spec, buildTarget string) testenv.TestFunc {
+func runTest(t *testing.T, f testenv.TestFunc) {
+	t.Helper()
+	ctx := startTestSpan(baseCtx, t)
+	testEnv.RunTest(ctx, t, f)
+}
+
+func newSpec() *dalec.Spec {
```
cpuguy83, line 22: This is a fairly generic function name that's only minimally useful.
`newSimpleSpec`?

test/signing_test.go
```diff
@@ -8,15 +8,420 @@ import (
...
+			},
+		},
+	})
+
+	return spec
+}
+
+func linuxSigningTests(ctx context.Context, testConfig testLinuxConfig) func(*testing.T) {
+	return func(t *testing.T) {
+		newSpec := func() *dalec.Spec {
```
cpuguy83, line 52: This mostly looks like a dup of the actual `newSpec`?

test/signing_test.go
```diff
@@ -8,15 +8,420 @@ import (
...
+				withBuildArg("DALEC_SIGNING_CONFIG_PATH", "/unusual_place.yml"),
+			))
+		})
+
+		t.Run("with no build context and config path build arg", func(t *testing.T) {
+			spec := newSpec()
+			spec.PackageConfig.Signer = nil
+
+			signConfig := llb.Scratch().
+				File(llb.Mkdir("/test/fixtures/signer/", 0o755, llb.WithParents(true))).
```
cpuguy83, line 210: Can we trim these paths down since they aren't needed?

frontend/request.go
```diff
@@ -135,22 +140,72 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
...
-	return signed, nil
+	return pc.Signer, nil
+}
+
+func MaybeSign(ctx context.Context, client gwclient.Client, st llb.State, spec *dalec.Spec, targetKey string, sOpt dalec.SourceOpts) (llb.State, error) {
+	if signingDisabled(client) {
+		return st, nil
+	}
+
+	switch cfgPath := getUserSignConfigPath(client); cfgPath {
```
cpuguy83, line 192: After #322 is merged, we should add a warning when the build-args are overwriting the signing in the spec.

cpuguy83 (changes requested):

frontend/request.go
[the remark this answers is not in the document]
cpuguy83: > Surprising -- why is that?

It is eagerly initializing a bunch of stuff on `NewClient`

frontend/request.go
pmengelbert: sgtm, will wait till that one's merged

frontend/request.go
pmengelbert: done

test/fixtures/signer/sign_config.yml
pmengelbert: done

test/signing_test.go
pmengelbert: done

test/signing_test.go
pmengelbert: It set up a signing config. I refactored it to deduplicate.

test/signing_test.go
pmengelbert: done

test/signing_test.go
pmengelbert: done

frontend/request.go
cpuguy83: That one is in now. Let's also make sure to bring back the tests that check to make sure the warning is present 😅 🚀

pmengelbert: Merged that, updated this

pmengelbert: done

cpuguy83: Looks like tests are missing the checks for the warning messages (see: `WithSolveStatusFn`)

pmengelbert: I need to slow down!

pmengelbert: done

test/signing_test.go
```diff
@@ -8,15 +8,436 @@ import (
...
+			spec := newSigningSpec()
+			spec.PackageConfig.Signer.Frontend.Image = "notexist"
+
+			signConfig := llb.Scratch().File(llb.Mkfile("/unusual_place.yml", 0o400, []byte(`
+signer:
+  image: `+phonySignerRef+`
+  cmdline: /signer
+`)))
+
+			var found bool
```
cpuguy83, line 150: The test is never checking this value.

test/signing_test.go
```diff
@@ -8,15 +8,436 @@ import (
...
+						Signer: spec.PackageConfig.Signer,
+					},
+				},
+			}
+			spec.PackageConfig.Signer = nil
+
+			runTest(t, distroSigningTest(t, spec, testConfig.SignTarget))
+		})
+
+		t.Run("target config takes precedence when root config is there", func(t *testing.T) {
```
cpuguy83, line 88: This could use a similar check to see if the warning is provided.

cpuguy83 (changes requested):

test/signing_test.go
pmengelbert: thanks

test/signing_test.go
pmengelbert: done -- had to implement the warning to get the test to pass :)

test/signing_test.go
pmengelbert: done

frontend/request.go
```diff
@@ -137,23 +142,86 @@ func marshalDockerfile(ctx context.Context, dt []byte, opts ...llb.ConstraintsOp
...
+}
+
+func MaybeSign(ctx context.Context, client gwclient.Client, st llb.State, spec *dalec.Spec, targetKey string, sOpt dalec.SourceOpts) (llb.State, error) {
+	if signingDisabled(client) {
+		Warnf(ctx, client, st, "Signing disabled by build-arg %q", keySkipSigningArg)
+		return st, nil
+	}
+
+	cfg, warning := spec.GetSigner(targetKey)
+	switch cfgPath := getUserSignConfigPath(client); cfgPath {
```
cpuguy83, line 191: Can we restructure this like to reduce indentation?

```go
cfgPath := getUserSignConfigPath(client)
if cfgPath != "" {
  // handle
  return updatedSt, nil
}

cfg, warning := spec.GetSigner(targetKey)
// handle the rest
```

helpers.go
```diff
@@ -368,18 +375,23 @@ func InstallPostSymlinks(post *PostInstall, rootfsPath string) llb.RunOption {
 	})
 }
 
-func (s *Spec) GetSigner(targetKey string) *PackageSigner {
+func (s *Spec) GetSigner(targetKey string) (*PackageSigner, Warning) {
```
cpuguy83, line 378: Instead this could return a bool to help simplify understanding what's going on here.
Also would need a doc string to explain what the 2nd return is for.
Since `Warning` isn't really an error this feels a bit strange.

test/signing_test.go
```diff
@@ -8,15 +8,453 @@ import (
...
+			spec.PackageConfig.Signer.Frontend.Image = "notexist"
+
+			signConfig := llb.Scratch().
+				File(llb.Mkfile("/sign_config.yml", 0o400, []byte(`
+signer:
+  image: `+phonySignerRef+`
+  cmdline: /signer
+`)))
+
+			runTest(t, distroSigningTest(
```
cpuguy83, line 244: Also needs a warning check.

test/signing_test.go
```diff
@@ -8,15 +8,453 @@ import (
...
+
+			runTest(t, distroSkipSigningTest(
+				t,
+				spec,
+				testConfig.SignTarget,
+				withBuildArg("DALEC_SIGNING_CONFIG_CONTEXT_NAME", "dalec_signing_config"),
+				withBuildArg("DALEC_SIGNING_CONFIG_PATH", "/sign_config.yml"),
+				withBuildContext(ctx, t, "dalec_signing_config", signConfig),
+			))
+			runTest(t, distroSkipSigningTest(t, spec, testConfig.SignTarget))
```
cpuguy83, line 294: Needs warning check.

test/signing_test.go
```diff
@@ -8,15 +8,453 @@ import (
...
+`)))
+
+			runTest(t, distroSkipSigningTest(
+				t,
+				spec,
+				testConfig.SignTarget,
+				withMainContext(ctx, t, signConfig),
+				withBuildArg("DALEC_SIGNING_CONFIG_PATH", "/sign_config.yml"),
+			))
+			runTest(t, distroSkipSigningTest(t, spec, testConfig.SignTarget))
```
cpuguy83, line 317: Needs warning check.

cpuguy83 (changes requested):

helpers.go
pmengelbert: We changed this to remove the bool which is why I went with this. But I'll put the bool back

cpuguy83: The old case was returning `true` if `PackageSIgner` was not nil.
I think what we want is to return a bool when the package signer in the target is superseding the base?

frontend/request.go
pmengelbert: done

test/signing_test.go
pmengelbert: done

test/signing_test.go
pmengelbert: done

test/signing_test.go
pmengelbert: done

cpuguy83 (approved):

PR #327, opened 2024-07-18
azlinux/mariner: Use provided platform for outputs

cpuguy83:
Before this change, the mariner/azl targets were ignoring the client provided platform for the produced aritifacts.

With this change the platform is set on images so it will rely on binfmt_misc to execute those images correctly.
This also includes some optimizations:

1. If there is a build dependency on golang it will mount the native golang toolchain over the platform-specific one and adds the proper envs to produce the correct binaries for the target platform
2. go modules fetched with the generator will use the native go
3. The rpm build root is generated using the native platform since there is no platform specific code included there
4. Native dnf is used to install packages into the rootfs with the target arch (this required a change to dnf as tdnf appears to be bugged in this case).

This does *not* add support for cross compilation (run x86 code to generate, e.g., arm64 code) outside of the specific optimizations listed above.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,12 @@ on:
       - go.mod
       - go.sum
 
+env:
+  # Used in tests to determine if certain tests should be skipped.
+  # Setting this ensures that they are *not* skipped and instead make sure CI
+  # is setup to be able to properly run all tests.
+  DALEC_CI: "1"
+
 permissions:
   contents: read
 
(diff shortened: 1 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -93,6 +99,8 @@ jobs:
         uses: docker/setup-buildx-action@988b5a0280414f521da01fcc63a27aeeb4b104db # v3.6.1
       - name: download deps
         run: go mod download
+      - name: Setup QEMU
+        run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
```
sozercan, line 103: would this run in arm64 agents? do we want to use tonis' image?

cpuguy83: This is specifically for dalec CI which only runs on amd64, and only to allow a specific test to function.

cpuguy83:
This is very outdated and long since fixed.

PR #333, opened 2024-07-22
fix example build dalec container command

DannyBrito:

Example command is missing context,
and point to spec `docs/examples/go-md2man.yml`

### DIFF
--- a/website/docs/build-source.md
+++ b/website/docs/build-source.md
@@ -179,7 +179,7 @@ This will create `RPM` and `SRPM` directories in the `_output` directory with th
 To build a container, we can use the following command:
 
 ```shell
-docker build -t go-md2man:2.0.3 -f examples/go-md2man.yml --target=mariner2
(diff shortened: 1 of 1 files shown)


website/docs/build-source.md
```diff
@@ -179,7 +179,7 @@ This will create `RPM` and `SRPM` directories in the `_output` directory with th
 To build a container, we can use the following command:
 
 ```shell
-docker build -t go-md2man:2.0.3 -f examples/go-md2man.yml --target=mariner2
+docker build -t go-md2man:2.0.3 -f docs/examples/go-md2man.yml --target=mariner2 .
```
paulgmiller, line 182: should we change the example targets to be azlinux3 as I think that has superseeded mariner?

cpuguy83 (approved):

website/docs/build-source.md
cpuguy83: azlinux3 isn't fully out yet, sadly.

PR #337, opened 2024-07-29
Properly Version Build Dependencies for Azlinux

### LINKED ISSUE #309: [BUG] Constraints for build dependencies are ignored in the build env
### Expected Behavior

When a dalec spec contains version requirements for a dependency 👍 

```yaml
dependencies:
  build:
    foo: [">= 1.0", "<2.0"]
```

The mariner2 and azlinux3 should respect those dependencies when setting up the build environment.

### Actual Behavior

The rpm spec is correct in that includes those dependency constraints, however the actual build environment does not.
I *believe* rpmbuild will error out if the versions are not within those constraints, but currently we just `tdnf install <dep1> <dep2> ...` which can land us in a situation where the build cannot succeed with version constraints in it.

### Steps To Reproduce

N/A

adamperlin:

Fixes #309 for **azlinux** targets. Currently, if version constraints are specified for build dependencies in the Dalec spec the dependencies are completely ignored when actually installing the packages with `tdnf`. However, the constraints are still emitted as `BuildRequires` in the generated RPM spec for the main package. This means that currently, we could run into cases where the versions of build dependencies installed by Dalec in the worker container would be incompatible with a user's constraints, causing rpmbuild to fail with no way for a user to acceptably resolve the issue without removing their build version constraints. 

**How it works**
A build dependency for a Dalec package is essentially just a runtime package dependency for the worker container. Thus, we can simply create a meta-package (for azlinux, azlinux-build-dependencies), which has as its *runtime dependencies* the *build dependencies* specified in the dalec spec. We then install our build dependency package in the worker container and let the package manager handle the version constraints for us. 

This bug has **not** yet been fixed for windows as those targets use a jammy worker container and thus will likely require building a `.deb` for build dependencies, functionality for which is pending with jammy support.


test/azlinux_test.go
```diff
@@ -1120,6 +1120,38 @@ Environment="KUBELET_KUBECONFIG_ARGS=--bootstrap-kubeconfig=/etc/kubernetes/boot
...
 	})
+
+	t.Run("pinned build dependencies", func(t *testing.T) {
+		t.Parallel()
+		ctx := startTestSpan(baseCtx, t)
+		testPinnedBuildDeps(ctx, t, testConfig.Target, testConfig.Worker)
+	})
+}
+
+func getTestPackageSpec(name, version string) *dalec.Spec {
```
cpuguy83, line 1131: Can you move this inline with `testPinnedBuildDeps`?

frontend/azlinux/azlinux3.go
```diff
@@ -60,6 +60,10 @@ func (w azlinux3) Install(pkgs []string, opts ...installOpt) llb.RunOption {
 	return dalec.WithRunOptions(tdnfInstall(&cfg, "3.0", pkgs), w.tdnfCacheMount(cfg.root))
 }
 
+func (w azlinux3) InstallWithReqs(deps map[string]dalec.PackageConstraints, opts ...installOpt) installFunc {
```
cpuguy83, line 63: This extra function on the interface doesn't seem since the implementations are identical.
To note: I think passing the hardcoded target name below is not correct since that should all be threaded through the request.

frontend/azlinux/common.go
```diff
@@ -0,0 +1,58 @@
...
+		pg := dalec.ProgressGroup("Building container for build dependencies")
+
+		// create an RPM with just the build dependencies, using our same base worker
+		rpmDir, err := specToRpmLLB(ctx, w, client, &depsOnly, sOpt, target, pg)
+		if err != nil {
+			return nil, err
+		}
+
+		// read the built RPMs (there should be a single one)
+		rpms, err := readRPMs(ctx, client, rpmDir)
```
cpuguy83, line 35: This is failry expensive since it un-lazies the build.
Do we need to read the rpm list vs installing all returned RPM's?

cpuguy83 (changes requested):

frontend/azlinux/common.go
adamperlin: This is a good point. I think you're right, in this case we can actually just install all returned RPMs

frontend/azlinux/azlinux3.go
cpuguy83: *doesn't seem necessary

frontend/azlinux/handle_rpm.go
```diff
@@ -52,25 +52,80 @@ func handleRPM(w worker) gwclient.BuildFunc {
...
-func installBuildDeps(w worker, spec *dalec.Spec, targetKey string, opts ...llb.ConstraintsOpt) llb.StateOption {
-	return func(in llb.State) llb.State {
-		deps := spec.GetBuildDeps(targetKey)
-		if len(deps) == 0 {
-			return in
+// Creates and installs an rpm meta-package that requires the passed in deps as runtime-dependencies
+func installBuildDepsPackage(target string, w worker, deps map[string]dalec.PackageConstraints, installOpts ...installOpt) installFunc {
+	// depsOnly is a simple dalec spec that only includes build dependencies and their constraints
+	depsOnly := dalec.Spec{
+		Name:        "azlinux-build-dependencies",
```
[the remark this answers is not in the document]
adamperlin: You mean like `<original package name>/azlinux-build-dependencies`?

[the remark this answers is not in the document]
adamperlin: Actually, '/' will cause a path issue, needs to be `<original package name>-azlinux-build-dependencies`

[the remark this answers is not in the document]
cpuguy83: Correct, with the hyphen.
Could even remove `azlinux`, I think.

cpuguy83 (approved):

cpuguy83:
Github says there is a conflict.

PR #345, opened 2024-08-06
Add subrequest for resolving spec

cpuguy83:
This is effectively a replacement for the `debug/resolve` target. The benefit here is it can be used in a pipeline, e.g. `docker buildx build --print=dalec.resolve | yq .` and does not require writing a file to disk before using it.

This will help with automation.

demo


test/handlers_test.go
```diff
@@ -121,3 +123,34 @@ func TestHandlerTargetForwarding(t *testing.T) {
 		})
 	})
 }
+
+func TestHandlerSubrequestResolve(t *testing.T) {
+	t.Parallel()
+
+	runTest(t, func(ctx context.Context, gwc gwclient.Client) {
```
adamperlin, line 130: Is it possible to add a test with multiple platforms?

cpuguy83: Done, but need #353  to make tests pass

adamperlin (approved):
This lgtm. Thanks for adding the extra test!

PR #351, opened 2024-08-21
docs: Fix edit URL

mauriciovasquezbernal:
There is an extra /docs/ causing a 404 when clicking on the "Edit this page" button, for instance the button on https://azure.github.io/dalec/build-source points to https://github.com/Azure/dalec/blob/main/website/docs/docs/build-source.md, but the correct one is https://github.com/Azure/dalec/blob/main/website/docs/build-source.md


adamperlin (approved):
This looks good to me, thank you for this fix!

cpuguy83 (approved):

PR #352, opened 2024-08-21
source: context: Optimize fetch by path

cpuguy83:
This makes sure that if a path is specified on a context source that we we only ever fetch the content from beneath that path.

Before this change we were doing the context path filtering after the content was fetched.

This also makes so that include/exclude patterns are automatically namespaced by the source path, since it doesn't make sense to include files that are outside of the source path.

### DIFF
--- a/source.go
+++ b/source.go
@@ -3,6 +3,7 @@ package dalec
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


source.go
```diff
@@ -83,8 +82,55 @@ func (src *SourceInline) AsState(name string) (llb.State, error) {
...
+		// beneath the provided path
+		if len(includes) > 0 {
+			updated := make([]string, 0, len(includes))
+			for _, p := range includes {
+				if !pathHasPrefix(p, path) {
+					p = filepath.Join(path, p)
+				}
+				updated = append(updated, p)
+			}
+			includes = updated
```
adamperlin, line 117: You could probably stick this logic into a small local function and call it for both the includes and the excludes

cpuguy83:
@adamperlin updated this with a much cleaner implementation now.
We now update exclude patterns by prepending an exclusion of all content (`*`) with an exception of the provided path (`!<path>`) and add `<path>` to the list of follow paths (ie include pattern + follow symlink).

adamperlin (approved):

PR #365, opened 2024-09-04
WIP: Spec out extra repo support

cpuguy83:

Fixes #

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -306,6 +306,34 @@ type PackageDependencies struct {
 	// running a command in the built container.
 	// See [TestSpec] for more information.
 	Test []string `yaml:"test,omitempty" json:"test,omitempty"`
+
+	// ExtraRepos is used to inject extra package repositories that may be used to
+	// satisfy package dependencies in various stages.
+	ExtraRepos []PackageRepositoryConfig `yaml:"extra_repos,omitempty" json:"extra_repos,omitempty"`
+}
+
+// PackageRepositoryConfig
+type PackageRepositoryConfig struct {
+	// Keys are the list of keys that need to be imported to use the configured
+	// repositories
+	Keys []Source `yaml:"keys,omitempty" json:"keys,omitempty"`
+	// Config list of repo configs to to add to the environment.  The format of
+	// these configs are distro specific (e.g. apt/yum configs).
+	Config []Source `yaml:"config" json:"config"`
+	// Data lists all the extra data that needs to be made available for the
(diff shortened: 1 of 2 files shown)


cpuguy83:
I think we should probably make this available from a separate file like we do with the signing spec so that we don't have to repeat the same thing in every single spec.

Also thinking of a slight change here: Instead of splitting up repo configs between build/install just add respective fields (or a string array with validated values) that includes which environments they need to go into. This may also help reduce duplication.

spec.go
```diff
@@ -306,6 +306,34 @@ type PackageDependencies struct {
...
+	Install []PackageRepositoryConfig `yaml:"install,omitempty" json:"install,omitempty"`
+}
+
+type PackageRepositoryConfig struct {
+	// Keys are the list of keys that need to be imported to use the configured
+	// repositories
+	Keys []Source `yaml:"keys,omitempty" json:"keys,omitempty"`
+	// Config list of repo configs to to add to the environment.  The format of
+	// these configs is going to be distro specific (e.g. apt/yum configs).
+	Config []Source `yaml:"config" json:"config"`
```
adamperlin, line 336: Is it likely for one repository "unit" to need multiple config files? Just trying to understand why we need an array here

cpuguy83: Mainly that I don't want to box in what can be done.
It also allows multiple repos to be installed at the same time using the same keys.

cpuguy83:
Pushed a pretty significant change to what was already there.

spec.go
```diff
@@ -306,6 +306,34 @@ type PackageDependencies struct {
...
+	Keys []Source `yaml:"keys,omitempty" json:"keys,omitempty"`
+	// Config list of repo configs to to add to the environment.  The format of
+	// these configs are distro specific (e.g. apt/yum configs).
+	Config []Source `yaml:"config" json:"config"`
+	// Data lists all the extra data that needs to be made available for the
+	// provided repository config to work.
+	// As an example, if the provided config is referencing a file backed repository
+	// then data would include the file data, assuming its not already available
+	// in the environment.
+	Data []SourceMount `yaml:"data,omitempty" json:"data,omitempty"`
```
cpuguy83, line 328: Figured we could add this so that we can get rid of the extra step of adding a repo to the worker image in the tests.

spec.go
adamperlin: Got it, that makes sense!

PR #366, opened 2024-09-07
Custom repo support for AzLinux

### LINKED ISSUE #211: [REQ] Support installing extra repos in target distro
Sometimes a package constraint cannot be resolved by a package directly from the distro.
As an example distros, especially LTS ones, rarely have new versions of a package, or sometimes just doesn't have the package at all.

Some builds will require loading custom repos.

- [x] - Add azlinux support (#366)
- [ ] - Add jammy support (#412)

adamperlin:
This PR adds custom repo support for **Azure Linux**. Jammy support soon to come. This means that packages from custom imported repos outside of the built-ins for a distro may be specified as dependencies in Dalec for the Build, Install, and Test phases.

The following spec fragment gives a good example of the feature:

```yaml
dependencies:
  build:
    custom-repo-package:

   # NEW PROPERTY of `dependencies`
  extra_repos:
    - envs:
         # this indicates which stage a custom repo will be available for. Other options are `test` and `install`
        - "build" 
      # list repo configs to import here
      config:
        custom.repo:
          # this must be a source which contains a single file, the contents of `custom.repo`
          http:
            url: https://mypackagecenter/custom.repo
        # add other repos here...

     # list GPG public keys to import here:
     keys:
       MY-RPM-GPG-PUBLIC-KEY: 
          http:
            url: https://anotherkeysource.url/MY-RPM-GPG-PUBLIC-KEY
     
     data: 
        - dest: /opt/repo
           spec:
             context:
               # this would have to be all repo data including packages and any metadata, essentially
               # a local repo created with `createrepo`
               name: "my-custom-repo-data"
       
```
Note that the `keys` portion of the spec would not be necessary if the repo config references a key directly via http in the `gpgkey` field. 

Closes #211

### DIFF
--- a/frontend/azlinux/handle_rpm.go
+++ b/frontend/azlinux/handle_rpm.go
@@ -93,18 +93,23 @@ func runTests(ctx context.Context, client gwclient.Client, w worker, spec *dalec
 	return ref, errors.Wrap(err, "TESTS FAILED")
 }
 
+var azLinuxRepoConfig = dalec.RepoPlatformConfig{
+	ConfigRoot: "/etc/yum.repos.d",
+	GPGKeyRoot: "/etc/pki/rpm-gpg",
+}
+
 func repoMountInstallOpts(repos []dalec.PackageRepositoryConfig, sOpt dalec.SourceOpts, opts ...llb.ConstraintsOpt) ([]installOpt, error) {
-	withRepos, err := withRepoConfig(repos, sOpt, opts...)
+	withRepos, err := dalec.WithRepoConfigs(repos, &azLinuxRepoConfig, sOpt, opts...)
 	if err != nil {
 		return nil, err
 	}
 
-	withData, err := withRepoData(repos, sOpt, opts...)
+	withData, err := dalec.WithRepoData(repos, sOpt, opts...)
 	if err != nil {
 		return nil, err
 	}
 
-	keyMounts, keyPaths, err := getRepoKeys(repos, sOpt, opts...)
+	keyMounts, keyPaths, err := dalec.GetRepoKeys(repos, &azLinuxRepoConfig, sOpt, opts...)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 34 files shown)


frontend/azlinux/handle_rpm.go
```diff
@@ -87,30 +87,115 @@ func installBuildDepsPackage(target string, packageName string, w worker, deps m
...
+	opts = append(opts, dalec.ProgressGroup("Install build deps"))
+	installOpt, err := installBuildDepsPackage(targetKey, spec.Name, w, deps, importRepo, installWithConstraints(opts))(ctx, client, sOpt)
 	if err != nil {
 		return nil, err
 	}
 
 	return func(in llb.State) llb.State {
-		return in.Run(installOpt, dalec.WithConstraints(opts...)).Root()
+		return in.
+			Run(importRepo, repoMounts, installOpt, dalec.ShArgs("echo 'OPT REPO:'; ls -lrt /opt/repo"), dalec.ShArgs("ls /etc/yum.repos.d/"), dalec.WithConstraints(opts...)).Root()
```
adamperlin, line 198: Some debugging here

frontend/azlinux/handle_rpm.go
```diff
@@ -87,30 +87,115 @@ func installBuildDepsPackage(target string, packageName string, w worker, deps m
...
+	}
 
-	installOpt, err := installBuildDepsPackage(targetKey, spec.Name, w, deps, installWithConstraints(opts))(ctx, client, sOpt)
+	importRepo, err := importRepos(repos, sOpt, repoOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	opts = append(opts, dalec.ProgressGroup("Install build deps"))
+	installOpt, err := installBuildDepsPackage(targetKey, spec.Name, w, deps, importRepo, installWithConstraints(opts))(ctx, client, sOpt)
```
adamperlin, line 191: the `importRepo` option which should consist of the mounts for the config files and keys is threaded through here in addition to the `Run` below. That probably isn't necessary, this was done during debugging

frontend/azlinux/handle_rpm.go
```diff
@@ -151,20 +180,117 @@ func installBuildDepsPackage(target string, packageName string, w worker, deps m
 	}
 }
 
+// meant to return a run option for mounting all repo state
+func withRepoData(repos []dalec.PackageRepositoryConfig, sOpts dalec.SourceOpts, opts ...llb.ConstraintsOpt) (llb.RunOption, error) {
```
cpuguy83, line 184: Can we split all these repo functions to a separate file?

cpuguy83 (changes requested):
Just one minor thing but otherwise this LGTM
Great work!

frontend/azlinux/handle_rpm.go
adamperlin: I actually just did this as part of the work I started for Jammy support! But actually, I could move those changes into this PR.

cpuguy83 (approved):

PR #368, opened 2024-09-11
fix: Use correct key for setting main build context

cpuguy83:
The signer implementation was not using standard docker keys for setting the main build context, this made it so you can't use signers as standalone frontends executing by `docker build`.

This is a breaking change as anything implementing a signer will need to be updated to use the correct key.


pmengelbert:
Needs a rebase, and there are conflicts. Otherwise LGTM, I'll repove once everything's green after the rebase

PR #384, opened 2024-10-02
spec.image.postinstall.symlink: Add CreateDir option

burak-ok:

Some images need binaries and files to be in some other places than traditional packages. To enable this the `Symlinks` field was introduced.
But if one wants to create a symlink in an non existing directory the creation fails.

This PR adds an option to create the underlying directory if it does not exist.

1. ~~As far as I understood the Windows path does not need to be update because it creates the directories every time~~ Windows path doesn't create the dir, too
2. It can be argued that this shouldn't be an option and the directory creation should be done every time. This also aligns with 1.

cc @blanquicet

### DIFF
--- a/test/azlinux_test.go
+++ b/test/azlinux_test.go
@@ -316,6 +316,14 @@ index 0000000..5260cb1
 						Name: src2Patch3ContextName,
 					},
 				},
+				"src3": {
+					Inline: &dalec.SourceInline{
+						File: &dalec.SourceInlineFile{
+							Contents:    "#!/usr/bin/env bash\necho goodbye",
+							Permissions: 0o700,
+						},
+					},
+				},
 			},
 			Patches: map[string][]dalec.PatchSpec{
 				"src2": {
@@ -394,6 +402,7 @@ echo "$BAR" > bar.txt
 				Post: &dalec.PostInstall{
 					Symlinks: map[string]dalec.SymlinkTarget{
 						"/usr/bin/src1": {Path: "/src1"},
+						"/usr/bin/src3": {Path: "/non/existing/dir/src3"},
 					},
 				},
 			},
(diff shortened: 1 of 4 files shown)


cpuguy83 (changes requested):
I think we can just update the implementation to automatically create the parent dir.

adamperlin:
I think I agree with @cpuguy83 here. Is there ever a case where someone might wish to create a nested symlink in Dalec but have the operation fail if a parent dir doesn't exist?

burak-ok:
Updated it to create the dir always

> Is there ever a case where someone might wish to create a nested symlink in Dalec but have the operation fail if a parent dir doesn't exist

I don't know :D That is why I had introduced the logic

test/azlinux_test.go
```diff
@@ -394,6 +394,7 @@ echo "$BAR" > bar.txt
 				Post: &dalec.PostInstall{
 					Symlinks: map[string]dalec.SymlinkTarget{
 						"/usr/bin/src1": {Path: "/src1"},
+						"/usr/bin/src2": {Path: "/non/exists/dir/src2"},
```
cpuguy83, line 397: I think this is not working because you haven't added `src2` (and also src2 is a dir) to the list of binary artifacts.
I think you'll need to add another source file with the content you want to check for.

cpuguy83 (changes requested):

test/azlinux_test.go
burak-ok: I guess I looked at the wrong line when I looked up `src2`: https://github.com/Azure/dalec/blob/main/test/azlinux_test.go#L1026

Thanks for spotting that

test/azlinux_test.go
```diff
@@ -394,6 +402,7 @@ echo "$BAR" > bar.txt
 				Post: &dalec.PostInstall{
 					Symlinks: map[string]dalec.SymlinkTarget{
 						"/usr/bin/src1": {Path: "/src1"},
+						"/usr/bin/src3": {Path: "/non/existing/dir/src3"},
```
cpuguy83, line 405: You'll need to include `src3` as a binary in the `artifacts` section.

cpuguy83 (approved):
LGTM assuming fresh test run passes.

PR #386, opened 2024-10-03
Support customizing git auth

cpuguy83:
Allows the client to pass in custom git auth secrets.
Also documents the existing defaults that are used.


sozercan (approved):
LGTM

cpuguy83:
Made a minor doc string fixup (bad copy/paste)

PR #385, opened 2024-10-03
Add support for headers artifacts

blanquicet:
**What this PR does / why we need it**: This PR adds support for header artifacts.

**Which issue(s) this PR fixes**: We want to give dalec users the ability to add header artifacts to their packages.

**Special notes for your reviewer**:

PR #387, opened 2024-10-04
deb: Use custom install script instead of dh-exec

cpuguy83:
dh-exec has some limitations that we can't really get around. The immediate need is it can't copy a directory to a new name in the target location.
As such we need more flexibility in how we set things up, so this switches to bash so we can have the power that we need.

Should resolve the issue with CI in #385

### DIFF
--- a/frontend/deb/debroot.go
+++ b/frontend/deb/debroot.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"bytes"
 	"context"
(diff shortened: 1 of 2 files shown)


blanquicet (approved):
It indeed fixes issues in https://github.com/Azure/dalec/pull/385.

cpuguy83:
Merged as part of #385

PR #396, opened 2024-10-09
deb: Disable dh_dwz which can cause issues for go

cpuguy83:
debbuild will automatically execute this `dh_dwz` target which, for some go builds, will error out like:

```
0.69 dwz: debian/msft-golang/usr/lib/msft-golang/bin/go: Found compressed .debug_abbrev section, not attempting dwz compression
40.69 dwz: debian/msft-golang/usr/lib/msft-golang/bin/gofmt: Found compressed .debug_abbrev section, not attempting dwz compression
40.69 dwz: Too few files for multifile optimization
```

This does what dh_golang does and overrides this so the target doesn't do anything.
We may want to make this opt-in in the future somehow (or scope it down to just go builds?).
Not really sure what a good approach will be here, but for now this is blocking being able to build certain packages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,28 +8,22 @@ on:
   pull_request:
     branches:
       - main
-    paths:
-        - "*.go"
-        - "**/*.go"
-        - Dockerfile
-        - docker-bake.hcl
-        - .github/workflows/ci.yml
-        - frontend/mariner2/Dockerfile
-        - "test/fixtures/*"
-        - go.mod
-        - go.sum
+    paths-ignore:
+      - 'website/**'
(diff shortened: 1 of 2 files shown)


adamperlin (approved):
I think this is an ok fix for now as it seems it is blocking, but it would be nice to try and make it configurable/scope to particular package builds in the future.

djsly:
Also affected by this. 

https://github.com/Azure/dalec-build-defs/actions/runs/11688802257/job/32550164925?pr=532

cpuguy83:
Noticed that CI was not triggering on this due to filter rules in the github actions yaml.
I pushed a commit to resolve this.

PR #397, opened 2024-10-10
Run package spec tests for packaging targets

cpuguy83:
Before this the package tests were only run when building a container and not when just building a package.


DannyBrito (approved):
lgtm

PR #402, opened 2024-10-14
add multi-module gen support

### LINKED ISSUE #392: [REQ] Support multi-module repos for gomod generator

New feature

Currently the gomod generator only supports generating gomods for a single module.
In order to have multiple modules one would have to define this as multiple sources.

I propose we add a field to the gomod generator type to allow specifying multiple module paths to generate from a single source.

Example:

```yaml
source:
  git:
    url: https://github.com/some/repo
    commit: v42.0.0
    generate:
      gomod:
        paths:
          - .
          - some/subpath
          - some/other/subpath
```

Where the implied default with no paths are specified is the working directory, e.g. `["."]`

DannyBrito:

Fixes #392

### DIFF
--- a/generator_gomod.go
+++ b/generator_gomod.go
@@ -40,19 +40,17 @@ func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.Constrai
 		if g.Gomod.Paths == nil {
 			paths = []string{"."}
 		}
-		states := make([]llb.State, 0, len(paths))
 
 		for _, path := range paths {
-			currentState := worker.Run(
+			in = worker.Run(
 				ShArgs("go mod download"),
 				llb.AddEnv("GOMODCACHE", gomodCacheDir),
 				llb.Dir(filepath.Join(joinedWorkDir, path)),
 				srcMount,
 				WithConstraints(opts...),
 			).AddMount(gomodCacheDir, in)
-			states = append(states, currentState)
 		}
-		return MergeAtPath(in, states, "/")
+		return in
 	}
 }
 
--- a/test/source_test.go
+++ b/test/source_test.go
@@ -363,42 +363,12 @@ index ea874f5..ba38f84 100644
 		}
 	}
 
-	tests := map[string]struct {
-		module       string
-		sourceInline *dalec.SourceInline
-	}{
-		"no patch": {
-			module: "github.com/cpuguy83/tar2go@v0.3.1",
-		},
-		"alternative go fixture": {
-			module: "github.com/stretchr/testify@v1.7.0",
-			sourceInline: &dalec.SourceInline{
-				Dir: &dalec.SourceInlineDir{
-					Files: map[string]*dalec.SourceInlineFile{
-						"main.go": {Contents: alternativeGomodFixtureMain},
-						"go.mod":  {Contents: alternativeGomodFixtureMod},
-						"go.sum":  {Contents: alternativeGomodFixtureSum},
-					},
-				},
-			},
-		},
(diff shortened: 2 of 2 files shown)


DannyBrito:
This is failing.
In a scenario that in theory it should work:

```yml
  src:
    path: test/fixtures/multi-modules
    context: {}
    generate:
      - gomod:
          paths:
            - module1
            - module2
        subpath: project
```

load.go
```diff
@@ -171,6 +171,15 @@ func fillDefaults(s *Source) {
 		fillDefaults(&s.Build.Source)
 	case s.Inline != nil:
 	}
+
+	// Fill in default paths for gomod generators
+	for _, generator := range s.Generate {
+		if generator.Gomod != nil {
+			if generator.Gomod.Paths == nil {
+				generator.Gomod.Paths = []string{"."}
```
cpuguy83, line 179: I think it may be cleaner to handle this in the gomod generator itself.

test/fixtures/multi-go-modules.yml
```diff
@@ -0,0 +1,46 @@
+# syntax=ghcr.io/azure/dalec/frontend:latest
+
+name: dalec-test-multi-go-modules-build
```
cpuguy83, line 3: Can we migrate this into an integration test under `test/`?
There should already be one in there for gomods we can extend for multi-module.

cpuguy83 (changes requested):

cpuguy83:
Also looks like this needs a `go generate ./...` so it updates the json schema.

Otherwise this seems like its working.

test/source_test.go
```diff
@@ -375,6 +448,77 @@ index ea874f5..ba38f84 100644
 	})
 }
 
+func TestMultiGoModuleGen(t *testing.T) {
```
cpuguy83, line 451: Can this be a subtest of `TestSourceWithGomod`?

generator_gomod.go
```diff
@@ -30,20 +32,27 @@ func (s *Spec) HasGomods() bool {
...
-			srcMount = llb.AddMount(workDir, srcSt)
+		workDir := "/work/src"
+		joinedWorkDir := filepath.Join(workDir, g.Subpath)
+		srcMount := llb.AddMount(workDir, srcSt)
+
+		paths := g.Gomod.Paths
+		if g.Gomod.Paths == nil {
+			paths = []string{"."}
+		}
+		states := make([]llb.State, 0, len(paths))
```
cpuguy83, line 43: Since there can be duplicate modules that need to get downloaded, instead of using multiple states and merging them, can we build up a single state?

test/source_test.go
```diff
@@ -320,12 +363,42 @@ index ea874f5..ba38f84 100644
 		}
 	}
 
-	t.Run("no patch", func(t *testing.T) {
-		t.Parallel()
-		testEnv.RunTest(baseCtx, t, func(ctx context.Context, gwc gwclient.Client) {
-			checkModule(ctx, gwc, "github.com/cpuguy83/tar2go@v0.3.1", baseSpec())
+	tests := map[string]struct {
```
cpuguy83, line 366: I'm thinking we probably don't need this extra test since it is going to add a fair amount of test time  and more potential for flakiness.

cpuguy83 (changes requested):

test/source_test.go
DannyBrito: ok, I drop this extra test

cpuguy83 (approved):

PR #404, opened 2024-10-16
Force evaluation before test execution

### LINKED ISSUE #400: Force the state to be evaluated before running package tests
Right now the solved state for packages and containers are not actually evaluated until it hits the test runner, which itself forces evaluation since it needs to inspect some of the real state.
This can make the errors confusing since it looks like a test failed when in reality the build just failed.

Since we are always going to need to evaluate the state to run tests (assuming there are tests), we should just always force the evaluation before passing it off to the test runner.

cpuguy83:
This makes it so error messages from the build are not wrapped in error messages for tests.

Before this change the build is performed lazily until the point when the tests try to interact with the rootfs under test making it appear like a build failure is a test error when in reality we didn't even get to the test.

Closes #400

### DIFF
--- a/frontend/azlinux/handle_rpm.go
+++ b/frontend/azlinux/handle_rpm.go
@@ -49,6 +49,10 @@ func handleRPM(w worker) gwclient.BuildFunc {
 				return nil, nil, err
 			}
 
+			if err := ref.Evaluate(ctx); err != nil {
+				return ref, nil, err
+			}
+
 			if imgRef, err := runTests(ctx, client, w, spec, sOpt, st, targetKey, pg); err != nil {
 				// return the container ref in case of error so it can be used to debug
 				// the installed package state.
--- a/frontend/test_runner.go
+++ b/frontend/test_runner.go
@@ -41,6 +41,14 @@ func RunTests(ctx context.Context, client gwclient.Client, spec *dalec.Spec, ref
 		return nil
 	}
 
+	if err := ref.Evaluate(ctx); err != nil {
+		// Force evaluation here so that any errors for the build itself can surface
+		// more cleanly.
+		// Otherwise an error for something wrong in the build (e.g. a failed compilation)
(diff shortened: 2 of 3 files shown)


frontend/azlinux/handle_rpm.go
```diff
@@ -49,6 +49,10 @@ func handleRPM(w worker) gwclient.BuildFunc {
 				return nil, nil, err
 			}
 
+			if err := ref.Evaluate(ctx); err != nil {
```
adamperlin, line 52: Since `runTests` now constructs the state that installs packages in the target container, this will actually unlazy the package build before the install in the target container, right?

cpuguy83: Correct.

frontend/test_runner.go
```diff
@@ -41,6 +41,14 @@ func RunTests(ctx context.Context, client gwclient.Client, spec *dalec.Spec, ref
 		return nil
 	}
 
+	if err := ref.Evaluate(ctx); err != nil {
```
adamperlin, line 44: Since we're already doing `ref.Evaluate` after we create package build state for `deb`/`rpm`, really the only extra errors this will surface here would be install errors from installing the package in the target container, correct? That should in general be the only extra state that hasn't yet been solved at this point?

cpuguy83: At this point, the package build state is only evaluated on the packaging build targets, not the container targets.
But other than that, yes you are correct.

adamperlin: Ah got it, the build state hasn't yet been evaluated if we've reached this code path from a container target. And in the case of packaging build targets, this `ref.Evaluate` would actually surface errors from installing in the **test** container, not the **target** container since there is no target container in that case. Thanks for clarifying

adamperlin (approved):
I think this LGTM!

PR #408, opened 2024-10-21
Enable libexec artifacts

pmengelbert:
This commit allows the user to specify artifacts to be installed under `usr/libexec`. I used the following documentation as a guide:

[/usr/libexec](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html#:~:text=Purpose,subdirectory%20under%20%2Fusr%2Flibexec%20.)

Below, you can see at a glance what the behavior of various settings will be. The following examples assume we are building a spec with top-level `name: docker`. The subpath field defaults to the package name.

```yaml
# goes to /usr/libexec/docker/docker-compose
libexec:
  bin/docker-compose:

# goes to /usr/libexec/docker/cli-plugins/docker-compose
libexec:
  bin/docker-compose:
    subpath: docker/cli-plugins

# goes to /usr/libexec/something_else/cli-plugins/docker-compose
libexec:
  bin/docker-compose:
    subpath: something_else/cli-plugins

# goes to /usr/libexec/docker/cli-plugins/hello
libexec:
  bin/docker-compose:
    subpath: docker/cli-plugins
    name: hello

# goes to /usr/libexec/docker/hello
libexec:
  bin/docker-compose:
    name: hello

```

Addresses (but does not finish) #242 


frontend/rpm/template.go
```diff
@@ -499,6 +499,14 @@ func (w *specWrapper) Install() fmt.Stringer {
 		}
 	}
 
+	if w.Spec.Artifacts.Libexec != nil {
```
cpuguy83, line 502: can you do the same in `frontend/deb` and add a test in `test/` with the other artifact types?

cpuguy83 (changes requested):

frontend/rpm/template.go
pmengelbert: Yes

pmengelbert: But will I?

(the answer is still yes)

pmengelbert: done

load.go
```diff
@@ -445,6 +445,13 @@ func (s *Spec) FillDefaults() {
 			s.Patches[k][i].Strip = &strip
 		}
 	}
+
+	for k, ac := range s.Artifacts.Libexec {
```
cpuguy83, line 449: We're not doing this for any of the other artifact types, I think it would be good to leave this up to the underlying implementation to deal with rather than changing it in the spec.

cpuguy83 (changes requested):

load.go
pmengelbert: Makes sense

pmengelbert: done

frontend/template.go
```diff
@@ -0,0 +1,11 @@
+package frontend
+
+import "github.com/Azure/dalec"
+
+func DefaultLibexecSubpath(s *dalec.Spec, k string) string {
```
cpuguy83, line 5: This doesn't look right.
The way we handle this for the other types (like libs) is always `<base>/<spec name>/<subpath>/<artifact name>`

As an example, see how `Libs` are handled.

Also at this time I'd like to avoid abstracting the behavior here.

cpuguy83 (changes requested):

frontend/template.go
cpuguy83: Let me double check a real system for common libexec usage, though.

cpuguy83: Yeah, I think that's the way to go, and if someone needs something directly in `/usr/libexec` or somewhere else they can use a symlink.

pmengelbert: What about when packages are renamed from upstream? For example, if we have the moby-buildx package, we want the binary to go into `/usr/libexec/docker/cli-plugins/docker-buildx` rather than `/usr/libexec/moby-buildx/cli-plugins/docker-buildx`. IMO we should let the user do the right thing without adding extra steps.

pmengelbert: > This doesn't look right. The way we handle this for the other types (like libs) is always `<base>/<spec name>/<subpath>/<artifact name>`

That seems pretty non-standard. Pretty much everything in my `/usr/libexec` directory is either in `/usr/libexec` directly, or nested under a single subdirectory bearing the main application's name. There's only a few files nested deeper than the application's name.

> If someone needs something directly in `/usr/libexec` or somewhere else they can use a symlink.

That would violate the standard referenced above: "Applications may use a single subdirectory under /usr/libexec". If we create a symlink, we have one application (docker) with two subdirectorys under `/usr/libexec` (`/usr/libexec/docker` and `/usr/libexec/moby-buildx`).

cpuguy83: Maybe a separate field to overwrite the package name is worthwhile in this case.
Or just a bool like `OmitAutoPackageName`

pmengelbert: done

artifacts.go
```diff
@@ -97,6 +97,9 @@ type ArtifactConfig struct {
 	// Name is file or dir name to use for the artifact in the package.
 	// If empty, the file or dir name from the produced artifact will be used.
 	Name string `yaml:"name,omitempty" json:"name,omitempty"`
+	// PackageName is the name of the package. This can be used to replace
+	// `spec.Name` when destination paths are generated.
+	PackageName string `yaml:"packageName,omitempty" json:"packageName,omitempty"`
```
cpuguy83, line 102: My reasoning for suggesting a bool here is because the package name can go into `SubPath` and we'd just skip including the spec's package name with that.

cpuguy83 (changes requested):

artifacts.go
pmengelbert: Next time I will clarify what you want before I implement the solution. I thought you meant that either was acceptable

pmengelbert: tldr: I think we should just use subPath and not do anything automatically. With an empty `subPath`, things go directly into `/usr/libexec` (which is standard). Otherwise, you can set `subPath` to `docker` if you want things to go into `/usr/libexec/docker` or `docker/cli-plugins` if you want `/usr/libexec/docker/cli-plugins`. Detailed explanation follows.

> My reasoning for suggesting a bool here is because the package name can go into `SubPath` and we'd just skip including the spec's package name with that.

So then in order to get the desired behavior in the example I've been using, you would need

```yaml
name: moby-buildx

artifacts:
  libexec:
    bin/docker-buildx:
      subPath: docker/cli-plugins
      omitAutoPackageName: true 
```

What I don't like about that is it kind of overloads the `subPath` field. When we don't need to change the package name, it would be

```yaml
name: docker

artifacts:
  libexec:
    bin/docker-buildx:
      subPath: cli-plugins
```

Both with the bool and without, it's a leaky abstraction. The user has to know several things:
1. the package name is inserted some times but not others (yes in /usr/lib artifacts, no in /usr/bin artifacts)
1. libexec will be one of the times when the name is inserted
1. in order to avoid inserting the name you need to set the bool
1. the bool changes the meaning of the `subPath` field, and changes the absolute path depth.

It's too much to ask of the user IMO. Better to just have `/usr/libexec` be the default (since that's a pretty standard thing to do) and have the user add the application name to the subpath if they want things nested in a subdirectory..

cpuguy83 (approved):

PR #415, opened 2024-10-24
Fix custom repo test

cpuguy83:
A few fixes here:

1. Minor refactor on custom repo tests so that the proper `*testing.T` is being used. This fixes a problem with the test build not producing any logs making it hard to figure out what the error is.
2. Namespace dep package names per test so that we don't have 2 parallel tests using the same package name, potentially creating a race where the package is cached and re-used in the other test.
3. Remove cached repo data for local repos (e.g., ones created in /top/repo) to fix the case where a repo can appear to become unsigned causing apt to error out.

### DIFF
--- a/frontend/jammy/handle_deb.go
+++ b/frontend/jammy/handle_deb.go
@@ -170,6 +170,11 @@ apt install -y `+strings.Join(ls, " ")+`
 
 func installWithConstraints(pkgPath string, pkgName string, opts ...llb.ConstraintsOpt) llb.RunOption {
 	return dalec.RunOptFunc(func(ei *llb.ExecInfo) {
+		// The apt solver always tries to select the latest package version even when constraints specify that an older version should be installed and that older version is available in a repo.
+		// This leads the solver to simply refuse to install our target package if the latest version of ANY dependency package is incompatible with the constraints.
+		// To work around this we first install the .deb for the package with dpkg, specifically ignoring any dependencies so that we can avoid the constraints issue.
+		// We then use aptitude to fix the (possibly broken) install of the package, and we pass the aptitude solver a hint to REJECT any solution that involves uninstalling the package.
+		// This forces aptitude to find a solution that will respect the constraints even if the solution involves pinning dependency packages to older versions.
 		script := llb.Scratch().File(
 			llb.Mkfile("install.sh", 0o755, []byte(`#!/usr/bin/env sh
 set -ex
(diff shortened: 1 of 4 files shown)


cpuguy83:
Ok, so the first reason the test was failing is the cached repo info in `/var/lib/apt/lists` for the local repo (at /opt/repo).
Particularly problematic due to using the same repo path for all the repo tests and the repo gets flipped from signed to unsigned and apt does not like that.

In general I think we don't want to keep any data about local repo mounts between builds, so fixed this by clearing any local repo data in that dir (local repos apparently start with `_` so `rm -f /var/lib/apt/lists/_*` did the trick.

Now it looks like the custom repo just isn't working for some reason when doing the installation of runtime dependencies. Looking into this.

Depending on the ordering of tests this could pass but usually fails in CI. Locally it's harder to make it fail but it does happen.

cpuguy83:
Ok, realized the extra failure was due to the same exact cause, there's just 2 places we are installing packages from and the fix needed to be applied in the 2nd case.

frontend/jammy/handle_deb.go
```diff
@@ -285,7 +317,8 @@ func buildDepends(worker llb.State, sOpt dalec.SourceOpts, spec *dalec.Spec, tar
 		)
 
 		return in.Run(
-			installWithConstraints(debPath+"/*.deb", depsSpec.Name, customRepoOpts),
+			installWithConstraints(debPath+"/*.deb", depsSpec.Name, opts...),
+			customRepoOpts,
```
adamperlin, line 321: Thank you for cleaning this up and addressing your review suggestion from before to add the `customRepoOpts` instead of threading through `install...`

frontend/jammy/handle_deb.go
```diff
@@ -145,26 +144,59 @@ func customRepoMounts(worker llb.State, repos []dalec.PackageRepositoryConfig, s
...
 
-func installPackages(ls []string, rOpts ...llb.RunOption) llb.RunOption {
+func installPackages(ls []string, opts ...llb.ConstraintsOpt) llb.RunOption {
+	script := llb.Scratch().File(
+		llb.Mkfile("install.sh", 0o755, []byte(`#!/usr/bin/env sh
+set -ex
+
+# Make sure any cached data from local repos is purged since this should not
+# be shared between builds.
+rm -f /var/lib/apt/lists/_*
```
adamperlin, line 154: cached local repo data is always under a dir prefixed by `_`?

adamperlin (approved):
I think this looks good to me, and the fixes seems very straightforward. Thank you for fixing these issues!

frontend/jammy/handle_deb.go
cpuguy83: Based on what I've seen, yes.

PR #420, opened 2024-10-30
Add support for allowing network access during build

cpuguy83:
Adds a field `network_mode`. The value of this field can be any of:

- Unset
- "none" - Disable networking (default)
- "sandbox" - Networking enabled with its own network namespace

Buildkit supports `host` mode as well, but I have opted to not support that here for now.

### DIFF
--- a/test/network_test.go
+++ b/test/network_test.go
@@ -1,0 +1,83 @@
+package test
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/Azure/dalec"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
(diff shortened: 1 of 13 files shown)


test/network_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/Azure/dalec"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+	moby_buildkit_v1_frontend "github.com/moby/buildkit/frontend/gateway/pb"
+	"gotest.tools/v3/assert"
+)
+
+func testBuildNetworkMode(ctx context.Context, t *testing.T, cfg targetConfig) {
+	type testCase struct {
+		mode            string
+		canHazInternetz bool // :)
```
sozercan, line 18: 😹

sozercan (approved):
minor comment, otherwise LGTM

PR #421, opened 2024-10-30
Various improvements to testing

### LINKED ISSUE #476: [BUG] Integration tests possible not reporting failures
### Expected Behavior

https://github.com/Azure/dalec/actions/runs/12283069321/job/34275695113?pr=472#step:9:158,
It seems that the test is timeout or not really sure the cause but there is a panic and CI is not failing.
Not sure if that is expected?

### Actual Behavior

I would expect the integration tests to mark CI as failure

### Steps To Reproduce

_No response_

cpuguy83:
- Add some missing `t.Parallal()` and `startTestSpan` calls
- Fix test log output to print line by line instead of per read
- Use the same cache keys for the main Dockerfile and the test fixtures
- Add test stats to GHA summary
- Get tracing reports from integration tests
- Remove GH cache since this wasn't really giving us anything, and maybe was even slowing things down
- ci: Fix issue reporting failures from test panic
- ci: Split integration tests per distro into separate jobs

Supersedes #482 
Fixes #476

### DIFF
--- a/cmd/test2json2gha/event.go
+++ b/cmd/test2json2gha/event.go
@@ -1,0 +1,56 @@
+package main
+
+import (
+	"os"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// TestEvent is the go test2json event data structure we receive from `go test`
+// This is defined in https://pkg.go.dev/cmd/test2json#hdr-Output_Format
+type TestEvent struct {
+	Time    time.Time
+	Action  string
+	Package string
+	Test    string
+	Elapsed float64 // seconds
+	Output  string
+}
+
+// TestResult is where we collect all the data about a test
+type TestResult struct {
+	output  *os.File
+	failed  bool
+	pkg     string
+	name    string
+	elapsed float64
+	skipped bool
+}
+
+func (r *TestResult) Close() {
+	r.output.Close()
+}
+
+func collectTestOutput(te *TestEvent, tr *TestResult) error {
+	if te.Output != "" {
+		_, err := tr.output.Write([]byte(te.Output))
+		if err != nil {
+			return errors.Wrap(err, "error collecting test event output")
+		}
+	}
+
+	tr.pkg = te.Package
+	tr.name = te.Test
+	if te.Elapsed > 0 {
+		tr.elapsed = te.Elapsed
+	}
+
+	if te.Action == "fail" {
+		tr.failed = true
+	}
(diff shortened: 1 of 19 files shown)


cpuguy83:
The test failure is actually on main... I'm bisecting this now.

cpuguy83:
Pushed commit to fix.

test/testenv/buildx.go
```diff
@@ -290,7 +290,6 @@ func (b *BuildxEnv) RunTest(ctx context.Context, t *testing.T, f TestFunc, opts
 
 	var so client.SolveOpt
 	withProjectRoot(t, &so)
-	withGHCache(t, &so)
```
adamperlin, line 293: Do we have any idea why the GH cache was so slow?

cpuguy83: A couple of things:

1. It's just slow to transfer all that
2. The cache is going to get overwritten by each PR
3. Having separate cache scopes for each test, as is done here, means very little reuse between tests and just a bloated cache.

All in all, I found that the cache wasn't giving us any speedup.

adamperlin: Thank you for the detailed answer on this, that makes a lot of sense.

adamperlin (changes requested):
Just some minor comments

cmd/test2json2gha/main.go
```diff
@@ -145,28 +152,42 @@ func do(in io.Reader, out io.Writer, modName string) (bool, error) {
...
 
-	if te.Action == "fail" {
-		tr.failed = true
+	return failCount > 0, nil
+}
+
+func (c *nopWriteCloser) Close() error { return nil }
+
+func getSummaryFile() io.WriteCloser {
+	v := os.Getenv("GITHUB_STEP_SUMMARY")
```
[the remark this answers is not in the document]
cpuguy83: Added

cmd/test2json2gha/event.go
```diff
@@ -0,0 +1,56 @@
...
+	name    string
+	elapsed float64
+	skipped bool
+}
+
+func (r *TestResult) Close() {
+	r.output.Close()
+}
+
+func handleEvent(te *TestEvent, tr *TestResult) error {
```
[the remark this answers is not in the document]
cpuguy83: Changed

adamperlin (approved):

PR #422, opened 2024-10-31
windows: Add pinned deps and custom repo support

cpuguy83:
This is mostly copied from the jammy implementation since the windows target is currently using jammy as the build worker.

### DIFF
--- a/frontend/jammy/handle_deb.go
+++ b/frontend/jammy/handle_deb.go
@@ -126,6 +126,7 @@ var jammyRepoPlatformCfg = dalec.RepoPlatformConfig{
 }
 
 func customRepoMounts(worker llb.State, repos []dalec.PackageRepositoryConfig, sOpt dalec.SourceOpts, opts ...llb.ConstraintsOpt) (llb.RunOption, error) {
+	worker = worker.Run(installPackages([]string{"gnupg2"}, opts...), dalec.WithConstraints(opts...)).Root() // make sure we have gpg installed
 	withRepos, err := dalec.WithRepoConfigs(repos, &jammyRepoPlatformCfg, sOpt, opts...)
 	if err != nil {
 		return nil, err
--- a/frontend/windows/handle_zip.go
+++ b/frontend/windows/handle_zip.go
@@ -9,12 +9,12 @@ import (
 
 	"github.com/Azure/dalec"
 	"github.com/Azure/dalec/frontend"
+	"github.com/Azure/dalec/frontend/deb"
 	"github.com/moby/buildkit/client/llb"
 	gwclient "github.com/moby/buildkit/frontend/gateway/client"
 	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"golang.org/x/exp/maps"
-	"golang.org/x/exp/slices"
 )
 
 const (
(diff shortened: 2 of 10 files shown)


frontend/jammy/handle_deb.go
```diff
@@ -269,7 +270,7 @@ func basePackages(opts ...llb.ConstraintsOpt) llb.StateOption {
 	return func(in llb.State) llb.State {
 		opts = append(opts, dalec.ProgressGroup("Install base packages"))
 		return in.Run(
-			installPackages([]string{"aptitude", "dpkg-dev", "devscripts", "equivs", "fakeroot", "dh-make", "build-essential", "dh-apparmor", "dh-make", "dh-exec", "debhelper-compat=" + deb.DebHelperCompat}, opts...),
+			installPackages([]string{"aptitude", "dpkg-dev", "devscripts", "equivs", "fakeroot", "dh-make", "build-essential", "dh-apparmor", "dh-make", "dh-exec", "debhelper-compat=" + deb.DebHelperCompat, "gnupg2"}, opts...),
```
adamperlin, line 273: Why did `gnupg` need to be added here?

cpuguy83: Forgot I added this, I'll remove it.

frontend/windows/handle_zip.go
```diff
@@ -175,7 +245,7 @@ func generateInvocationScript(binaries []string) *strings.Builder {
 	return script
 }
 
-func workerImg(sOpt dalec.SourceOpts, opts ...llb.ConstraintsOpt) (llb.State, error) {
+func workerImg(sOpt dalec.SourceOpts, spec *dalec.Spec, targetKey string, opts ...llb.ConstraintsOpt) (llb.State, error) {
```
adamperlin, line 248: Are these new params needed?

test/azlinux_test.go
```diff
@@ -111,6 +112,23 @@ func TestAzlinux3(t *testing.T) {
 	})
 }
 
+func signRepoAzLinux(gpgKey llb.State) llb.StateOption {
+	// key should be a state that has a public key under /public.key
+	return func(in llb.State) llb.State {
+		return in.Run(
+			dalec.ShArgs("gpg --import < /tmp/gpg/PUBLIC-RPM-GPG-KEY"),
```
adamperlin, line 119: This path isn't correct anymore, it should be under `/tmp/gpg/public.key` in the `gpgKey` state. I think the reason it's still working though is because the public key doesn't actually need to be imported successfully to do the signing, so can we remove this line?

helpers.go
```diff
@@ -542,7 +548,7 @@ func GetRepoKeys(worker llb.State, configs []PackageRepositoryConfig, cfg *RepoP
 				outPath := filepath.Join("/tmp/out", name)
 				keySt := worker.Run(
 					// dearmor key if necessary
-					ShArgs(fmt.Sprintf("cat '%s' | gpg --dearmor --output '%s'", inPath, outPath)),
+					ShArgs(fmt.Sprintf("gpg --dearmor --output '%s' < %q", outPath, inPath)),
```
adamperlin, line 551: Can we be consistent with `'%s'` vs `%q` here?

adamperlin (changes requested):

test/azlinux_test.go
cpuguy83: Hah, yeah the 1st line is not doing anything at all since the second `dalec.ShArgs` is replacing it.

helpers.go
cpuguy83: Fixed.

test/azlinux_test.go
cpuguy83: Fixed.

frontend/windows/handle_zip.go
cpuguy83: Fixed.

adamperlin (approved):
This looks good to me now!

PR #434, opened 2024-11-13
Custom Repo Fixes

### LINKED ISSUE #425: Fill defaults for sources in the extra_repos section of the spec
### Expected Behavior

The sources listed under the `extra_repos` section in the Dalec spec which are used to fetch gpg keys and repo configs currently do not have defaults filled when the spec is loaded. So for example, `context: {}` as a source for a key or repo config is currently broken. This needs to be fixed

### Actual Behavior

_No response_

### Steps To Reproduce

_No response_

adamperlin:

This PR adds a few fixes for the custom repos feature. 

1. Adds code to fill defaults in the extra_repos section of the spec.
3. Adds a `permissions` field to the http source. This allows us to fill in the permissions by default if they are on an http source that is nested under a key. It is important that fetched gpg keys have the proper permissions, otherwise apt cannot import them properly. Assuming that they do have proper permissions, apt can handle both `.asc` and `.gpg` keys just fine. This makes the explicit de-armor step unnecessary and means we no longer need gpg as a dependency in the jammy worker images.
4. Adds armored and de-armored key test variants to the custom repo tests to ensure everything is working properly. 

Fixes #425

### DIFF
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -836,6 +836,10 @@
 				"digest": {
 					"type": "string",
 					"description": "Digest is the digest of the file to download.\nThis is used to verify the integrity of the file.\nForm: \u003calgorithm\u003e:\u003cdigest\u003e"
+				},
+				"permissions": {
+					"type": "integer",
+					"description": "Permissions is the octal file permissions to set on the file."
 				}
 			},
 			"additionalProperties": false,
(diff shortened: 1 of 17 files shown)


cpuguy83:
So this is no longer dearmoring keys at all?

adamperlin:
> So this is no longer dearmoring keys at all?

Yes. It should not be needed as long as the file extension on the keys is correct.

cpuguy83:
There may be something in the docs about this, can you update that?

cpuguy83 (approved):

PR #438, opened 2024-11-20
add multi-module gomod support docs

DannyBrito:

Fixes #

### DIFF
--- a/website/docs/sources.md
+++ b/website/docs/sources.md
@@ -97,7 +97,7 @@ by the client, not the actual secret values.
 
 ### HTTP
 
-HTTP sources fetch a file from an HTTP URL. The HTTP source type is considered to be a "file" source. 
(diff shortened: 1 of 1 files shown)


website/docs/sources.md
```diff
@@ -347,8 +347,33 @@ sources:
         gomod: {} # Generates a go module cache to cache dependencies
 ```
 
-`gomod` currently does not have any options but may in the future.
+The `gomod` generator also supports generating multiple modules in a single source. The `paths` field is a list of paths where the generator should fetch the dependencies. Assuming `src` looks like this:
 
+```
+test/fixtures/multi-modules/project/
```
cpuguy83, line 353: Can we change this to something simpler like `.`?

cpuguy83 (changes requested):

cpuguy83 (approved):

PR #439, opened 2024-11-21
Add more Ubuntu versions and Debian support

cpuguy83:
Individual commits detail each change.
Basically:

1. Added some best-effort support to support older go verisons that may not support the `GOMODCACHE` env var
2. Sets $PATH in the deb build scripts to what's in the worker image to improve reproducibility
3. Add support for non-standard Ubuntu/Debian go packages from backports that do not install `go` into $PATH
4. Add Ubuntu versions
5. Add Debian versions


website/docs/targets.md
```diff
@@ -11,29 +11,73 @@ Many components, such as package dependencies and base images, are specific to a
...
+$ docker buildx build --call targets --build-arg ghcr.io/azure/dalec/frontend:latest - <<< "null"
+TARGET                           DESCRIPTION
 azlinux3/container (default)     Builds a container image for
 azlinux3/container/depsonly      Builds a container image with only the runtime dependencies installed.
 azlinux3/rpm                     Builds an rpm and src.rpm.
 azlinux3/rpm/debug/buildroot     Outputs an rpm buildroot suitable for passing to rpmbuild.
 azlinux3/rpm/debug/sources       Outputs all the sources specified in the spec file in the format given to rpmbuild.
 azlinux3/rpm/debug/spec          Outputs the generated RPM spec file
 azlinux3/worker                  Builds the base worker image responsible for building the rpm
+bionic/deb (default)             Builds a deb package.
```
sozercan, line 23: i am confused how can we have different defaults?

cpuguy83: Its default for the distro.
We could just not set this but perhaps in a separate change.

frontend/rpm/handle_sources.go
```diff
@@ -77,6 +77,13 @@ func buildScript(spec *dalec.Spec) string {
 	fmt.Fprintln(b, "set -e")
 
 	if spec.HasGomods() {
+		// Older go versions did not have support for the `GOMODCACHE` var
+		// This is a hack to try and make the build work by linking the go modules
+		// we've already fetched into to module dir under $GOPATH
```
adamperlin, line 82: Can you add a comment mentioning that `${GOPATH}/pkg/mod` is the default module dir?

website/docs/targets.md
```diff
@@ -11,29 +11,35 @@ Many components, such as package dependencies and base images, are specific to a
...
-mariner2/container/depsonly      Builds a container image with only the runtime dependencies installed.
-mariner2/rpm                     Builds an rpm and src.rpm.
-mariner2/rpm/debug/buildroot     Outputs an rpm buildroot suitable for passing to rpmbuild.
-mariner2/rpm/debug/sources       Outputs all the sources specified in the spec file in the format given to rpmbuild.
-mariner2/rpm/debug/spec          Outputs the generated RPM spec file
-mariner2/worker                  Builds the base worker image responsible for building the rpm
-windowscross/container (default) Builds binaries and installs them into a Windows base image
-windowscross/worker              Builds the base worker image responsible for building the rpm
-windowscross/zip                 Builds binaries combined into a zip file
+$ docker buildx build --call targets --build-arg ghcr.io/azure/dalec/frontend:latest - <<< "null"
```
adamperlin, line 14:
```suggestion
$ docker buildx build --call targets --build-arg BUILDKIT_SYNTAX=ghcr.io/azure/dalec/frontend:latest - <<< "null"
```

adamperlin: Was this a typo? This is the only way I was able to get that command to work

adamperlin (changes requested):

cpuguy83:
Updated with suggestions.

frontend/deb/distro/worker.go
```diff
@@ -79,6 +79,7 @@ func (cfg *Config) Worker(sOpt dalec.SourceOpts, opts ...llb.ConstraintsOpt) (ll
 		Run(
 			dalec.WithConstraints(opts...),
 			AptInstall(cfg.BuilderPackages, opts...),
+			dalec.WithMountedAptCache(cfg.AptCachePrefix),
```
adamperlin, line 82: Was this just not using the shared apt cache before?

frontend/deb/template_rules.go
```diff
@@ -81,7 +81,12 @@ func (w *rulesWrapper) OverridePerms() fmt.Stringer {
 	}
 
 	if fixPerms {
-		b.WriteString("execute_after_dh_fixperms:\n")
+		// Normally this should be `execute_aftr_dh_fixperms`, howwever this doesn't
```
adamperlin, line 84:
```suggestion
		// Normally this should be `execute_after_dh_fixperms`, however this doesn't
```

frontend/deb/template_rules.go
```diff
@@ -81,7 +81,12 @@ func (w *rulesWrapper) OverridePerms() fmt.Stringer {
...
+		// work on UBuntu 18.04.
```
adamperlin, line 85:
```suggestion
		// work on Ubuntu 18.04.
```

frontend/mux.go
```diff
@@ -586,25 +586,52 @@ func WithTargetForwardingHandler(ctx context.Context, client gwclient.Client, m
...
+}
 
-		if len(spec.Targets) > 0 {
-			t, ok := spec.Targets[key]
-			if !ok {
-				bklog.G(ctx).WithField("spec targets", maps.Keys(spec.Targets)).WithField("targetKey", key).Info("Target not in the spec, skipping")
-				return nil
-			}
+// shouldLoadTarget is used to determine if the spec is overriding the built-in
+// target with the same targetKey.
```
adamperlin, line 598: Can you add a comment here to clarify the logic and rationale a bit more, i.e., we will always load a target if there are no targets specified in the spec, otherwise, we will only load the targets specified in the spec?

adamperlin (approved):
This looks good to me, just a few small things

frontend/deb/distro/worker.go
cpuguy83: Yes, somehow got left out when I moved things around.

frontend/mux.go
cpuguy83: Done.

frontend/deb/template_rules.go
cpuguy83: Fixed

frontend/deb/template_rules.go
cpuguy83: Fixed

PR #440, opened 2024-11-21
Update Quickstart Docs

adamperlin:

This changes the "building from source" section of the docs into a quick start guide by adding some additional background and context on Dalec, attempting to clarify some concepts up front which may be confusing. 

Another addition is the section on container only builds, which seeks to clarify this use-case of Dalec.

Fixes #


website/docs/container-only-builds.md
```diff
@@ -0,0 +1,29 @@
...
+        curl:
+        bash:
+
+image:
+    entrypoint: /bin/bash
+
+```
+
+Then, to build:
+`docker build -f my-minimal-image.yml --target=mariner2 -t my-minimal-image:0.1.0 .`
```
sozercan, line 26: do we need a docker version requirement or use `buildx build` explicitly? in older version of docker, `build` doesn't symlink to `buildx build`. do we think we need to highlity?

website/docs/quickstart.md
```diff
@@ -0,0 +1,211 @@
...
+title: Quickstart
+---
+
+In this section, we'll go over how to build packages and containers from source with Dalec. Note that, in this context, "source" refers to the source code of the package being built. Before continuing, it first useful to go over some preliminary background.
+
+## Some Preliminaries
+
+### But what actually *is* Dalec?
+
+Dalec is what is known as a *frontend* for [Docker Buildkit](https://docs.docker.com/build/buildkit/frontend/) . If you've ever used a `Dockerfile` before (under newer versions of docker) you have interacted with a buildkit frontend. A frontend is a little like a compiler in that it translates higher level syntax into specific build instructions which the buildkit engine knows how to execute. Dalec, then, provides a specialized spec format for specifying particular artifacts -- in this case packages and containers -- and then translates that spec into build instructions to be run in buildkit. This is why Dalec has no dependencies other than Docker -- it actually becomes a component loaded during the docker build.
```
sozercan, line 11:
```suggestion
Dalec is what is known as a *frontend* for [Docker Buildkit](https://docs.docker.com/build/buildkit/frontend/). If you've ever used a `Dockerfile` before (under newer versions of docker) you have interacted with a buildkit frontend. A frontend is a little like a compiler in that it translates higher level syntax into specific build instructions which the buildkit engine knows how to execute. Dalec, then, provides a specialized spec format for specifying particular artifacts -- in this case packages and containers -- and then translates that spec into build instructions to be run in buildkit. This is why Dalec has no dependencies other than Docker -- it actually becomes a component loaded during the docker build. 
```

website/docs/quickstart.md
```diff
@@ -0,0 +1,211 @@
...
+
+1. A list of sources to pull from
+2. A build script to build the sources
+3. A list of artifacts to include in the package
+
+In this example, we'll build the `go-md2man` package and container from the [`go-md2man`](https://github.com/cpuguy83/go-md2man) repo using `v2.0.3` tag in the repo.
+
+First, let's start with the constructing a [Dalec spec](spec.md) file.
+
+We define the metadata for the package in the spec. This includes the name, packager, vendor, license, website, and description of the package. You may notice that many of these fields appear in package manager metadata for rpm and deb packages. This is because, for certain targets, Dalec will generate package files for these package managers and will utilize this information for that purpose.
```
sozercan, line 43: can you call out the optional ones in here, like `website (optional)` etc

website/docs/quickstart.md
```diff
@@ -0,0 +1,211 @@
...
+website: https://github.com/cpuguy83/go-md2man
+```
+
+:::tip
+In metadata section, `packager`, `vendor` and `website` are optional fields.
+:::
+
+In the next section of the spec, we define the [sources](sources.md) that we will be pulling from. In this case, we are pulling from a git repository.
+
+One thing to note: in many build systems you will not have access to the Internet while building the package, and indeed that is the case with the `mariner2` target. As such, this build will fail because `go build` will try to download the go modules. For this reason, we added a `generate` section to the source to run `go mod download` in a docker image with the `src` source mounted and then extract the go modules from the resulting filesystem.
```
sozercan, line 63: can you add a tip for network connection flag

website/docs/targets.md
```diff
@@ -22,7 +22,7 @@ azlinux3/worker                  Builds the base worker image responsible for bu
 debug/gomods                     Outputs all the gomodule dependencies for the spec
 debug/resolve                    Outputs the resolved dalec spec file with build args applied.
 debug/sources                    Outputs all sources from a dalec spec file.
-mariner2/container (default)     Builds a container image for
+mariner2/container (default)     Builds a container image with mariner2 as the base image.
```
sozercan, line 25: which one is the default?

website/docs/container-only-builds.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Container-only builds
+---
+
+# Container-only builds
```
sozercan, line 5:
```suggestion
```

no need to add this again

website/docs/container-only-builds.md
adamperlin: Oh that is a really good point. I will use `docker buildx build` in the example here and then call that out.

website/docs/quickstart.md
```diff
@@ -0,0 +1,211 @@
...
```
cpuguy83, line 63: Add note that this is to ensure that the generated source package can be built without internet access and would be the default for all dalec targets.

website/docs/quickstart.md
```diff
@@ -0,0 +1,211 @@
...
+
+In the next section, we define the dependencies that are needed to build the package. In this case, we need the `golang` dependency at the build time, and `openssl-libs` at runtime. Build dependencies are dependencies that are needed to build the package, while runtime dependencies are dependencies that are needed to run the package, i.e., they will be installed alongside the package when it is installed on a system. Runtime dependencies are not required for this specific example, but they are included for illustrative purposes.
+
+```yaml
+dependencies:
+  build:
+    golang:
+  runtime:
+    # as stated above, included for illustrative purposes
+    openssl-libs:
```
cpuguy83, line 83: Maybe something more relevant like `man-db` so one can view generated manpages 😄

website/docs/quickstart.md
```diff
@@ -0,0 +1,211 @@
...
+
+```yaml
+build:
+  # the env section allows us to define environment variables 
+  # that will be set during the build process
+  env:
+    CGO_ENABLED: "0"
+  steps:
+    - command: |
+        cd src
```
cpuguy83, line 96: Maybe we can add a note as to where `src` comes from.

website/docs/targets.md
cpuguy83: > which one is the default?

Its default for `path.Dir(target)`.
So when `--target=mariner2`, the `mariner2/container` target is what gets executed.

website/docs/virtual-packages.md
```diff
@@ -76,8 +72,8 @@ website: http://contoso.com
 
 dependencies:
   runtime:
-    - my-package-foo
-    - my-package-bar
+    my-package-foo:
```
cpuguy83, line 75: Oh nice find!

website/docs/targets.md
adamperlin: Oh I'm realizing this is confusingly worded, it should be "default for distro" or "default for platform". I think I'm going to move the high-level targets `mariner2`, `azlinux3` and `windowscross` out as their items and mark them as aliases.

website/docs/quickstart.md
sozercan: +1 this has been confusing for me in the past for sure

website/docs/build.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Building with Dalec
+title: Building Packages and Containers with Dalec - An Overview
```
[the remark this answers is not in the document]
sozercan: @adamperlin did you push your changes, i am still seeing the same?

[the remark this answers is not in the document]
adamperlin: They are now pushed!

cpuguy83 (approved):

jeremyrickard (approved):

PR #441, opened 2024-11-23
Set image create

cpuguy83:
Set image config created time

Before this was just using whatever created time was set on the base
image config which was incorrect.

Example using the `docker buildx bake runc-jammy-container` command:

Before:

	docker image ls | grep runc
	runc                                          jammy             8d1d42e074f3   25 minutes ago   228MB

After:

	docker image ls | grep runc
	runc                                          jammy             03c24620c8fa   5 seconds ago    228MB

**Note**: how the created time before was 25mins ago (b/c apparently the base
jammy image was recently updated, I've seen this be 2 weeks) compared to
after its 5 seconds ago.

### DIFF
--- a/test/azlinux_test.go
+++ b/test/azlinux_test.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/Azure/dalec"
 	"github.com/Azure/dalec/frontend/azlinux"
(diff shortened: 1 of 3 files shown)


test/azlinux_test.go
```diff
@@ -528,7 +530,16 @@ echo "$BAR" > bar.txt
...
+			beforeBuild := time.Now()
+			res := solveT(ctx, t, gwc, sr)
+
+			dt, ok := res.Metadata[exptypes.ExporterImageConfigKey]
+			assert.Assert(t, ok, "result metadata should contain an image config: available metadata: %s", strings.Join(maps.Keys(res.Metadata), ", "))
+
+			var cfg dalec.DockerImageSpec
+			assert.Assert(t, json.Unmarshal(dt, &cfg))
+			assert.Check(t, cfg.Created.After(beforeBuild))
+			assert.Check(t, cfg.Created.Before(time.Now()))
```
DannyBrito, line 542: should we do a similar check in debian tests?

cpuguy83: Debian/Ubuntu is running these tests.

DannyBrito (approved):

PR #460, opened 2024-12-06
Fix issue mounting files

### LINKED ISSUE #459: [BUG] Mounting HTTP source into image cmd source causes error
### Expected Behavior

http source should not have any issues being mounted

### Actual Behavior

Given a source:

```yaml
http:
  url: http://example.com
```

The following occurs when placed into the "mounts" section of the image source's command struct.

```
Error: failed to solve: spec: msft-golang: error handling requested build target "bionic/deb": open /var/lib/containerd-stargz-grpc/snapshotter/snapshots/315975/fs/tmp/whatever: no such file or directory
```

### Steps To Reproduce

_No response_

cpuguy83:
Mounts were using the destination path as the source name when generating a source, this made it so mounts would only work correctly if mounted at the root of the filesystem since source names generally should not have path separators.

Fixes #459

### DIFF
--- a/frontend/test_runner.go
+++ b/frontend/test_runner.go
@@ -17,6 +17,19 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// This is used as the source name for sources in specified in `SourceMount`
+	// For any sources we need to mount we need to give the source a name.
+	// We don't actually care about the name here *except* the way file-backed
+	// sources work the name of the file becomes the source name.
+	// So we at least need to track it.
+	// Source names must also not contain path separators or it can screw up the logic.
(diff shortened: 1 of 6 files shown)


adamperlin (changes requested):
Does `frontend.RunTests` need to be updated with this fix as well since it also uses source mounts?

cpuguy83:
> Does frontend.RunTests need to be updated with this fix as well since it also uses source mounts?

Yep, I actually thought I did do that.

cpuguy83:
Good catch, fixed it.

adamperlin (approved):

PR #461, opened 2024-12-07
fix: Initialize nil maps in image config

cpuguy83:


imgconfig.go
```diff
@@ -37,10 +37,16 @@ func MergeSpecImage(spec *Spec, targetKey string) *ImageConfig {
 		cfg.Env = append(cfg.Env, i.Env...)
 
 		for k, v := range i.Volumes {
+			if cfg.Volumes == nil {
```
MorrisLaw, line 40: Would it be reasonable to make this a one-time check outside of the loop (same for Labels)?

cpuguy83: Done.

MorrisLaw (approved):

PR #465, opened 2024-12-09
feat: Add debugging setup for vscode

pmengelbert:
In order to use the debugger, you should Run any build as usual, but use substitute `debug/debug.sh` for `docker` in a docker invocation. The script uses `pgrep` and `socat`, as such they are required to run the script. Example:

```
./debug/debug.sh -f /tmp/moby-runc.yml --target=azlinux3/container --output=type=docker,name=hello:world .
```

Then, connect to the debugging setup using this `launch.json` config in vscode. You need the go plugin installed:

```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Connect to server",
            "type": "go",
            "request": "attach",
            "mode": "remote",
            "port": 30157,
            "host": "127.0.0.1",
            "apiVersion": 2
        }
    ]
}
```

Fixes #

### DIFF
--- a/Dockerfile.debug
+++ b/Dockerfile.debug
@@ -1,0 +1,37 @@
+FROM --platform=${BUILDPLATFORM} mcr.microsoft.com/oss/go/microsoft/golang:1.22@sha256:9e5243001d0f43d6e1c556a0ce8eedc1bc80eb37b5210036b289d2c601bc08b6 AS go
+FROM scratch AS dlv-ctx
+
+FROM go  AS frontend-build
+ARG HOSTDIR
+WORKDIR ${HOSTDIR}
+COPY . .
+ENV CGO_ENABLED=0
+ARG TARGETARCH TARGETOS GOFLAGS=-trimpath
+ENV GOOS=${TARGETOS} GOARCH=${TARGETARCH} GOFLAGS=${GOFLAGS}
+RUN \
+    --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    go build -tags=debug -gcflags="all=-N -l" -o /frontend ./cmd/frontend && \
+    go build -o /dalec-redirectio ./cmd/dalec-redirectio
+
+FROM go AS dlv-build
+ARG HOSTDIR
+WORKDIR /dlv
+ADD https://github.com/go-delve/delve.git#v1.23.1 .
+ENV CGO_ENABLED=0
+ARG TARGETARCH TARGETOS GOFLAGS=-trimpath
+ENV GOOS=${TARGETOS} GOARCH=${TARGETARCH} GOFLAGS=${GOFLAGS}
+RUN \
+    --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    make build
+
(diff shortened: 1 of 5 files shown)


Dockerfile.debug
```diff
@@ -0,0 +1,37 @@
+FROM --platform=${BUILDPLATFORM} mcr.microsoft.com/oss/go/microsoft/golang:1.22@sha256:9e5243001d0f43d6e1c556a0ce8eedc1bc80eb37b5210036b289d2c601bc08b6 AS go
```
cpuguy83, line 1: Can we unify the dockerfiles and pass the debug in as a build-arg?
`if [ "${DEBUG}" = "1" ]; then
    extra_flags=...
fi
go build ${extra_fiags} ...
```

File: debug/debug.sh [hunk 2]
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+
```
MorrisLaw, line 2: Should we add a `set -eux` here?

cmd/frontend/debug.go
```diff
@@ -0,0 +1,43 @@
...
+		return err
+	}
+
+	go func() {
+		if err := cmd.Wait(); err != nil {
+			panic(err)
+		}
+	}()
+
+	s, cancel := signal.NotifyContext(ctx, syscall.SIGCONT)
```
cpuguy83, line 38: I wonder if instead of using a signal here, if we can look at `/proc/self/status`: https://man7.org/linux/man-pages/man5/proc_pid_status.5.html

There should be a field called `TracerPid` which would be non-zero when being traced.

*IF* this works, there'd be no need to interact with the process at all other than to have dlv attach to it.

cpuguy83 (changes requested):

pmengelbert:
Another limitation here is that this probably won't work on Docker Desktop for Mac. it might work on DD for Windows if the Daemon is running in WSL. Not sure how to get around this, but it's a pretty big limitation.

cmd/frontend/debug.go
pmengelbert: This works!

Out of curiosity, how did you know this / find it out?

debug/debug.sh
pmengelbert: Sure. The polling for the pid is a bit noisy, so I'll set `+x` around there to keep it readable.

cmd/frontend/debug.go
```diff
@@ -0,0 +1,64 @@
...
+
+	go func() {
+		if err := cmd.Wait(); err != nil {
+			panic(err)
+		}
+	}()
+
+	tracerPid := "0"
+	for tracerPid == "0" {
+		b, err := os.Open("/proc/self/status")
```
MorrisLaw, line 42: Should we ensure the file gets closed? i.e. `defer b.Close()`

cpuguy83: Yes, but maybe we need to create a function for all this so we don't end up with a bunch of defers in a loop.

cpuguy83:
Closing since this is extremely old.

PR #466, opened 2024-12-09
apt: fix potential issue from prior build errors in cache

cpuguy83:
In some cases, if `apt update` had an error, the content `apt update`
was trying to fetch will be stored in `/var/lib/apt/lists/partial`.
In Dalec, `/var/lib/apt/lists` is a persistent cache.
As a note: `apt update` does not exit non-zero here, but `apt install`
might... depending on if the package is already in the local package
cache, which can make finding this issue pretty difficult.

For certain errors this can make builds unrecoverable since there's no
real way to clean up this cache outside of blowing away the entire build
cache.

This change just clears out anything in `/var/lib/apt/lists/partial`
before calling `apt update` since it should not be needed anyway and is
potentially problematic to have.

---

The issue here is sometimes a user can provide bad input (like gpg keys with the wrong permissions) and the user cannot recover from this on their own even if they fix the perms.

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -507,7 +507,7 @@ func repoConfigAsMount(config PackageRepositoryConfig, platformCfg *RepoPlatform
 
 	for name, repoConfig := range config.Config {
 		// each of these sources represent a repo config file
-		repoConfigSt, err := repoConfig.AsState(name, sOpt, append(opts, ProgressGroup("Importing repo config: "+name))...)
+		repoConfigSt, err := repoConfig.AsMount(name, sOpt, append(opts, ProgressGroup("Importing repo config: "+name))...)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


helpers.go
```diff
@@ -507,7 +507,7 @@ func repoConfigAsMount(config PackageRepositoryConfig, platformCfg *RepoPlatform
 
 	for name, repoConfig := range config.Config {
 		// each of these sources represent a repo config file
-		repoConfigSt, err := repoConfig.AsState(name, sOpt, append(opts, ProgressGroup("Importing repo config: "+name))...)
+		repoConfigSt, err := repoConfig.AsMount(name, sOpt, append(opts, ProgressGroup("Importing repo config: "+name))...)
```
MorrisLaw, line 510: what's the difference between `AsMount` and `AsState`?

cpuguy83: I mentioned this in the second commit.
This is a minor optimization that avoids an extra copy operation.

The difference between `AsState` and `AsMount` is that `AsMount` expects that you are going to mount the returned state somewhere and can just use `llb.SourcePath` in that mount call rather than copying the data from the subpath into the root of a new `llb.State`

Basically, when using `AsMount` you need to create a mount that points at the requested subpath.
When using `AsState`, the state itself reflects that subpath but can require an extra copy (depending on the source type).

MorrisLaw: Ah ok, I see it now:
```
This removes some extra copy operations required to use repo keys and
configs.
```

Ok yeah that makes sense as an optimization. Thanks for clarifying!

MorrisLaw (approved):

PR #480, opened 2024-12-18
Add support for outputing multiple windows images from one build

cpuguy83:
This allows builders to specify a file that contains a list of base images that the windowscross/container target should produce images for.
The result is a single multiplatform image: https://oci.dag.dev/?image=docker.io/cpuguy83/test@sha256:d88cf477b7f2464d4581f9490a8c1946300093461837272537e83879105f55e6&mt=application%2Fvnd.oci.image.index.v1%2Bjson&size=1693

### DIFF
--- a/frontend/build.go
+++ b/frontend/build.go
@@ -104,15 +104,7 @@ func fillPlatformArgs(prefix string, args map[string]string, platform ocispecs.P
 
 type PlatformBuildFunc func(ctx context.Context, client gwclient.Client, platform *ocispecs.Platform, spec *dalec.Spec, targetKey string) (gwclient.Reference, *dalec.DockerImageSpec, error)
 
-// BuildWithPlatform is a helper function to build a spec with a given platform
-// It takes care of looping through each tarrget platform and executing the build with the platform args substituted in the spec.
-// This also deals with the docker-style multi-platform output.
-func BuildWithPlatform(ctx context.Context, client gwclient.Client, f PlatformBuildFunc) (*gwclient.Result, error) {
-	dc, err := dockerui.NewClient(client)
-	if err != nil {
-		return nil, err
-	}
-
+func BuildWithPlatformFromUIClient(ctx context.Context, client gwclient.Client, dc *dockerui.Client, f PlatformBuildFunc) (*gwclient.Result, error) {
 	rb, err := dc.Build(ctx, func(ctx context.Context, platform *ocispecs.Platform, idx int) (gwclient.Reference, *dalec.DockerImageSpec, *dalec.DockerImageSpec, error) {
 		spec, err := LoadSpec(ctx, dc, platform)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


frontend/build.go
```diff
@@ -101,6 +125,17 @@ func BuildWithPlatform(ctx context.Context, client gwclient.Client, f PlatformBu
 	return rb.Finalize()
 }
 
+// BuildWithPlatform is a helper function to build a spec with a given platform
+// It takes care of looping through each tarrget platform and executing the build with the platform args substituted in the spec.
```
MorrisLaw, line 129: super nit:
```suggestion
// It takes care of looping through each target platform and executing the build with the platform args substituted in the spec.
```

frontend/windows/dockerui.go
```diff
@@ -0,0 +1,123 @@
...
+				}
+			}
+
+			p := platforms.DefaultSpec()
+			if tp != nil {
+				p = *tp
+			}
+
+			// in certain conditions we allow input platform to be extended from base image
+			if p.OS == "windows" && img.OS == p.OS {
```
MorrisLaw, line 69: The raw string for the OS is added here and also already exists in this file for the `windowsAmd64` initialization. Should we make this a constant?

frontend/windows/handle_container.go
```diff
@@ -33,81 +41,228 @@ var (
...
+		if err := json.Unmarshal(cfgs[idx], &baseCfg); err != nil {
+			return nil, nil, nil, errors.Wrap(err, "error unmarshalling base image config")
 		}
 
+		// Get a copy of the cfg so we can modify it
 		var img dalec.DockerImageSpec
-		if err := json.Unmarshal(dt, &img); err != nil {
-			return nil, nil, errors.Wrap(err, "error unmarshalling base image config")
+		if err := json.Unmarshal(cfgs[idx], &img); err != nil {
+			return nil, nil, nil, errors.Wrap(err, "error unmarshalling base image config")
```
MorrisLaw, line 251: Are this message and line 245 both true in that they're both `unmarshalling base image config`?

MorrisLaw (commented):
Initial pass through. looking good (coming from my somewhat limited perspective)

frontend/build.go
cpuguy83: Fixed.

cpuguy83:
> q: why you decided to go for refs to be build-arg instead of part of the spec?

Good question. Mostly because I wasn't sure how this would look and trying to decide if it is truly needed in the spec.
After some more thought I do have an idea of how to do this and will be a nice replacement for the `image.base`.

cpuguy83:
That said, I'd like to get the underlying functionality in first, I think, because adding support in the spec is going to be a much broader change.

DannyBrito (approved):

cpuguy83:
I have re-implemented this using just the spec here: https://github.com/Azure/dalec/pull/492

PR #481, opened 2024-12-20
fix: make sure fill defualts and expand args for all types 

### LINKED ISSUE #467: [BUG] apt: Repos keys must be readable by non-root user
### Expected Behavior

Specifying a repo key should not require explicitly setting permissions on the key file(s).

### Actual Behavior

Keys must have the correct permissions already set (or defined in the source type).

### Steps To Reproduce

```yaml
extra_repos:
  - keys:
      msft.asc:
        http:
          url: https://packages.microsoft.com/keys/microsoft.asc
```

Building a spec with the above `extra_repos` on any apt-based distro will error with an unrelated error due to `apt` not being able to read the key.

### LINKED ISSUE #479: [REQ] Some, typically nested, fields are not getting args resolved

None

In some cases args may not get resolved, such as in `extra_repos` or other places where source types are used (and potentially others).

cpuguy83:
Cherry-picks 8cc2df0f06e991e5078b843a80d6454ac0a15787 from #480 to avoid merge conflicts

Adds tests and fixes implementations where we weren't filling default
values and expanding args.
This was mostly a problem for anything under the `targets` section, but
also a few other fields where arg expansion is nice to have (such as in
test commands).

Also splits some types/functions off to separate files so we don't have one mega file.

Fixes #479
Fixes #467

### DIFF
--- a/frontend/mux.go
+++ b/frontend/mux.go
@@ -204,7 +204,14 @@ func (m *BuildMux) loadSpec(ctx context.Context, client gwclient.Client) (*dalec
 	}
 
 	// Note: this is not suitable for passing to builds since it does not have platform information
-	spec, err := LoadSpec(ctx, dc, nil)
+	spec, err := LoadSpec(ctx, dc, nil, func(cfg *LoadConfig) {
+		cfg.SubstituteOpts = append(cfg.SubstituteOpts, func(cfg *dalec.SubstituteConfig) {
+			// Allow any args here since we aren't trying to validate the spec at this point.
+			cfg.AllowArg = func(string) bool {
+				return true
+			}
+		})
+	})
 	if err != nil {
 		return nil, err
 	}
--- a/frontend/mux.go
+++ b/frontend/mux.go
@@ -205,12 +205,9 @@ func (m *BuildMux) loadSpec(ctx context.Context, client gwclient.Client) (*dalec
 
 	// Note: this is not suitable for passing to builds since it does not have platform information
 	spec, err := LoadSpec(ctx, dc, nil, func(cfg *LoadConfig) {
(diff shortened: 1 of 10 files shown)


frontend/mux.go
```diff
@@ -204,7 +204,14 @@ func (m *BuildMux) loadSpec(ctx context.Context, client gwclient.Client) (*dalec
 	}
 
 	// Note: this is not suitable for passing to builds since it does not have platform information
-	spec, err := LoadSpec(ctx, dc, nil)
+	spec, err := LoadSpec(ctx, dc, nil, func(cfg *LoadConfig) {
+		cfg.SubstituteOpts = append(cfg.SubstituteOpts, func(cfg *dalec.SubstituteConfig) {
+			// Allow any args here since we aren't trying to validate the spec at this point.
+			cfg.AllowArg = func(string) bool {
```
adamperlin, line 210: This can be a change for later, but it might be nice to make this a global like `AllowAny`

load.go
```diff
@@ -76,186 +74,30 @@ func expandArgs(lex *shell.Lex, s string, args map[string]string) (string, error
...
-		count++
+	for _, o := range opts {
+		o(&cfg)
 	}
 
-	if s.Inline != nil {
-		if err := s.Inline.validate(s.Path); err != nil {
-			retErr = goerrors.Join(retErr, err)
+	if cfg.AllowArg == nil {
+		cfg.AllowArg = func(string) bool {
```
adamperlin, line 96: Similarly here, might be nice to have something like
```go 
var DisallowAllArgs = func(string) bool { return false }
```

frontend/mux.go
cpuguy83: I went ahead an implemented this.

adamperlin (approved):
This looks good to me! I think moving different spec sections to different files makes it much clearer which spots are getting defaults filled/args substituted and helps break out the logic nicely.

PR #482, opened 2024-12-20
fix: test2json: detect test timeouts and other package-level errors

### LINKED ISSUE #476: [BUG] Integration tests possible not reporting failures

cpuguy83:
This was ignoring all non-test events, unfortunately this means if there is a panic or a test timeout we may not detect the failure.

Fixes #476

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
       - name: download deps
         run: go mod download
       - name: Run integration tests
-        run: go test -v -json ./test | go run ./cmd/test2json2gha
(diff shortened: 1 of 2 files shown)


cpuguy83:
Looking into what's taking the packages so long.
This timed after at 30 minutes, that is very much unexpected.

cpuguy83:
Superseded by #421

PR #491, opened 2025-01-06
Allow build args in target image config

### LINKED ISSUE #487: [REQ] image.labels should support arg substitution

None

_No response_

### LINKED ISSUE #490: [BUG] Build args are not substituted into `image.base`
### Expected Behavior

```yaml
args:
  BUSYBOX_TAG: 1.36
image:
  base: busybox:${BUSYBOX_TAG}
```

### Actual Behavior

With the above:
```
error handling requested build target "azlinux3/container": could not resolve base image config: error resolving base image config: docker.io/library/busybox:${BUSYBOX_VERSION}: not found
```

### Steps To Reproduce

_No response_

pmengelbert:

Fixes #490
Fixes #487

### DIFF
--- a/load.go
+++ b/load.go
@@ -199,6 +199,10 @@ func (s *Spec) SubstituteArgs(env map[string]string, opts ...SubstituteOpt) erro
 		}
 	}
 
+	if err := s.Image.processBuildArgs(lex, args, cfg.AllowArg); err != nil {
+		appendErr(errors.Wrap(err, "package config"))
+	}
+
 	if err := s.Dependencies.processBuildArgs(args, cfg.AllowArg); err != nil {
 		appendErr(errors.Wrap(err, "dependencies"))
 	}
@@ -473,7 +477,7 @@ func validateSymlinks(symlinks map[string]SymlinkTarget) error {
 			errs = append(errs, err)
 		}
 
-		if (cfg.Path != "" && len(cfg.Paths) > 0) || (cfg.Path == "" && len(cfg.Paths) == 0 ) {
+		if (cfg.Path != "" && len(cfg.Paths) > 0) || (cfg.Path == "" && len(cfg.Paths) == 0) {
 			err = fmt.Errorf("'path' and 'paths' fields are mutually exclusive, and at least one is required: "+
 				"symlink to %s", oldpath)
 
@@ -535,4 +539,23 @@ func validateSymlinks(symlinks map[string]SymlinkTarget) error {
 	}
 
 	return goerrors.Join(errs...)
+}
+
+func (img *ImageConfig) processBuildArgs(lex *shell.Lex, args map[string]string, allowArg func(string) bool) error {
+	if img == nil {
+		return nil
+	}
+
+	var errs error
+
+	for k, v := range img.Labels {
+		updated, err := expandArgs(lex, v, args, allowArg)
+		if err != nil {
+			errs = goerrors.Join(errs, errors.Wrapf(err, "env %s=%s", k, v))
+			continue
+		}
+		img.Labels[k] = updated
+	}
+
+	return errs
 }
--- a/load_test.go
+++ b/load_test.go
@@ -590,6 +590,14 @@ func TestSpec_SubstituteBuildArgs(t *testing.T) {
 					Args: maps.Clone(pairs),
 				},
 			},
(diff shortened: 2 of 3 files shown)


load.go
```diff
@@ -447,3 +453,57 @@ func (b *ArtifactBuild) processBuildArgs(lex *shell.Lex, args map[string]string,
...
+	for k, v := range img.Labels {
+		updated, err := expandArgs(lex, v, args, allowArg)
+		if err != nil {
+			errs = goerrors.Join(errs, errors.Wrapf(err, "env %s=%s", k, v))
+			continue
+		}
+		img.Labels[k] = updated
+	}
+
+	if img.Post != nil {
```
cpuguy83, line 485: I don't see why this would be necessary and would prefer to not support it.

load.go
```diff
@@ -447,3 +453,57 @@ func (b *ArtifactBuild) processBuildArgs(lex *shell.Lex, args map[string]string,
...
+	var errs error
+	for _, p := range []*string{&img.Base, &img.Cmd, &img.User, &img.Entrypoint, &img.StopSignal, &img.WorkingDir} {
+		updated, err := expandArgs(lex, *p, args, allowArg)
+		if err != nil {
+			errs = goerrors.Join(errs, errors.Wrap(err, "imgconfig"))
+		}
+		*p = updated
+	}
+
+	for i, s := range img.Env {
```
cpuguy83, line 467: I'm a little hesitant support this here but not against it.

cpuguy83 (changes requested):

load.go
```diff
@@ -447,3 +453,57 @@ func (b *ArtifactBuild) processBuildArgs(lex *shell.Lex, args map[string]string,
 
 	return goerrors.Join(errs...)
 }
+
+func (img *ImageConfig) processBuildArgs(lex *shell.Lex, args map[string]string, allowArg func(string) bool) error {
+	var errs error
+	for _, p := range []*string{&img.Base, &img.Cmd, &img.User, &img.Entrypoint, &img.StopSignal, &img.WorkingDir} {
```
cpuguy83, line 459: Likewise, I don't see why image base would need build args.

load.go
```diff
@@ -447,3 +453,57 @@ func (b *ArtifactBuild) processBuildArgs(lex *shell.Lex, args map[string]string,
 
 	return goerrors.Join(errs...)
 }
+
+func (img *ImageConfig) processBuildArgs(lex *shell.Lex, args map[string]string, allowArg func(string) bool) error {
```
MorrisLaw, line 457: Labels seems like a good candidate for build args. But why the other fields? Why not just limit support to Labels for now until there's a compelling use case/need for additional build arg support?

**Note:** I could be missing context.

load.go
pmengelbert: At the very least I think it would be useful for the tag. Some repos do not have a `latest` tag, and not allowing a build arg here forces a hard-coded tag. I'm not married to the idea of introducing this, but it does seem useful.

Other uses could be not hard-coding a registry.

load.go
pmengelbert: Understandable. Not strictly necessary, just a little more convenient for certain things. If you don't agree with the rationale feel free to close

load.go
pmengelbert: :+1:

load.go
pmengelbert: See above, I think it could be useful for setting base image tags or registry host, particularly when the base image provider doesn't have a `latest` tag.

load.go
cpuguy83: Not hard-coding a registry could be taken care of by source policies.

pmengelbert: What do you think about tags?

cpuguy83: I think I'd opt for explicitness in the spec in this case until someone actually asks for this.

pmengelbert:
> @pmengelbert any update here?

I'm going to update the PR.

load.go
cpuguy83: Also base image is now deprecated in favor of the new `bases`.

pmengelbert:
@cpuguy83 @MorrisLaw Now the args are only substituted in the `Labels` field of the image config. Should be ready to go

cpuguy83 (approved):

PR #493, opened 2025-01-08
Refactor RPM handlers and Reorganize Code for Linux targets

adamperlin:
This PR refactors the handlers for RPM targets to be methods on a distro config, almost identical to how the DEB targets work. For example, installing build dependencies, building and testing a package, installing in a container, and the like should not vary much between rpm-based distros, so the functionality has been made into methods that encode the functionality for all rpm distros.

As part of this change, the project structure was altered to nest functionality common to RPM targets under `frontend/linux/rpm`, and functionality common to `deb` targets has been moved to `frontend/linux/deb`. A `DistroConfig` interface common to both is under `frontend/linux`. The `DistroConfig` interface is meant to encode what kinds of functionality any new type of distro would need to provide, such that the generic handlers `HandleContainer(DistroConfig)` and `HandlePackage(DistroConfig)` can be used for `distro/container` and `distro/<package>` type targets. 

Many of the changed files in this PR are simply renames or changes of a single import, so the number of substantively changed files is much less than the diff shows.


frontend/linux/distro_handler.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/moby/buildkit/client/llb"
+	"github.com/moby/buildkit/client/llb/sourceresolver"
+	"github.com/pkg/errors"
+
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type DistroConfig interface {
```
adamperlin, line 20: The idea here is to have an interface that would work across distros, even beyond rpm/deb based packaging. Right now though, it just conforms to what is needed for rpm and deb. There are a few things that don't fit perfectly here, in particular the need for an extra `worker` parameter for building an rpm container when it isn't needed for deb. The signatures are similar enough though that it seemed like an interface would be reasonable, but I'm definitely open to rethinking this if it's too much of a stretch.

frontend/linux/rpm/buildroot.go
```diff
@@ -0,0 +1,49 @@
+package rpm
```
adamperlin, line 1: This code hasn't changed beyond method renames

frontend/linux/rpm/distro/distro.go
```diff
@@ -0,0 +1,83 @@
...
+	var cfg dnfInstallConfig
+	dnfInstallOptions(&cfg, opts)
+
+	return dalec.WithRunOptions(c.InstallFunc(&cfg, c.ReleaseVer, pkgs), c.PackageCacheMount(cfg.root))
+}
+
+func (cfg *Config) Handle(ctx context.Context, client gwclient.Client) (*gwclient.Result, error) {
+	var mux frontend.BuildMux
+
+	mux.Add("rpm", linux.HandlePackage(cfg), &targets.Target{
```
adamperlin, line 56: We can use the root linux package target handler for this, as well as the container

frontend/linux/rpm/distro/dnf_install.go
```diff
@@ -0,0 +1,263 @@
...
+
+var dnfRepoPlatform = dalec.RepoPlatformConfig{
+	ConfigRoot: "/etc/yum.repos.d",
+	GPGKeyRoot: "/etc/pki/rpm-gpg",
+	ConfigExt:  ".repo",
+}
+
+type PackageInstaller func(*dnfInstallConfig, string, []string) llb.RunOption
+
+type dnfInstallConfig struct {
```
adamperlin, line 23: We only have a dnf install config and not a tdnfInstallConfig, since in Dalec thus far every option we use is both dnf and tdnf compatible. In the future as we add more distros, we may need to rethink this portion, but if we stick to only using the options that are the same between the two, then it makes our life easier!

frontend/linux/rpm/distro/dnf_install.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	runOpts = append(runOpts, dalec.ShArgs(cmdArgs))
+	runOpts = append(runOpts, cfg.mounts...)
+
+	return dalec.WithRunOptions(runOpts...)
+
+}
+
+func DnfInstall(cfg *dnfInstallConfig, releaseVer string, pkgs []string) llb.RunOption {
```
adamperlin, line 163: I've split out the install methods maybe prematurely, in case there are differences between dnf and tdnf. In particular, I'm not sure if the separate `/import-keys.sh` script is needed for dnf yet, since we haven't yet added distros that use dnf.

frontend/linux/distro_handler.go
cpuguy83: I'll say the large interface makes me a little hesitant but it doesn't seem too bad here.
If we need to add something that can't conform to it then it'll just need to handle some of these higher level functions on their own.

Basically, SGTM.

cpuguy83 (commented):
WDYT about moving the rpm/deb stuff into a top-level package like `packaging/rpm` and `packaging/deb` or even `packaging/linux/rpm` and `packaging/linux/deb`?

Likewise I think we can move all the targets stuff into a top-level package called `targets`.

adamperlin:
@cpuguy83 I am definitely open to that! I think splitting the targets and the packaging pieces out makes a lot of sense!

adamperlin:
This should be ready for another look @cpuguy83 @MorrisLaw!

cpuguy83 (approved):
This LGTM

I do have some nits about the package structure, but not a blocker.

For the `targets` package, I feel like each of the distro subpackages are uneccessary.
Instead we can have something like `targets.Bookworm` which would be the distro config for bookworm, as an example.

I also think the `linux` and `windows` subpackages probably aren't neccessary.
LIkewise in the `packaging` package.

But overall, this is a nice change.
Thank you!

adamperlin:
That makes sense! I can possibly do a follow up to remove some of those subpackages!

PR #494, opened 2025-01-08
feat: Add llb symlinks [don't merge until buildkit v0.19.x is released]

pmengelbert:


pmengelbert:
- [ ] Needs a regression test

Issue #489, opened 2025-01-06
[BUG] Multiple symlinks cannot be made to point to the same destination


pmengelbert:
### Expected Behavior

You should be able to create multiple symlinks that point to the same location. For example:

```yaml
image:
  post:
    symlinks:
      - oldpath: /bin/busybox
        newpath: /usr/bin/sh
      - oldpath: /bin/busybox
         newpath: /usr/bin/test
```

### Actual Behavior

The spec format requires:

```yaml
image:
  post:
    symlinks:
      /bin/busybox:
        path: /usr/bin/sh
      /bin/busybox:
         path: /usr/bin/test
```

This causes a yaml parsing error at best, and silently overwrites the `/bin/busybox` map key at worst. 

### Steps To Reproduce

n/a, see above.

### Are you willing to submit PRs to contribute to this bug fix?

- [x] Yes, I am willing to implement it.

cpuguy83:
Note, we already have a type for this in the artifact section: https://github.com/Azure/dalec/blob/08d2e6c5336862ce4cadf4e32c346ca04f40d4db/artifacts.go#L49-L57

The case for symlinks in the image-post is a slightly different one though (really a hack so we don't need to create multiple packages).

cpuguy83:
I think we could change https://github.com/Azure/dalec/blob/08d2e6c5336862ce4cadf4e32c346ca04f40d4db/spec.go#L131-L134 to include a `paths` field that would be muttually exclusive with `path` (so specify one or the other, not both) which is a `[]string`

pmengelbert:
> I think we could change
> 
> [dalec/spec.go](https://github.com/Azure/dalec/blob/08d2e6c5336862ce4cadf4e32c346ca04f40d4db/spec.go#L131-L134)
> 
> Lines 131 to 134 in [08d2e6c](/Azure/dalec/commit/08d2e6c5336862ce4cadf4e32c346ca04f40d4db)
> 
>  type SymlinkTarget struct { 
>  	// Path is the path where the symlink should be placed 
>  	Path string `yaml:"path" json:"path" jsonschema:"required"` 
>  } 
> to include a `paths` field that would be muttually exclusive with `path` (so specify one or the other, not both) which is a `[]string`

Sounds like a good approach. I will implement this prior to #494.

pmengelbert:
Fixed by #495

PR #492, opened 2025-01-07
Add support for specifying multiple base images to use in the spec

cpuguy83:
Allows windowscross to produce a single multiplatform image using the specified base images.

Supersedes #480 
Closes #480

PR #495, opened 2025-01-09
feat: Multiple postinstall symlinks pointing to the same thing

pmengelbert:
Currently, you can't create multiple symlink `newpath`s that point to the same `oldpath`. This is due to the fact that map keys have to be unique.

For more information, see #489

This should be merged before #494


spec.go
```diff
@@ -130,7 +130,8 @@ type PostInstall struct {
 // SymlinkTarget specifies the properties of a symlink
 type SymlinkTarget struct {
 	// Path is the path where the symlink should be placed
-	Path string `yaml:"path" json:"path" jsonschema:"required"`
+	Path  string   `yaml:"path" json:"path" jsonschema:"oneof_required=path"`
```
cpuguy83, line 133: We should mark this is deprecated and have people use `paths` instead`.
We can also make it so when we load the spec we migrate `Path` to `Paths` then there's only one field to deal with.

target.go
```diff
@@ -42,6 +42,12 @@ func (t *Target) validate() error {
 		}
 	}
 
+	if t.Image != nil {
```
cpuguy83, line 45: Let's add a `validate` to `t.Image`, which itself could `return nil` if `*Image` is `nil` (note: this is what I did in #492

cpuguy83 (changes requested):

spec.go
pmengelbert: How do we mark it as deprecated?

cpuguy83: https://go.dev/wiki/Deprecated

target.go
pmengelbert: done

spec.go
pmengelbert: done, but I'm going to add tests for it.

pmengelbert: tests added.

pmengelbert:
@cpuguy83 This should be ready for another round of review.

load.go
```diff
@@ -447,3 +455,75 @@ func (b *ArtifactBuild) processBuildArgs(lex *shell.Lex, args map[string]string,
...
 }
+
+func (i *ImageConfig) validate() error {
+	if i == nil {
+		return nil
+	}
+
+	var errs []error
+	if err := i.Post.validate(); err != nil {
+		errs = append(errs, errors.Wrap(err, "postinstall"))
```
cpuguy83, line 466: Nit, just "post"

load.go
```diff
@@ -447,3 +455,75 @@ func (b *ArtifactBuild) processBuildArgs(lex *shell.Lex, args map[string]string,
...
+			errs = append(errs, err)
+		}
+
+		if err != nil {
+			continue
+		}
+
+		// By this point, both the oldpath and the SymlinkTarget are
+		// well-formed. We still need to make sure each 'newpath' is unique.
+		if cfg.Paths == nil {
```
cpuguy83, line 508: Validations should not be modifying anything.
We can also iterate on `Paths` without having to have it be non-nil.

load.go
```diff
@@ -447,3 +455,75 @@ func (b *ArtifactBuild) processBuildArgs(lex *shell.Lex, args map[string]string,
...
+			cfg.Paths = []string{}
+		}
+
+		newpaths := slices.Clone(cfg.Paths)
```
cpuguy83, line 512: Why the clone? Should not need to modify anything here.
If we want to make sure that `cfg.Path` is not also somewhere in `cfg.Paths` you can check that directly.

```go
for _, link := range symlinks {
  if link.Path == cfg.Path {
    // error
  }
}
```

load.go
```diff
@@ -447,3 +455,75 @@ func (b *ArtifactBuild) processBuildArgs(lex *shell.Lex, args map[string]string,
...
+	if err := validateSymlinks(p.Symlinks); err != nil {
+		errs = append(errs, err)
+	}
+
+	return errors.Wrap(goerrors.Join(errs...), "symlink")
+}
+
+func validateSymlinks(symlinks map[string]SymlinkTarget) error {
+	var errs []error
+	dests := make(map[string]string, len(symlinks)<<1)
```
cpuguy83, line 488: Maybe lets call this `seen` and only allocate the map if its needed.

cpuguy83 (changes requested):

load.go
pmengelbert: I could have sworn I put this in `fillDefaults`. I need to move this code there. The clone from moving the old code around. I'll take another pass

pmengelbert: > Why the clone? Should not need to modify anything here. If we want to make sure that `cfg.Path` is not also somewhere in `cfg.Paths` you can check that directly.
> 
> ```go
> for _, link := range symlinks {
>   if link.Path == cfg.Path {
>     // error
>   }
> }
> ```

And the goal is to make sure the there are no duplicate `oldpath`s in **the whole `map[string]SymlinkTarget`**. The desired behavior is unclear (follow the symlink that has already been created? Or overwrite the existing symlink?) Map iteration is not reliably ordered (and we sort the keys to avoid cache misses), so the user is unlikely to get what they want anyway. Better to just not allow it.

load.go
pmengelbert: fixed

load.go
pmengelbert: done.

load.go
pmengelbert: done. This exhibits the desired behavior, doesn't modify the spec, is tested properly, and only allocates the map when needed.

frontend/windows/handle_container.go
```diff
@@ -116,21 +116,19 @@ func copySymlinks(post *dalec.PostInstall) llb.StateOption {
...
 			return s
 		}
-		keys := dalec.SortMapKeys(lm)
-		for _, srcPath := range keys {
-			l := lm[srcPath]
-			dstPath := l.Path
-			s = s.File(llb.Mkdir(path.Dir(dstPath), 0755, llb.WithParents(true)))
-			s = s.File(llb.Copy(s, srcPath, dstPath))
+
+		for oldpath, newpaths := range post.Symlinks {
```
cpuguy83, line 123: I think we still need to sort this.

helpers.go
```diff
@@ -383,9 +383,11 @@ func InstallPostSymlinks(post *PostInstall, rootfsPath string) llb.RunOption {
 		buf := bytes.NewBuffer(nil)
 		buf.WriteString("set -ex\n")
 
-		for src, tgt := range post.Symlinks {
-			fmt.Fprintf(buf, "mkdir -p %q\n", filepath.Join(rootfsPath, filepath.Dir(tgt.Path)))
-			fmt.Fprintf(buf, "ln -s %q %q\n", src, filepath.Join(rootfsPath, tgt.Path))
+		for oldpath, newpaths := range post.Symlinks {
```
cpuguy83, line 386: Should also sort this

load_test.go
```diff
@@ -1073,3 +1074,359 @@ func Test_validatePatch(t *testing.T) {
...
+				},
+			},
+		},
+		{
+			desc: "Path should be appended to Paths",
+			input: ImageConfig{
+				Post: &PostInstall{
+					Symlinks: map[string]SymlinkTarget{
+						"oldpath": {
+							Path:  "/newpath3",
```
cpuguy83, line 1307: I don't think we need to test this since its not a valid case.

target.go
```diff
@@ -68,4 +73,30 @@ func (t *Target) processBuildArgs(lex *shell.Lex, args map[string]string, allowA
...
+	}
+
+	for oldpath := range p.Symlinks {
+		cfg := p.Symlinks[oldpath]
+		if cfg.Path != "" {
+			cfg.Paths = append(cfg.Paths, cfg.Path)
+			cfg.Path = ""
+		}
+
+		sort.Strings(cfg.Paths)
```
cpuguy83, line 99: I don't think this should be sorted since this should be left to the user.

cpuguy83 (changes requested):

load_test.go
pmengelbert: Good catch. Fixed it so the test code will reject invalid tests.

target.go
pmengelbert: done.

frontend/windows/handle_container.go
pmengelbert: done

helpers.go
pmengelbert: done

pmengelbert:
This has been rebased and is ready for another round of review @cpuguy83

image.go
```diff
@@ -199,10 +207,43 @@ func (s *BaseImage) validate() error {
...
+	// validation has already taken place
+	for oldpath := range p.Symlinks {
+		cfg := p.Symlinks[oldpath]
+		if cfg.Path == "" {
+			continue
+		}
+
+		cfg.Paths = append(cfg.Paths, cfg.Path)
+		cfg.Path = ""
+		sort.Strings(cfg.Paths)
```
cpuguy83, line 242: This looks like an unnecessary sort since it should only have one entry.
Also just to note, generally don't want to sort user provided arrays.

load.go
```diff
@@ -452,3 +459,80 @@ func (b *ArtifactBuild) processBuildArgs(lex *shell.Lex, args map[string]string,
...
+	)
+
+	for oldpath, cfg := range symlinks {
+		var err error
+		if oldpath == "" {
+			err = fmt.Errorf("symlink source is empty")
+			errs = append(errs, err)
+		}
+
+		if cfg.Path != "" && len(cfg.Paths) != 0 || cfg.Path == "" && len(cfg.Paths) == 0 {
```
cpuguy83, line 476:
```suggestion
		if (cfg.Path != "" && len(cfg.Paths) > 0) || (cfg.Path == "" && len(cfg.Paths) == 0 ) {
```

cpuguy83 (changes requested):

image.go
cpuguy83:
```suggestion
```

image.go
```diff
@@ -3,6 +3,7 @@ package dalec
 import (
 	"context"
 	goerrors "errors"
+	"sort"
```
cpuguy83, line 6:
```suggestion
```

cpuguy83 (approved):

PR #502, opened 2025-01-23
Allow for target-defined artifacts

### LINKED ISSUE #474: [REQ] Allow artifacts to be defined per-target

None

For the most part Dalec supports specifying multiple build targets in one spec, however sometimes different targets need different artifact configurations, especially if you have windows and linux in one spec.

Today you cannot add per-target artifact configs but this is becoming more of a need.

MorrisLaw:
This PR allows for target-defined artifacts which is especially useful for windows builds.

Fixes #474

### DIFF
--- a/frontend/rpm/template.go
+++ b/frontend/rpm/template.go
@@ -437,10 +437,6 @@ func (w *specWrapper) Install() fmt.Stringer {
 	fmt.Fprintln(b, "%install")
 
 	artifacts := w.GetArtifacts(w.Target)
-	if artifacts.IsEmpty() {
-		b.WriteString("\n")
-		return b
-	}
 
 	copyArtifact := func(root, p string, cfg *dalec.ArtifactConfig) {
 		if cfg == nil {
@@ -576,10 +572,6 @@ func (w *specWrapper) Files() fmt.Stringer {
 	fmt.Fprintf(b, "%%files\n")
 
 	artifacts := w.GetArtifacts(w.Target)
-	if artifacts.IsEmpty() {
-		b.WriteString("\n")
-		return b
-	}
 
 	if len(artifacts.Binaries) > 0 {
 		binKeys := dalec.SortMapKeys(artifacts.Binaries)
--- a/frontend/rpm/template.go
+++ b/frontend/rpm/template.go
@@ -140,18 +140,16 @@ OrderWithRequires(postun): systemd
 func (w *specWrapper) Requires() fmt.Stringer {
 	b := &strings.Builder{}
 
+	artifacts := w.GetArtifacts(w.Target)
+
 	// first write systemd requires if they exist,
 	// as these do not come from dependencies in the spec
-	b.WriteString(getSystemdRequires(w.Artifacts.Systemd))
+	b.WriteString(getSystemdRequires(artifacts.Systemd))
 
-	deps := w.Spec.Targets[w.Target].Dependencies
-	if deps == nil {
-		deps = w.Spec.Dependencies
-	}
+	deps := w.GetPackageDeps(w.Target)
 	if deps == nil {
 		return b
 	}
-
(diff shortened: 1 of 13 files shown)


cpuguy83:
For simplicity, do you think we can deal with refactoring things in a separate PR?

MorrisLaw:
> For simplicity, do you think we can deal with refactoring things in a separate PR?

Yeah, that's reasonable.

frontend/rpm/template.go
```diff
@@ -140,18 +140,18 @@ OrderWithRequires(postun): systemd
 func (w *specWrapper) Requires() fmt.Stringer {
 	b := &strings.Builder{}
 
+	artifacts := w.GetArtifacts(w.Target)
+	if artifacts == nil || artifacts.IsEmpty() {
+		return b
+	}
```
cpuguy83: I don't think we can return early here since this is where dependencies get written.

```suggestion
```

helpers.go
```diff
@@ -357,6 +357,13 @@ func (s *Spec) GetImagePost(target string) *PostInstall {
 	return nil
 }
 
+func (s *Spec) GetArtifacts(targetKey string) *Artifacts {
```
cpuguy83: Can we always use a value object?
This way we don't need to deal with a nil everywhere and code can continue to work the way it was before (mostly).

```suggestion
func (s *Spec) GetArtifacts(targetKey string) Artifacts {
```

cpuguy83: Also looks like `Artifacts.IsEmpty` could use a nil check anyway since a nil artifacts should be considered empty.
This is causing a panic currently in the debian side of things.

Though returning a zero value of `Artifacts` here instead of nil should take care of the panic, I think it makes sense to add the nil check directly to `Artifacts.IsEmpty` as well.

cpuguy83 (changes requested):

frontend/rpm/template.go
MorrisLaw: good catch, this was not supposed to be added here

target.go
```diff
@@ -28,6 +28,9 @@ type Target struct {
 	// PackageConfig is the configuration to use for artifact targets, such as
 	// rpms, debs, or zip files containing Windows binaries
 	PackageConfig *PackageConfig `yaml:"package_config,omitempty" json:"package_config,omitempty"`
+
+	// Artifacts describes all of the artifact configurations to include for this specific target.
+	Artifacts Artifacts `yaml:"artifacts,omitempty" json:"artifacts,omitempty"`
```
cpuguy83, line 33: I think this still needs to be a pointer so we know when its empty vs not set.

cpuguy83: Also `IsEmpty` does not check all cases after https://github.com/Azure/dalec/pull/514 which can complicate things.

cpuguy83 (changes requested):

helpers.go
```diff
@@ -357,6 +357,15 @@ func (s *Spec) GetImagePost(target string) *PostInstall {
 	return nil
 }
 
+func (s *Spec) GetArtifacts(targetKey string) Artifacts {
+	if t, ok := s.Targets[targetKey]; ok {
+		if t.Artifacts != nil && !t.Artifacts.IsEmpty() {
```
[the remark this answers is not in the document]
MorrisLaw: When you say an empty artifact is a deliberate choice, does that mean if an empty artifacts is present it should take precedent over spec level artifacts that do have something?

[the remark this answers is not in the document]
cpuguy83: Correct.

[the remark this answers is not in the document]
cpuguy83: So if there is any non-nil value, it should take precedence.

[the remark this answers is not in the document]
MorrisLaw: Updated

cpuguy83 (approved):

PR #504, opened 2025-01-24
Add recover for panics from handlers

### LINKED ISSUE #181: [REQ] Capture frontend panics and propagate messages back to the client

None

Currently, when the frontend panics, there is no error message aside from an `exit status 2`. It would be nicer to be able to see what caused the panic (and ideally, where).

cpuguy83:
This allows the error to actually propagate back to the client instead of the ominous "exit code 2" and having to search through the daemon logs to see what happened.

Closes #181
Closes #505

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -16,10 +16,6 @@ variable "DALEC_DISABLE_DIFF_MERGE" {
     default = "0"
 }
 
-variable "DALEC_NO_CACHE_EXPORT" {
-    default = "0"
-}
-
 target "frontend" {
     target = "frontend"
     tags = [FRONTEND_REF]
@@ -76,9 +72,6 @@ target "runc-azlinux" {
     tags = tgt == "container" ? ["runc:mariner2"] : []
     // only output non-container targets to the fs
     output = tgt != "container" ? ["_output"] : []
-
-    cache-from = ["type=gha,scope=dalec/runc/mariner2/${tgt}"]
-    cache-to = DALEC_NO_CACHE_EXPORT != "1" ? ["type=gha,scope=dalec/runc/mariner2/${tgt},mode=max"] : []
 }
 
 target "runc-jammy" {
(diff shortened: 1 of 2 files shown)


docker-bake.hcl
```diff
@@ -142,9 +141,12 @@ target "test-fixture" {
     dockerfile = "test/fixtures/${f}.yml"
 
     args = {
-        "BUILDKIT_SYNTAX" = FRONTEND_REF
+        "BUILDKIT_SYNTAX" = "dalec_frontend"
         "DALEC_DISABLE_DIFF_MERGE" = DALEC_DISABLE_DIFF_MERGE
     }
+    contexts = {
```
adamperlin, line 137: Does this mean `target:frontend` needs to be a build context?

cpuguy83: This meas the bake target *is* passing the frontend as a build context and then the build-arg tells it to read the frontend from the context.

adamperlin:
It looks like the first two commits will be part of #505 - does that PR also need to be open still? Otherwise looks good to me!

adamperlin (approved):
Very happy to have this, I've certainly experienced the `error exit code 2`. Does this make #505 unnecessary?

docker-bake.hcl
```diff
@@ -23,10 +16,6 @@ variable "DALEC_DISABLE_DIFF_MERGE" {
     default = "0"
 }
 
-variable "DALEC_NO_CACHE_EXPORT" {
-    default = "0"
-}
-
 target "frontend" {
```
adamperlin, line 19: Moving this over from the other PR, but could you add a comment here to clarify that this will build the frontend image implicitly because it will use the Dockerfile in the current dir?

cpuguy83:
> It looks like the first two commits are in #505 - does that PR also need to be open still? I just had one comment on that PR that I duplicated here.

Yeah, all the recent PR's have those commits in them so that CI can pass since buildx was updated on the build agents and there's a change (likely HCL parser update) that broke our bake file.

PR #505, opened 2025-01-24
bake file: Directly inject frontend to bake targets

cpuguy83:
This makes it so we no longer need to manually build the frontend image. It also means we don't need to use dockerd as the builder for these targets to work.

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -16,10 +16,6 @@ variable "DALEC_DISABLE_DIFF_MERGE" {
     default = "0"
 }
 
-variable "DALEC_NO_CACHE_EXPORT" {
-    default = "0"
-}
-
 target "frontend" {
     target = "frontend"
     tags = [FRONTEND_REF]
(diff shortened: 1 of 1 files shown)


docker-bake.hcl
```diff
@@ -23,10 +16,6 @@ variable "DALEC_DISABLE_DIFF_MERGE" {
     default = "0"
 }
 
-variable "DALEC_NO_CACHE_EXPORT" {
-    default = "0"
-}
-
 target "frontend" {
```
adamperlin, line 19: Can you add a comment here clarifying that this will build the root Dockerfile (that builds the frontend image) since no other build steps are specified?

adamperlin (changes requested):

docker-bake.hcl
cpuguy83: Done.

cpuguy83: done

PR #506, opened 2025-01-24
tests: Do not write test logs to summary

cpuguy83:
Test logs can be quite large and overflow the max summary size.
Instead just log the test name in the summary (this is for slow tests).


pmengelbert (approved):
:+1:

PR #509, opened 2025-01-29
feat: Add gomod git auth

pmengelbert:
See Commit Descriptions


generator_gomod.go
```diff
@@ -32,7 +38,12 @@ func (s *Spec) HasGomods() bool {
 
 func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.ConstraintsOpt) func(llb.State) llb.State {
 	return func(in llb.State) llb.State {
-		workDir := "/work/src"
+		const (
+			fourKB           = 4 * (1 << 10)
```
MorrisLaw, line 42: Why not just hardcode 4096 here? It's more obvious and we won't need to unnecessarily compute this every time a worker is ran for a path.

pmengelbert: It's a const so it'll be computed once at compile time, but 4096 is a lot clearer so I'll use that instead.

source_test.go
```diff
@@ -934,6 +1018,98 @@ func checkGitOp(t *testing.T, ops []*pb.Op, src *Source) {
...
+			continue
+		}
+
+		secrets[mnt.SSHOpt.ID] = struct{}{}
+	}
+
+	assert.Check(t, cmp.Len(secrets, 3), secrets)
+
+	for _, auth := range src.Generate[0].Gomod.Auth {
+		chk := auth.Token
```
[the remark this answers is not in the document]
pmengelbert: done

test/testenv/buildx.go
```diff
@@ -220,6 +222,21 @@ type FrontendSpec struct {
...
+	k string
+	v string
+}
+
+func withSecrets(so *client.SolveOpt, kvs ...keyVal) {
+	m := map[string][]byte{}
+	for _, kv := range kvs {
+		k := kv.k
+		v := kv.v
+		m[k] = []byte(v)
```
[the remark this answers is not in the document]
pmengelbert: done

source_test.go
```diff
@@ -934,6 +1018,98 @@ func checkGitOp(t *testing.T, ops []*pb.Op, src *Source) {
...
+
+func hasTmpFSMount(scriptOp *pb.ExecOp) bool {
+	for _, mnt := range scriptOp.Mounts {
+		trimmed := strings.TrimRight(mnt.Dest, "/")
+		if trimmed != "/dev/shm/git" {
+			continue
+		}
+
+		tfso := mnt.GetTmpfsOpt()
+		if tfso == nil {
```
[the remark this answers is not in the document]
pmengelbert: done

generator_gomod.go
pmengelbert: done

pmengelbert:
- [ ] Write docs for feature, including auto-propagating the auth from the git source

generator_gomod.go
```diff
@@ -54,6 +72,101 @@ func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.Constrai
 	}
 }
 
+func (g *SourceGenerator) gitconfigGeneratorScript() llb.State {
```
cpuguy83, line 75: Have you considered utilizing `GIT_ASKPASS` or just more generally git credential helpers?
I think it would probably be worthwhile to handle secrets with a helper of some sort that can just read secrets directly from `/run/secrets` (where buildkit puts them) instead of trying to deal with the configuration file.

source.go
```diff
@@ -712,8 +714,82 @@ func fillDefaults(s *Source) {
...
+	var (
+		initialize = func() {
+			if gm.Auth == nil {
+				gm.Auth = make(map[string]GomodGitAuth)
+			}
+		}
+
+		noop = func() {}
+
+		initOnce = func() {
```
cpuguy83, line 757: This seems a bit overzealous here.

cpuguy83 (changes requested):

generator_gomod.go
pmengelbert: AFAIK `GIT_ASKPASS` won't work with github sources since username/password authentication is not allowed by github. I'll look into the credential helpers.

pmengelbert: TIL about `GIT_ASKPASS`

source.go
pmengelbert: Fixed. Can just init right before assiging.

TIL you can read the value and bool from nil go maps (will always be the zero-value and false) but writing to them panics.

generator_gomod.go
```diff
@@ -32,7 +38,12 @@ func (s *Spec) HasGomods() bool {
 
 func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.ConstraintsOpt) func(llb.State) llb.State {
 	return func(in llb.State) llb.State {
-		workDir := "/work/src"
+		const (
+			fourKB           = 4096
+			workDir          = "/work/src"
+			scriptMountPoint = "/tmp/mnt"
```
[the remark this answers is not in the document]
pmengelbert: done

generator_gomod.go
```diff
@@ -1,14 +1,20 @@
...
 
 	"github.com/moby/buildkit/client/llb"
 	"github.com/pkg/errors"
 )
 
 const (
-	gomodCacheDir = "/go/pkg/mod"
+	gomodCacheDir       = "/go/pkg/mod"
+	gitConfigMountpoint = "/dev/shm/git"
+	scriptRelativePath  = "/go_mod_download.sh"
```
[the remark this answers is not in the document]
pmengelbert: done

generator_gomod.go
```diff
@@ -1,14 +1,20 @@
 package dalec
 
 import (
+	"bytes"
+	_ "embed"
```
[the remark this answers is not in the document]
pmengelbert: removed

pmengelbert:
- [ ] Fix rebase disaster.

pmengelbert:
@cpuguy83 This should be ready for another round of review. I'll probably have to modify the tests, but I'd like to get your thoughts on the overall approach.

generator_gomod.go
pmengelbert: done

source.go
```diff
@@ -712,8 +721,68 @@ func fillDefaults(s *Source) {
...
+	}
+
+	if gm.Auth == nil {
+		gm.Auth = make(map[string]GomodGitAuth)
+	}
+
+	gm.Auth[host] = gomodAuth
+}
+
+// func (gm *GeneratorGomod) fillDefaults
```
MorrisLaw, line 784: leftover?

pmengelbert: fixed

generator_gomod.go
```diff
@@ -30,9 +34,18 @@ func (s *Spec) HasGomods() bool {
 	return false
 }
 
-func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.ConstraintsOpt) func(llb.State) llb.State {
+func withGomod(g *SourceGenerator, srcSt, worker, credHelper llb.State, opts ...llb.ConstraintsOpt) func(llb.State) llb.State {
 	return func(in llb.State) llb.State {
-		workDir := "/work/src"
+		const (
+			fourKB                       = 4096
```
cpuguy83, line 40: Doesn't look like this is used anywhere?

generator_gomod.go
```diff
@@ -54,6 +77,100 @@ func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.Constrai
...
+}
+
+func (g *SourceGenerator) mountGitAuthConfig(mountPoint, basename string) llb.RunOption {
+	return RunOptFunc(func(ei *llb.ExecInfo) {
+		if g.Gomod == nil || g.Gomod.Auth == nil {
+			return
+		}
+
+		var b bytes.Buffer
+		enc := gob.NewEncoder(&b)
```
cpuguy83, line 93: Why gob?

generator_gomod.go
```diff
@@ -54,6 +77,100 @@ func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.Constrai
...
+			if sshConfig.Username != "" {
+				username = sshConfig.Username
+			}
+
+			fmt.Fprintf(&script, `git config --global "url.ssh://%[1]s@%[2]s/.insteadOf" https://%[2]s/`, username, host)
+			script.WriteRune('\n')
+			continue
+		}
+
+		fmt.Fprintf(&script, `git config --global credential."https://%s".helper "gomod %s"`, host, configPath)
```
cpuguy83, line 123: Let's give this a name that's more dalec specific?
`dalec-git-credentials`?

generator_gomod.go
```diff
@@ -54,6 +77,100 @@ func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.Constrai
...
+			if auth.SSH != nil && auth.SSH.ID != "" {
+				llb.AddSSHSocket(llb.SSHID(auth.SSH.ID)).SetRunOption(ei)
+
+				setenv()
+				setenv = noop
+			}
+		}
+
+		for secret := range secrets {
+			llb.AddSecret("/run/secrets/"+secret, llb.SecretID(secret)).SetRunOption(ei)
```
cpuguy83, line 169: What it help simplify things to mount the secret such that we can find it by host name instead of having to encode configs?
e.g. something like `/run/secrets/dalec-git-auth/<host>`

source.go
```diff
@@ -18,6 +19,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	GitCredentialHelperGomod       = "git-credential-gomod"
+	GitCredentialHelperGomodSrcDir = "/usr/bin"
```
cpuguy83, line 24: Doesn't seem like this is used.

source.go
```diff
@@ -226,11 +232,13 @@ var (
...
+type GitCredHelperGetter func() (llb.State, error)
 
 type SourceOpts struct {
-	Resolver   llb.ImageMetaResolver
-	Forward    ForwarderFunc
-	GetContext func(string, ...llb.LocalOption) (*llb.State, error)
+	Resolver             llb.ImageMetaResolver
+	Forward              ForwarderFunc
+	GetContext           func(string, ...llb.LocalOption) (*llb.State, error)
+	GitCredentialHelpers map[string]GitCredHelperGetter
```
cpuguy83, line 241: Shouldn't there only ever be one helper here and it doesn't seem like it should be tied to any particular need (e.g. gomod).
This is how we make our buildkit secrets available to git, shouldn't need much more than that.

cpuguy83 (changes requested):

generator_gomod.go
pmengelbert: to avoid yaml messiness -- we've already got a go struct, why not keep it that way rather than translating it into another form and back?

generator_gomod.go
pmengelbert: I was thinking the same thing earlier today. I'll set the destination path on the secrets instead.

generator_gomod.go
cpuguy83: Mostly because its a weird encoding and is actually less efficient than the stdlib json encoding.

generator_gomod.go
pmengelbert: We'll need to encode the auth type also (token or header).

generator_gomod.go
pmengelbert: You can give it an absolute path also, so I'm just giving it the path to the frontend.

cmd/frontend/git_credential_gomod.go
```diff
@@ -0,0 +1,272 @@
...
+	return payload
+}
+
+func printPayload(payload *gitPayload) (string, error) {
+	var buf bytes.Buffer
+
+	var errs []error
+	fill := func(k, v string) {
+		if v != "" {
+			if _, err := fmt.Fprintln(&buf, k+"="+v); err != nil {
```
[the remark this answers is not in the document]
pmengelbert: done

cmd/frontend/main.go
```diff
@@ -19,9 +20,26 @@ import (
 
 const (
 	Package = "github.com/Azure/dalec/cmd/frontend"
+
+	frontendBasename        = "frontend"
+	gomodCredHelperBasename = "git-credential-gomod"
 )
 
 func main() {
+	cmd := filepath.Base(os.Args[0])
```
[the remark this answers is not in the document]
pmengelbert: Fixed this. There are now 2 subcommands, `frontend` and `credential-helper`. If the subcommand is blank, it behaves as though the subcommand were `frontend`.

generator_gomod.go
pmengelbert: Removed this and other unused `const`s.

generator_gomod.go
pmengelbert: Fixed this.

pmengelbert: By using the filepath to convey the necessary information.

generator_gomod.go
pmengelbert: done.

source.go
pmengelbert: Removed this and others.

source.go
pmengelbert: Changed this to a `func() (llb.RunOption, error)`. The error is thrown if the frontend state can't be acquired. The runopt mounts the frontend at `/usr/local/bin/frontend` and can be used anywhere the credential helper is needed.

test/testenv/buildx.go
```diff
@@ -292,6 +307,16 @@ func (b *BuildxEnv) RunTest(ctx context.Context, t *testing.T, f TestFunc, opts
 	withProjectRoot(t, &so)
 	withResolveLocal(&so)
 
+	withSecrets(&so,
```
cpuguy83, line 310: Seems like secrets should be a `TestRunnerOpt`

cpuguy83: Also doesn't seem like we're actually using this in any tests?

cpuguy83 (changes requested):
Let's think about how we can test these changes in the integration suite.
I think if we can run a service on the host that is accessible to buildkit it shouldn't be a problem (other than just setting up multiple different git services).

cmd/frontend/main.go
```diff
@@ -19,12 +21,43 @@ import (
...
+
+	subCmd := fs.Arg(1)
+
+	// each "sub-main" function handles its own exit
+	switch subCmd {
+	case credHelperSubcmd:
+		args := flag.Args()[2:]
+		// skip os.Args[0] and "credential-helper"
+		gomodMain(args)
+	case "", frontendBasename:
```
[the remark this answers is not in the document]
pmengelbert: done.

test/testenv/buildx.go
pmengelbert: It got added to the main `RunTest`. I'll move it to the test proper as a `TestRunnerOpt`.

pmengelbert: Nevermind. This was before I got rid of the integration tests.

pmengelbert:
> Let's think about how we can test these changes in the integration suite. I think if we can run a service on the host that is accessible to buildkit it shouldn't be a problem (other than just setting up multiple different git services).

I don't  know how we do this without monkeying around with network namespaces. If that's the case, the tests would only run on linux. Setting the namespaces such that the frontend can communicate with a service running on the host

source.go
```diff
@@ -712,6 +715,64 @@ func fillDefaults(s *Source) {
...
+}
+
+func (git *SourceGit) fillDefaults(generators []*SourceGenerator) {
+	if git == nil {
+		return
+	}
+
+	u, err := url.Parse(git.URL)
+	if err != nil {
+		// This should be validated elsewhere
```
adamperlin, line 728: Is this a panic situation?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=${BUILDPLATFORM} mcr.microsoft.com/oss/go/microsoft/golang:1.22@sha256:9e5243001d0f43d6e1c556a0ce8eedc1bc80eb37b5210036b289d2c601bc08b6 AS go
+FROM --platform=${BUILDPLATFORM} mcr.microsoft.com/oss/go/microsoft/golang:1.24@sha256:7c2024496593127e761a4610d628e7f653168342e7b60b21d6745adc236b0bd9 AS go
```
cpuguy83, line 1: Should stick with 1.23 here (which, IIRC dependabot updated this week)

source.go
```diff
@@ -712,6 +720,64 @@ func fillDefaults(s *Source) {
...
+		return
+	}
+
+	for _, generator := range generators {
+		generator.fillDefaults(u.Host, &git.Auth)
```
cpuguy83, line 738: Is this so we don't have to duplicate auth in source and generator config?
Can you add a comment on that?

source_test.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/moby/buildkit/solver/pb"
 	"github.com/opencontainers/go-digest"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/stretchr/testify/require"
```
cpuguy83, line 22: Let's stick with `gotest.tools/v3/assert`

test/source_test.go
```diff
@@ -1,20 +1,338 @@
...
+	// this is needed because go behaves differently when the host is
+	// recognized if it sees `host.docker.internal` it will first try to
+	// resolve the metadata by connecting to the resolved IP address on port
+	// 443, which will fail.
+	const host = "host.docker.internal"
+
+	ctx := startTestSpan(baseCtx, t)
+	sourceName := "gitauth"
+
+	// randomData := identity.NewID()
```
cpuguy83, line 46: Delete

test/source_test.go
```diff
@@ -1,20 +1,338 @@
...
+		}
+	}
+
+	if pid == 0 {
+		t.Fatal("cannot find pid of docker daemon")
+	}
+	return pid
+}
+
+func runGitServer(addr string, port int) {
```
cpuguy83, line 241: Can you pass through a `*testing.T` shutdown the listener in a `t.Cleanup` function?

test/source_test.go
```diff
@@ -1,20 +1,338 @@
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+
+		outsideAddr, insideAddr := getGitServerAddrs(ctx, t, c)
```
cpuguy83, line 87: I don't think we are using these values anywhere other than `runGitServer`, maybe it makes sense to have `runGitServer` handle all the things here?

test/source_test.go
```diff
@@ -1,20 +1,338 @@
...
+
+		filename := filepath.Join(host, "user/private.git@v0.0.0/hello")
+		checkFile(ctx, t, filename, res, []byte("hello\n"))
+	}, testenv.WithSecrets(testenv.KeyVal{
+		K: "super-secret",
+		V: "value",
+	}))
+}
+
+func initGomodWorker(c gwclient.Client, host string, port int) llb.State {
```
cpuguy83, line 106: Can you explain why we need to configure git this way?

test/source_test.go
```diff
@@ -713,3 +1031,53 @@ func checkFileStat(t *testing.T, dir fs.FS, p string, opt checkFileStatOpt) {
 		t.Errorf("expected file %q isDir=%v, got %v", p, opt.IsDir, stat.IsDir())
 	}
 }
+
+func isRootless(ctx context.Context, t *testing.T, client gwclient.Client) bool {
+	isRootlessOnce.Do(func() {
+		st := llb.Image("mcr.microsoft.com/azurelinux/base/core:3.0").Run(llb.Args([]string{
```
cpuguy83, line 1037: Needs a platform... probably `path.Join("linux", runtime.GOARCH)` would suffice here.

test/source_test.go
```diff
@@ -1,20 +1,338 @@
...
+		filename := filepath.Join(host, "user/private.git@v0.0.0/hello")
+		checkFile(ctx, t, filename, res, []byte("hello\n"))
+	}, testenv.WithSecrets(testenv.KeyVal{
+		K: "super-secret",
+		V: "value",
+	}))
+}
+
+func initGomodWorker(c gwclient.Client, host string, port int) llb.State {
+	worker := llb.Image("alpine:latest", llb.WithMetaResolver(c)).
```
cpuguy83, line 107: Needs platform here

cpuguy83 (changes requested):
The test can't run locally for me due to this error (beyond needing to add the platform to the image refs used).

> go: host.docker.internal/user/private.git@v0.0.0: no secure protocol found for repository

test/source_test.go
```diff
@@ -1,20 +1,338 @@
...
+		llb.AddEnv("HOST", host),
+		llb.AddEnv("PORT", fmt.Sprintf("%d", port)),
+	).Root()
+	return worker
+}
+
+func getGitServerAddrs(ctx context.Context, t *testing.T, c gwclient.Client) (string, string) {
+	const (
+		rootlessOutsideAddr = "localhost"
+		rootlessInsideAddr  = "10.0.2.2" // as per the docs
```
adamperlin, line 129: can you clarify which docs in your comment?

frontend/gateway.go
```diff
@@ -74,6 +74,9 @@ func ForwarderFromClient(ctx context.Context, client gwclient.Client) dalec.Forw
 		req, err := newSolveRequest(
 			toDockerfile(ctx, st, dockerfileDt, spec, dalec.ProgressGroup("prepare dockerfile to forward to frontend")),
 			copyForForward(ctx, client),
+			func(sr *gwclient.SolveRequest) error {
```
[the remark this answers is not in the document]
pmengelbert: Started doing something here, then took another approach. Left it in by mistake.

test/source_test.go
pmengelbert: `insideAddr` gets added as an extra host a few lines down.

test/source_test.go
pmengelbert: Because you can't have a port number in a go module name. We tell git to use `host:port` so that when go invokes git to fetch the module, git knows to use `host:port` instead of just `host`. This is all because the git server is not running on 443/80.

test/source_test.go
pmengelbert: Thanks.

test/source_test.go
pmengelbert: Had to add this to the `debug/gomods` handler as well -- there was no platform there.

test/source_test.go
pmengelbert: Added to comment (done).

test/source_test.go
pmengelbert: done

generator_gomod.go
```diff
@@ -41,19 +50,108 @@ func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.Constrai
 			paths = []string{"."}
 		}
 
+		sort.Strings(paths)
```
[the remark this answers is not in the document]
pmengelbert: removed

frontend/gateway.go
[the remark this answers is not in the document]
pmengelbert: removed

frontend/debug/handle_sources.go
```diff
@@ -24,18 +24,24 @@ func Sources(ctx context.Context, client gwclient.Client) (*client.Result, error
...
+		// st := llb.Scratch().File(llb.Mkfile("/out", 0o644, []byte(fmt.Sprintf("%#v", client.BuildOpts().Opts))))
+		def, err := st.Marshal(ctx)
 		if err != nil {
 			return nil, nil, err
 		}
 
 		res, err := client.Solve(ctx, gwclient.SolveRequest{
 			Definition: def.ToPB(),
+			FrontendOpt: map[string]string{
+				"add-hosts": "host.docker.internal=10.0.2.2",
```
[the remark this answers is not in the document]
pmengelbert: removed

source_test.go
pmengelbert: done

test/source_test.go
pmengelbert: done

Dockerfile
pmengelbert: done

source.go
pmengelbert: done

cmd/frontend/main.go
```diff
@@ -19,12 +21,42 @@ import (
...
+
+	subCmd := fs.Arg(1)
+
+	// each "sub-main" function handles its own exit
+	switch subCmd {
+	case credHelperSubcmd:
+		args := flag.Args()[2:]
+		// skip os.Args[0] and "credential-helper"
+		gomodMain(args)
+	case "":
```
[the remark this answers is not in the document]
pmengelbert: done

test/source_test.go
cpuguy83: The `debug/gomods` is not a big deal since that LLB is generated on the server side.
Here it is generated on the client side and it's wanting to use the client platform by default, which is a problem when trying to run tests from non-linux clients.

pmengelbert:
> The test can't run locally for me due to this error (beyond needing to add the platform to the image refs used).
> 
> > go: host.docker.internal/user/private.git@v0.0.0: no secure protocol found for repository

Can you please try this again with the latest changes? I was experiencing the same error this morning but I believe it's fixed now.

pmengelbert:
> > The test can't run locally for me due to this error (beyond needing to add the platform to the image refs used).
> > > go: host.docker.internal/user/private.git@v0.0.0: no secure protocol found for repository
> 
> Can you please try this again with the latest changes? I was experiencing the same error this morning but I believe it's fixed now.

Disregard for now. I need to fix my rebase.

test/gomod_git_auth_test.go
```diff
@@ -0,0 +1,986 @@
...
+			)
+
+			solveResultChan := make(chan *gwclient.Result)
+			solveErrChan := make(chan error)
+			solveTCh(ctx, t, testState.client(), sr, solveResultChan, solveErrChan)
+
+			var res *gwclient.Result
+			select {
+			case err := <-httpErrChan:
+				t.Fatalf("ssh server unexpededly failed: %s", err)
```
cpuguy83: Couple of typos here.

test/source_test.go
```diff
@@ -2,6 +2,7 @@ package test
 
 import (
 	"context"
+	_ "embed"
```
cpuguy83: Is that needed?

test/testenv/buildx.go
```diff
@@ -245,20 +255,57 @@ type TestRunnerConfig struct {
...
 type SolveStatus = client.SolveStatus
+type SolveOpt = client.SolveOpt
 
 func WithSolveStatusFn(f func(*SolveStatus)) TestRunnerOpt {
 	return func(cfg *TestRunnerConfig) {
 		cfg.SolveStatusFn = f
 	}
 }
 
+func WithSecrets(kvs ...KeyVal) TestRunnerOpt {
```
cpuguy83: Any reason to not just do a single key/value here and make multiple calls?

test/testenv/buildx.go
```diff
@@ -320,6 +371,57 @@ func (b *BuildxEnv) RunTest(ctx context.Context, t *testing.T, f TestFunc, opts
...
+)
+
+func NewWithNetHostBuildxInstance(ctx context.Context, t *testing.T) *BuildxEnv {
+	dgst := digest.Canonical.Encode([]byte(t.Name()))
+
+	var randomBytes [8]byte
+	_, err := rand.Read(randomBytes[:])
+	if err != nil {
+		t.Fatal(err)
+	}
```
cpuguy83: Not needed anymore.

test/testenv/buildx.go
```diff
@@ -15,11 +16,15 @@ import (
...
 	"testing"
+	"time"
 
 	"github.com/moby/buildkit/client"
 	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+	"github.com/moby/buildkit/session/secrets/secretsprovider"
+	"github.com/moby/buildkit/session/sshforward/sshprovider"
 	"github.com/moby/buildkit/solver/pb"
 	spb "github.com/moby/buildkit/sourcepolicy/pb"
+	"github.com/opencontainers/go-digest"
```
cpuguy83, line 26: Make sure `crypto/sha256` is imported otherwise go-digest will panic on `digest.Canonical.<whatever>`

test/gomod_git_auth_test.go
```diff
@@ -0,0 +1,986 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+	return spec
+}
+
+func (ts *TestState) startHTTPServer(gitHost llb.State) chan error {
```
cpuguy83: Can use `<- chan error` if the caller is receive-only

test/gomod_git_auth_test.go
```diff
@@ -0,0 +1,986 @@
...
+	cont := ts.newContainer(gitHost)
+
+	env := ts.getStateEnv(gitHost)
+	errChan := ts.runContainer(cont, env, serverScript)
+
+	t.Log("waiting for http server to come online")
+
+	timeout := waitOnlineTimeout
+	ts.runWaitScript(cont, env, waitScript, timeout)
+	// time.Sleep(time.Second * 86400)
```
cpuguy83: Remove

test/testenv/buildx.go
```diff
@@ -224,6 +229,11 @@ type FrontendSpec struct {
 	Build gwclient.BuildFunc
 }
 
+type KeyVal struct {
```
cpuguy83: Remove is changing `WithSecrets` below.

test/testenv/buildx.go
pmengelbert: It won't panic on `digest.Canonical.Encode` because all `Encode` does is `fmt.Sprintf("%x")`. It relies on the caller to do the digesting first.

test/gomod_git_auth_test.go
pmengelbert: done

test/source_test.go
pmengelbert: done

test/testenv/buildx.go
pmengelbert: done

test/testenv/buildx.go
pmengelbert: done

test/testenv/buildx.go
pmengelbert: fixed

test/gomod_git_auth_test.go
pmengelbert: done

test/testenv/buildx.go
pmengelbert: done

test/gomod_git_auth_test.go
pmengelbert: done

test/helpers_test.go
```diff
@@ -283,6 +283,20 @@ func solveT(ctx context.Context, t *testing.T, gwc gwclient.Client, req gwclient
 	return res
 }
 
+func solveTCh(ctx context.Context, t *testing.T, gwc gwclient.Client, req gwclient.SolveRequest, rc chan<- *gwclient.Result, ec chan error) {
```
MorrisLaw: Good job defining a parameter that guarantees a send-only channel is used/expected!

MorrisLaw:
```suggestion
func solveTCh(ctx context.Context, t *testing.T, gwc gwclient.Client, req gwclient.SolveRequest, rc chan<- *gwclient.Result, ec chan<- error) {
```

Let's update this too to make the intent clear that it's also a send-only channel similar to `rc`.

cmd/frontend/git_credential_helper.go
```diff
@@ -0,0 +1,227 @@
...
+			payload.url = v
+		case "authtype":
+			payload.authtype = v
+		case "credential":
+			payload.credential = v
+		case "ephemeral":
+			payload.ephemeral = v
+		case "continue":
+			payload.kontinue = v
+		case "wwwauth[]":
```
MorrisLaw: I'm wondering if it'd make sense to have these payload values be constants.

- there are a lot of them
- we use them in at least 2 places which could lead to possible drift or typos
- defining the payload constants in one location will mitigate that risk and make it better for maintainability

docs/spec.schema.json
```diff
@@ -918,6 +928,60 @@
...
 			"description": "GoBuildCache is a cache for Go build artifacts."
 		},
+		"GomodGitAuth": {
+			"properties": {
+				"header": {
+					"type": [
+						"string",
+						"null"
+					],
+					"description": "Header is the name of the secret which contains the git auth header.\nwhen using git auth header based authentication.\nNote: This should not have the *actual* secret value, just the name of\nthe secret which was specified as a build secret."
```
MorrisLaw:
```suggestion
					"description": "Header is the name of the secret that contains the git auth header\nwhen using git auth header based authentication.\nNote: This should not have the *actual* secret value, just the name of\nthe secret which was specified as a build secret."
```

generator_gomod.go
```diff
@@ -45,26 +53,120 @@ func withGomod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.Constrai
...
+
+func (g *SourceGenerator) withGomodSecretsAndSockets() llb.RunOption {
+	return runOptionFunc(func(ei *llb.ExecInfo) {
+		if g.Gomod == nil {
+			return
+		}
+
+		const basePath = "/run/secrets"
+
+		for host, auth := range g.Gomod.Auth {
```
MorrisLaw, line 145: I know we have a `fillDefaults` function that makes sure to initialize the Auth map. Despite that, are we 100% sure there is not a code path that will hit this with a nil `Auth` value? Should we add a check just in case?

cmd/frontend/git_credential_helper.go
```diff
@@ -0,0 +1,227 @@
...
+	fill("ephemeral", payload.ephemeral)
+	fill("continue", payload.kontinue)
+	fillArray("wwwauth[]", payload.wwwauth)
+	fillArray("capability[]", payload.capability)
+	fillArray("state[]", payload.state)
+
+	return buf.String()
+}
+
+func generateResponse(payload *gitPayload, secret []byte, kind string) (string, error) {
```
MorrisLaw, line 225: I think it's a bit more idiomatic and clear to make the `unrecognized type` case a `default` within the switch statement. That way, all return paths are handled explicitly inside the switch.

If the fallback case was complex then I think leaving it out would make sense, but since it's simple here, keeping it inside the switch improves readability and maintainability.

```suggestion
func generateResponse(payload *gitPayload, secret []byte, kind string) (string, error) {
	switch kind {
	case "token":
		return handleSecretToken(secret, payload)
	case "header":
		return handleSecretHeader(secret, payload)
	default:
		return "", fmt.Errorf("unrecognized authType: %q", kind)
	}
}
```

cmd/frontend/main.go
```diff
@@ -21,12 +23,40 @@ import (
...
+
+func main() {
+	fs := flag.CommandLine
+	fs.Usage = func() {
+		fmt.Fprintf(os.Stderr, `usage: %s [subcommand [args...]]`, os.Args[0])
+	}
+
+	if err := fs.Parse(os.Args); err != nil {
+		bklog.L.WithError(err).Fatal("error parsing frontend args")
+		os.Exit(137)
```
MorrisLaw, line 43: I know we use 137 when erroring on `running the frontend`. And now we will use it when erroring on `parsing the frontend`.

IIUC, the 137 means SIGKILL, which typically means on Linux that either we've done a `kill -9` ourselves or the system has done it (typically because it's ran out of memory).

I'm just curious about the historical context/rationale behind choosing 137 before (for running frontend), and now parsing as well. Why not something more conventional like `os.Exit(1)` or `os.Exit(2)`?

test/cmd/git_repo/host.go
```diff
@@ -0,0 +1,61 @@
...
+	if len(args) < 4 {
+		return errUsage
+	}
+
+	repo := args[1]
+	addr := args[2]
+	port := args[3]
+	gitHandler := githttp.New(repo)
+	authr := auth.Authenticator(func(ai auth.AuthInfo) (bool, error) {
+		if ai.Push {
```
MorrisLaw, line 37: For a split second I thought you snuck some AI in there 😅

test/helpers_test.go
pmengelbert: fixed

cmd/frontend/git_credential_helper.go
pmengelbert: done

docs/spec.schema.json
pmengelbert: done

generator_gomod.go
pmengelbert: If `Auth` is nil, the loop will not execute:
[https://go.dev/ref/spec#For_range](https://go.dev/ref/spec#For_range) (see #3 in the bullet points, if the map is nil the number of iterations is zero)

cmd/frontend/git_credential_helper.go
pmengelbert: done

cmd/frontend/main.go
pmengelbert: @cpuguy83 Do you remember why you used 137 at the outset? Is there some kind of expectation that buildkit has around exiting with that code?

@MorrisLaw I'm not really sure

cpuguy83: 137 is incorrect for sure.

test/gomod_git_auth_test.go
```diff
@@ -0,0 +1,1067 @@
...
+
+			filename := calculateFilename(ctx, t, attr, res)
+			checkFile(ctx, t, filename, res, []byte("bar\n"))
+		})
+	})
+}
+
+// GitServicesAttributes are the basic pieces of information needed to host two git
+// servers, one via SSH and one via HTTP
+type GitServicesAttributes struct {
```
cpuguy83: Maybe this along with all the git setup stuff (and things like the `file` type) should go into a separate package so as not to pollute the namespace of this package.

cpuguy83 (changes requested):

cmd/frontend/main.go
MorrisLaw: > Is there some kind of expectation that buildkit has around exiting with that code?

I like your line of thinking here, @pmengelbert. Let’s clarify exactly what that expectation is, how it ties into the Dalec UX, and then align that behavior with the appropriate OS exit code.

generator_gomod.go
MorrisLaw: Oh wow, I expected a nil pointer exception. That makes sense then. Thanks for sharing!

P.S. `#` before a number in GitHub will point to an issue within the repo

MorrisLaw (changes requested):
Looking good, we're close! Just need to address this: https://github.com/Azure/dalec/pull/509#discussion_r2152228303

And also Brian's comment.

test/gomod_git_auth_test.go
pmengelbert: done

cmd/frontend/main.go
pmengelbert: @MorrisLaw IMO this is out of scope for this PR, since this PR didn't add 137 -- I just moved it to a different location in code to accomplish what was needed for this feature. I'll open an issue for it, but this PR is big enough.

pmengelbert: https://github.com/Azure/dalec/issues/662

PR #511, opened 2025-01-31
Preserve extension fields when marshalling/unmarshalling

cpuguy83:
This makes it so external callers don't need to handle this themselves.


spec.go
```diff
@@ -94,6 +94,8 @@ type Spec struct {
 	// Each item in this list is run with a separate rootfs and cannot interact with other tests.
 	// Each [TestSpec] is run with a separate rootfs, asynchronously from other [TestSpec].
 	Tests []*TestSpec `yaml:"tests,omitempty" json:"tests,omitempty"`
+
+	Ext map[string]interface{} `yaml:"-" json:"-"`
```
cpuguy83, line 98: I wonder if we can do something different here and allow callers to unmarshal directly to their own data type?

cpuguy83: Seems difficult to deal with while also requiring fields to start with `x-`

cpuguy83: Actually I think this doesn't matter since the map value would be the callers data type, which the caller can assert.

cpuguy83:
Converting this to draft since I realized it really doesn't help much except for when the extension field value is a core type.

load.go
```diff
@@ -210,36 +211,93 @@ func (s *Spec) SubstituteArgs(env map[string]string, opts ...SubstituteOpt) erro
...
+	if err := yaml.UnmarshalWithOptions([]byte(parsed.Docs[0].String()), &s2, yaml.Strict()); err != nil {
+		return err
+	}
+
+	dt, err = yaml.Marshal(ext)
+	if err != nil {
+		return err
+	}
+
+	parsedExt, err := parser.ParseBytes(dt, 0)
```
pmengelbert, line 272: ditto

spec.go
```diff
@@ -94,6 +98,14 @@ type Spec struct {
...
 	// Each [TestSpec] is run with a separate rootfs, asynchronously from other [TestSpec].
 	Tests []*TestSpec `yaml:"tests,omitempty" json:"tests,omitempty"`
+
+	// extRaw is the raw AST of the extension fields in the spec.
+	// This is used to extract the ext fields in [Spec.Ext]
+	extRaw *ast.File
+	// ext is all the ext fields in the spec.
+	// This gets used when marshalling the spec back to yaml.
+	// It is used to avoid having to re-parse the raw AST.
+	ext map[string]interface{}
```
pmengelbert, line 108: I don't like adding private fields to the spec struct, especially abstract stuff like `*ast.File`. We worked hard early on to ensure that the spec remained simple data, rather than data plus functionality.

For example, we decided not to implement the `Source` type as a group of interfaces because we wanted to maintain the spec as simple data. This also means that anyone importing the spec object will be importing `goccy/go-yaml/ast`.

I recognize that it will be a pain to wrap the spec and change function signatures, but a lot of that could be done with a find/replace operation and having the first member of the wrapper type be the spec:
```go
type specExtensions struct {
        extRaw *ast.File
        ext map[string]interface{}
}

type SpecWithExtensions struct {
        *Spec
        ext specExtensions
}
````

pmengelbert (changes requested):
See the above comment. At the very least, we need to give it some serious thought before adding private fields with abstract data types into the main `Spec` struct.

spec.go
cpuguy83: `Spec` is just a data type.
Currently it is unable to represent all the data that can go into a spec since we can't dynamically generate fields for extension fields.

These changes make it so `Spec` can represent that data and make it accessible to those who want it.
Its not changing the actual dalec spec in any, is not exposed outside of this package.

What concerns do you have for this?

pmengelbert (approved):
LGTM

PR #518, opened 2025-02-04
Make matches a slice of regexes instead of a single regex

### LINKED ISSUE #513: [REQ] CheckOutput.Matches should take a slice instead of string

None

Right now the [`Matches` field](https://github.com/Azure/dalec/blob/3c4c901fdd24c274f240bf4dcf297b6dd2c31f87/tests.go#L62) is just a string.
This makes it much more difficult to write a regex that matches all possible things you might want to check.

I suggest instead it should take a slice of strings to simplify writing regex matches for different cases.

bchuo:
**What this PR does / why we need it**: Changes the typing of `matches` such that it now takes a slice of strings, allowing for easier writing of regex matches for different cases.

Fixes #513 

### DIFF
--- a/tests.go
+++ b/tests.go
@@ -58,7 +58,7 @@ type CheckOutput struct {
 	Equals string `yaml:"equals,omitempty" json:"equals,omitempty"`
 	// Contains is the list of strings to check if they are contained in the output.
 	Contains []string `yaml:"contains,omitempty" json:"contains,omitempty"`
-	// Matches is the regular expression to match the output against.
+	// Matches is the list of regular expressions to match the output against.
 	Matches []string `yaml:"matches,omitempty" json:"matches,omitempty"`
 	// StartsWith is the string to check if the output starts with.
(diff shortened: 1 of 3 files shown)


pmengelbert:
Thanks for the PR! You'll need to run `go generate` to get the auto-generated docs in line with the current code.

tests.go
```diff
@@ -260,14 +262,16 @@ func (c CheckOutput) Check(dt string, p string) (retErr error) {
...
-	if c.Matches != "" {
-		regexp, err := regexp.Compile(c.Matches)
-		if err != nil {
-			return err
-		}
-
-		if !regexp.Match([]byte(dt)) {
-			return &CheckOutputError{Kind: "matches", Expected: c.Matches, Actual: dt, Path: p}
+	for _, matches := range c.Matches {
+		if matches != "" {
```
cpuguy83, line 266: I think we should remove this check here since this would mean someone gave an invalid input.

cpuguy83 (changes requested):

cpuguy83 (approved):

PR #524, opened 2025-02-11
Add tests cases for malformed and conflicting patches

### LINKED ISSUE #500: Add test case to ensure conflicting patches cause an error
When applying patches to a source there could be a couple of things that go wrong:

1. Patch just does not apply (be it malformed, or just incorrect content for the given source/commit)
2. When applying multiple patches and patch A conflicts with patch B

We should make sure there is a test case to ensure that we actually produce an error in these cases.

bchuo:
**What this PR does / why we need it**: We need to test to ensure we fail when a patch is either malformed or conflicts with another patch.

Fixes #500 

### DIFF
--- a/source_test.go
+++ b/source_test.go
@@ -481,6 +481,54 @@ func TestSourceDockerImage(t *testing.T) {
 	})
 }
 
+func TestPatchSource_MalformedPatch(t *testing.T) {
+	worker := llb.Scratch()
+	sourceState := llb.Scratch().File(
+		llb.Mkfile("/file.txt", 0644, []byte("Hello World")),
+	)
+	sourceToState := map[string]llb.State{
+		"malformed_patch": llb.Scratch().File(
+			llb.Mkfile("/malformed.patch", 0644, []byte("invalid patch content")),
+		),
+	}
+	patchNames := []PatchSpec{
+		{
+			Source: "malformed_patch",
+		},
+	}
+
+	assert.Assert(t, func() {
+		patchSource(worker, sourceState, sourceToState, patchNames)
+	} != nil, "patch -p1 < /patch: exit status 1")
+}
+
+func TestPatchSource_ConflictingPatches(t *testing.T) {
+	worker := llb.Scratch()
+	sourceState := llb.Scratch().File(
+		llb.Mkfile("/file.txt", 0644, []byte("Hello World")),
+	)
+	sourceToState := map[string]llb.State{
+		"patch1": llb.Scratch().File(
+			llb.Mkfile("/patch1.patch", 0644, []byte("diff --git a/file.txt b/file.txt\nindex 123..456 100644\n--- a/file.txt\n+++ b/file.txt\n@@ -1 +1 @@\n-Decaf Bad\n+Cafe Good")),
+		),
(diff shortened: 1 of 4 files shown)


source_test.go
```diff
@@ -481,6 +481,54 @@ func TestSourceDockerImage(t *testing.T) {
 	})
 }
 
+func TestPatchSource_MalformedPatch(t *testing.T) {
+	worker := llb.Scratch()
+	sourceState := llb.Scratch().File(
+		llb.Mkfile("/file.txt", 0644, []byte("Hello World")),
```
cpuguy83, line 487: Can you use the updated octal format `0o644` for everything?

source_test.go
```diff
@@ -481,6 +481,54 @@ func TestSourceDockerImage(t *testing.T) {
...
+			llb.Mkfile("/malformed.patch", 0644, []byte("invalid patch content")),
+		),
+	}
+	patchNames := []PatchSpec{
+		{
+			Source: "malformed_patch",
+		},
+	}
+
+	assert.Assert(t, func() {
```
cpuguy83, line 500: It looks like this is testing that the closure is not nil which doesn't look correct.

I think the easiest way to test these is in the integration tests (under `test/`) where we can get it to actually run the patch and produce an error.

cpuguy83 (changes requested):

test/source_test.go
```diff
@@ -713,3 +714,69 @@ func checkFileStat(t *testing.T, dir fs.FS, p string, opt checkFileStatOpt) {
...
+	}
+	patchNames := []dalec.PatchSpec{
+		{
+			Source: "malformed_patch",
+		},
+	}
+
+	spec := &dalec.Spec{
+		Patches: map[string][]dalec.PatchSpec{
+			"patchSource": patchNames,
```
cpuguy83, line 736: This should be the name of the source that we are going to patch.
e.g.

```
"source1": patchNames,
```

cpuguy83 (changes requested):

test/source_test.go
```diff
@@ -713,3 +713,85 @@ func checkFileStat(t *testing.T, dir fs.FS, p string, opt checkFileStatOpt) {
...
+					Inline: &dalec.SourceInline{
+						File: &dalec.SourceInlineFile{
+							Contents: "diff --git a/file.txt b/file.txt\nindex 123..789 100644\n--- a/file.txt\n+++ b/file.txt\n@@ -1 +1 @@\n-Hello World\n+Hello Galaxy",
+						},
+					},
+				},
+			},
+		}
+
+		req := newSolveRequest(withSpec(ctx, t, spec))
```
cpuguy83, line 790: I think we need to add `withTarget("debug/sources")` here and potentially need to check error type/value.
I think buildkit would pass along a `*moby_buildkit_v1_frontend.ExitError` error in this case which should signify that the patch command exited non-zero.

cpuguy83: Same goes for the other test case.

cpuguy83 (changes requested):

cpuguy83 (approved):

PR #535, opened 2025-02-19
Merge Dependency Fields from Global Spec Dependencies into Targets

adamperlin:
This applies an inheritance type relationship between package dependencies for individual dalec targets and package dependencies specified in the default spec `dependencies` block. Basically, if a dependency field is specified for a particular target, it will be used, otherwise the field from the global dependencies will be inherited. Fields themselves are not merged together; this is an "override if unset in specific target" type of behavior.

Fixes #

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -286,55 +286,27 @@ func ProgressGroup(name string) llb.ConstraintsOpt {
 }
 
 func (s *Spec) GetRuntimeDeps(targetKey string) []string {
-	var deps *PackageDependencies
-	if t, ok := s.Targets[targetKey]; ok {
-		deps = t.Dependencies
-	}
-
+	deps := s.GetPackageDeps(targetKey)
 	if deps == nil {
-		deps = s.Dependencies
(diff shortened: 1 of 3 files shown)


helpers.go
```diff
@@ -445,14 +417,73 @@ func SortedMapValues[T any](m map[string]T) []T {
 	return out
 }
 
+// MergeDependencies merges two sets of package dependencies, a base and a target.
+// If a dependency is set in both, the one from `target` is used, otherwise, the dependency from parent is used.
+// MergeDependencies(nil, child) = child, MergeDependencies(parent, nil) = parent
```
MorrisLaw, line 422:
```suggestion
// MergeDependencies(nil, child) = child, MergeDependencies(parent, nil) = parent, MergeDependencies(parent, child) = child
```

Is this correct too?

adamperlin: Assuming all **fields** of child are non-nil then yes, otherwise some fields may be taken from the parent.

PR #536, opened 2025-02-19
add mode field for artifacts

DannyBrito:

add mode field for artifacts

Fixes #


frontend/deb/template_rules.go
```diff
@@ -67,6 +67,88 @@ func (w *rulesWrapper) OverridePerms() fmt.Stringer {
 	artifacts := w.GetArtifacts(w.target)
 
 	var fixPerms bool
+	// also need to check for fields that are of type map[string]dalec.ArtifactConfig
+	// because they can also have a Mode field.
+
+	for _, cfg := range artifacts.Binaries {
+		if cfg.Mode != 0 {
```
MorrisLaw, line 74: Is there a way to D.R.Y. this up?

Something like create `artifactTypes` that's a slice of slices of type dalec.ArtifactConfig(?), iterate through those types then check if the mode field is there for each one (which itself could be it's own function).

Would remove a lot of this redundancy. Wdyt?

packaging/linux/deb/debroot.go
```diff
@@ -32,6 +32,18 @@ var patchHeader []byte
 //go:embed templates/debian_install_header.sh
 var debianInstall []byte
 
+var artifactConfigRefs = map[string]string{
```
cpuguy83, line 35: Seems like these would be better as just pure consts.

packaging/linux/deb/debroot.go
```diff
@@ -205,30 +217,123 @@ func fixupArtifactPerms(spec *dalec.Spec, target string, cfg *SourcePkgConfig) [
...
 	}
 
-	sorted = dalec.SortMapKeys(artifacts.Directories.GetState())
-	for _, name := range sorted {
-		cfg := artifacts.Directories.State[name]
-		if cfg.Mode.Perm() != 0 {
-			p := filepath.Join(basePath, "var/lib", name)
-			fmt.Fprintf(buf, "chmod %o %q\n", cfg.Mode.Perm(), p)
+	if artifacts.Binaries != nil {
+		for key, cfg := range artifacts.Binaries {
```
cpuguy83, line 265: In all of these cases we need to sort by key otherwise we'll end up busting the build cache potentially on every run due to randomness of maps.

targets/windows/handle_zip.go
```diff
@@ -150,13 +148,16 @@ func getZipLLB(worker llb.State, name string, artifacts llb.State, opts ...llb.C
...
 }
 
-func generateInvocationScript(binaries []string) *strings.Builder {
+func generateInvocationScript(binaries map[string]dalec.ArtifactConfig) *strings.Builder {
 	script := &strings.Builder{}
 	fmt.Fprintln(script, "#!/usr/bin/env sh")
 	fmt.Fprintln(script, "set -ex")
 	fmt.Fprintf(script, "/tmp/scripts/%s\n", buildScriptName)
-	for _, bin := range binaries {
+	for bin, config := range binaries {
```
cpuguy83, line 156: Needs sorting.

artifacts.go
```diff
@@ -100,6 +100,9 @@ type ArtifactConfig struct {
 	// Name is file or dir name to use for the artifact in the package.
 	// If empty, the file or dir name from the produced artifact will be used.
 	Name string `yaml:"name,omitempty" json:"name,omitempty"`
+	// Mode is the file mode to use for the artifact.
+	// Mode is represented as an fs.FileMode.
+	Mode fs.FileMode `yaml:"mode,omitempty" json:"mode,omitempty"`
```
cpuguy83, line 105: Need to add this to the artifact config info in website/docs.

cpuguy83 (changes requested):

artifacts.go
DannyBrito: the mode field was already specified in [docs](https://github.com/Azure/dalec/blob/main/website/docs/artifacts.md?plain=1#L42) Is that what you mean?

cpuguy83: Oh yep, you are correct!

cpuguy83, line 105: I think this should be called `permissions`

cpuguy83: I mean all the things.

```suggestion
	// Permissions is the file permissions to set on the artifact.
	// If not set, the default value will depend on the kind of artifact or the underlying artifact's already set permissions.
	Permissions fs.FileMode `yaml:"permissions,omitempty" json:"permissions,omitempty"`
```

cpuguy83: We use permissions everywhere except in `ArtifactDirConfig`, which we should probably change before 1.0

cpuguy83 (approved):

PR #537, opened 2025-02-19
Improve ergonomics for extension fields

cpuguy83:
This makes it easier to determine if the error is due to the specified field not existing or some other reason.
It also fixes a potential panic there were no extensions set.

### DIFF
--- a/load.go
+++ b/load.go
@@ -19,6 +19,8 @@ const (
 	// KeyDalecTarget is the key used the build arg name which may be used to read
 	// the target name.
 	KeyDalecTarget = "DALEC_TARGET"
+
+	parseModeIgnoreComments = 0
 )
 
 func knownArg(key string) bool {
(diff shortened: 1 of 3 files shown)


load.go
```diff
@@ -247,6 +335,8 @@ func (s *Spec) UnmarshalYAML(dt []byte) error {
 		return errors.New("expected exactly one yaml document")
 	}
 
+	// Remove extension nodes from the main AST and stor them so we can unmarshal
```
pmengelbert, line 338: typo: stor

cpuguy83: Fixed.

load.go
```diff
@@ -237,6 +239,61 @@ func LoadSpec(dt []byte) (*Spec, error) {
...
+
+	body, ok := doc.Body.(*ast.MappingNode)
+	if !ok {
+		return errors.Errorf("expected a mapping node, got %T", body)
+	}
+
+	var ext extensionFields
+	for _, v := range body.Values {
+		key := v.Key.String()
+		if !strings.HasPrefix(key, "x-") && !strings.HasPrefix(key, "X-") {
```
MorrisLaw, line 281: for my context, what's special about `x-` and `X-`?

load.go
```diff
@@ -237,6 +239,61 @@ func LoadSpec(dt []byte) (*Spec, error) {
 	return &spec, nil
 }
 
+// rawYAML is similar to json.RawMesage
```
MorrisLaw, line 242: typo:
```suggestion
// rawYAML is similar to json.RawMessage
```

load.go
```diff
@@ -237,6 +239,61 @@ func LoadSpec(dt []byte) (*Spec, error) {
...
+		if ext == nil {
+			ext = make(extensionFields)
+		}
+		ext[key] = rawYAML(v.Value.String())
+	}
+
+	if ext != nil {
+		*f = ext
+	}
+	return nil
```
MorrisLaw, line 294: It seems we eventually panic when the extensions are nil later on on line 350 when we unmarshal the extension nodes:

```go
s.extensions = ext
 if len(ext) == 0 {
 	panic("ext should not be empty")
 }
 ```
 
Why do we do that in the Spec.UnmarshalYAML function and not here for the individual extension fields UnmarshalYAML function? Is this because some fields might be nil and that's ok for later on when we unmarshal all of the extension nodes and just need to know if there are no extensions at all?

load.go
cpuguy83: Dalec does strict unmarshalling, so if there are unknown fields it errors out.
We have specifically carved out top-level fields that start with `x-` or `X-`, which is a common way to deal with adding extra stuff to yaml files.

So this allows people to add `x-whatever` to their spec and not have unmarshalling error out.

load.go
cpuguy83: In this function its not an error to not have anything to unmarshal.
In the spec, by the time we are doing that panic we have already collected a list of `x-` fields and so `ext` must not be empty.

load.go
cpuguy83: Fixed.

MorrisLaw (approved):

PR #543, opened 2025-02-25
deb: Fix patches when patch is in subdir

cpuguy83:
When a patch is in a subdir, the wrong path was used to extract the
patch from the source from.

Additionally, the call to `mountSources` was causing issues afer this
fix.
`mountSources` is just in this case is an optimization so to prevent
dpkg-source from having to re-extract data we already have available.
Since this is causing an issue I decided to just remove it for now.

I believe the two changes included in this PR are playing off each other (ie, one change requires a change in the other).
We'll need to look into this before including this optimization again.

### DIFF
--- a/test/azlinux_test.go
+++ b/test/azlinux_test.go
@@ -325,10 +325,40 @@ index 0000000..5260cb1
 +
 +echo "Added another new file"
 `)
-		src2Patch3Context := llb.Scratch().File(
-			llb.Mkfile(src2Patch3File, 0o600, src2Patch3Content),
-		)
-		src2Patch3ContextName := "patch-context"
+
+		src2Patch4Content := []byte(`
+diff --git a/file4 b/file4
+new file mode 100700
+index 0000000..5260cb1
+--- /dev/null
++++ b/file4
+@@ -0,0 +1,3 @@
++#!/usr/bin/env bash
++
++echo "Added yet another new file"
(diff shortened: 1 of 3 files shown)


test/azlinux_test.go
```diff
@@ -325,10 +325,40 @@ index 0000000..5260cb1
...
+
+		const src2Patch4File = "patches/patch4"
+		const src2Patch5File = "patches/patch5"
+		const patchContextName = "patch-context"
+
+		patchContext := llb.Scratch().
+			File(llb.Mkfile(src2Patch3File, 0o600, src2Patch3Content)).
+			File(llb.Mkdir("patches", 0o755)).
+			File(llb.Mkfile(src2Patch4File, 0o600, src2Patch4Content)).
+			File(llb.Mkfile(src2Patch5File, 0o600, src2Patch5Content))
```
DannyBrito, line 361: question, so here you are only building the llb definition that's why the file action's order does not matter? I am referring to seeing here first creating the file `src2Patch3File` and then the directory where this lives `patches`

cpuguy83: `src2Patch3` is not in `patches`, it is at the root of the context.
Order actually does matter here in that you can't create a file in a subpath that does not exist.

DannyBrito (approved):

PR #545, opened 2025-02-25
deb: Do not create conffiles

cpuguy83:
Historically, conffiles is used to help dpkg track configuration files so that user modifications don't get overwritten on upgrades.

Since debhelper v9 this is not needed for anything that goes into /etc as dpkg automatically tracks things under that path. Since we only put conf files into /etc we really don't need conffiles.

This resolves an issue where a spec lists a directory of configs in `artifacts.configFiles` instead of each individual config. With conffiles we end up listing the directory in there which will cause dpkg to error out on install.

### DIFF
--- a/test/azlinux_test.go
+++ b/test/azlinux_test.go
@@ -1326,6 +1326,28 @@ Environment="KUBELET_KUBECONFIG_ARGS=--bootstrap-kubeconfig=/etc/kubernetes/boot
 						},
 					},
 				},
+				"src3": {
+					Inline: &dalec.SourceInline{
+						Dir: &dalec.SourceInlineDir{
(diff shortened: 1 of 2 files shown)


test/azlinux_test.go
DannyBrito, line 1: should we have a negative test to check the conffiles do not get created?

MorrisLaw (approved):

test/azlinux_test.go
cpuguy83: conffiles is specific to the debian packaging and is strictly part of the package itself and looked at by `dpkg`.
I don't think a negative test would be useful here.

DannyBrito (approved):

PR #553, opened 2025-03-01
Struct tag linter

cpuguy83:
Example (this is from a commit I added temporarily: https://github.com/Azure/dalec/commit/5bc521734d090c919ea902137dd332fbfd35b091).

image

### DIFF
--- a/linters/yaml_json_tags.go
+++ b/linters/yaml_json_tags.go
@@ -1,0 +1,81 @@
+package linters
+
+import (
+	"go/ast"
+	"strings"
+
+	"golang.org/x/tools/go/analysis"
+)
+
+var YamlJSONTagsMatch = &analysis.Analyzer{
+	Name: "yaml_json_names_match",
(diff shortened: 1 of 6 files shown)


linters/yaml_json_tags.go
```diff
@@ -0,0 +1,82 @@
...
+		json = "json"
+	)
+
+	tag = strings.Trim(tag, "`")
+
+	var out [2]string
+	for tag != "" {
+		var key, value string
+		key, tag, _ = strings.Cut(tag, ":")
+		value, tag, _ = strings.Cut(tag, " ")
```
adamperlin, line 64: Can you please add a test case for this function and make sure it is fully robust to whitespace between `key:value` pairs?

adamperlin (changes requested):

adamperlin (approved):
Looks good to me pending the unrelated CI failures!

PR #557, opened 2025-03-07
Deb and multi-sources structure

### LINKED ISSUE #363: [BUG] jammy: Use of multi-origin sources leaves some room for issues with source names
Currently jammy uses multi-origin sources when building the deb packages.
For reference: multi-origin is the practice of splitting sources into separate tarballs and creating an empty "orig" tarball (see the deb packaging layout for more details here).

Multi-orig tarballs have stricter naming requirements and as such we've had to sanitize source names which will probably work most of the time, but really the sanitization we have is bug-prone.
e.g., a source named `foobar` and `foo-bar` will almost certainly collide when building a deb package with dalec b/c the sanitization we did just strips forbidden characters.

I suggest we switch this strategy to just merge all the sources into one tarball with the desired layout.

DannyBrito:

Fixes #363

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -62,7 +62,7 @@ target "runc-azlinux" {
         "DALEC_DISABLE_DIFF_MERGE" = DALEC_DISABLE_DIFF_MERGE
     }
     contexts = {
-      "dalec_frontend" = "target:frontend"
+        "dalec_frontend" = "target:frontend"
     }
     matrix = {
         distro = ["mariner2", "azlinux3"]
@@ -86,7 +86,7 @@ target "runc-jammy" {
         "DALEC_DISABLE_DIFF_MERGE" = DALEC_DISABLE_DIFF_MERGE
     }
     contexts = {
-      "dalec_frontend" = "target:frontend"
+        "dalec_frontend" = "target:frontend"
     }
     matrix = {
         tgt = ["deb", "container"]
@@ -124,9 +124,9 @@ target "test-fixture" {
     matrix = {
         f = DALEC_DISABLE_NESTED == "1" ? (
             ["http-src", "frontend", "local-context", "cmd-src-ref"]
-         ) : (
+        ) : (
             ["http-src", "frontend", "local-context", "cmd-src-ref", "nested"]
-         )
+        )
         tgt = ["mariner2/container"]
     }
     dockerfile = "test/fixtures/${f}.yml"
@@ -136,7 +136,7 @@ target "test-fixture" {
         "DALEC_DISABLE_DIFF_MERGE" = DALEC_DISABLE_DIFF_MERGE
     }
     contexts = {
-      "dalec_frontend" = "target:frontend"
+        "dalec_frontend" = "target:frontend"
     }
     target = tgt
 }
@@ -156,7 +156,7 @@ target "build" {
         "BUILDKIT_SYNTAX" = "dalec_frontend"
     }
     contexts = {
-      "dalec_frontend" = "target:frontend"
+        "dalec_frontend" = "target:frontend"
     }
     target = "${distro}/${tgt}"
     // only tag the container target
@@ -175,7 +175,7 @@ target "examples" {
         "BUILDKIT_SYNTAX" = "dalec_frontend"
     }
     contexts = {
-      "dalec_frontend" = "target:frontend"
+        "dalec_frontend" = "target:frontend"
     }
     target = "${distro}/container"
     dockerfile = "docs/examples/${f}.yml"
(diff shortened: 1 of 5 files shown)


packaging/linux/deb/debroot.go
```diff
@@ -147,8 +146,8 @@ func Debroot(ctx context.Context, sOpt dalec.SourceOpts, spec *dalec.Spec, worke
 				return in
```
cpuguy83, line 146: This whole block needs to be removed.
(I'd do it as a suggestion but GH won't let me select the starting from L145...).

This along with removing the `fix-sources` stuff in the rules file fixes the issue we discussed.

packaging/linux/deb/pkg.go
```diff
@@ -129,13 +134,24 @@ func BuildDeb(worker llb.State, spec *dalec.Spec, srcPkg llb.State, distroVersio
...
+// 		st := dalec.Tar(worker, src, "src.tar.gz", opts...)
+// 		res = res.File(llb.Mkdir("/result/"+key, 0o755)).File(
+// 			llb.Copy(st, "/", "/result/"+key))
+// 		out[key] = st
+// 	}
+
+// 	return out
+// }
+
+func singleDeb(worker llb.State, sources map[string]llb.State, opts ...llb.ConstraintsOpt) llb.State {
```
cpuguy83, line 151: I think it would be worthwhile to add a new function similar to `dalec.Tar` that takes a `map[string]llb.State` (or even just `map[string]Source`...) that handles all this.

Then we can mount each source in and tar it all up at once instead of creating this dependency chain like we have here.

cpuguy83:
@DannyBrito  Any updates on this?

DannyBrito:
two current main issues happening.
- Converting the sources to a single tar artifact, if there is a single src (also the dalec go mod generation is counted as extra src) then the dkpg or at whatever point of the tar extraction the src lets uncompress directly in the given directory so it would result in something like `/files-here` instead of something like `/<src-name>/files-here`.
- It seems like when we compress/uncompress if the src is a file it ends becoming a directory.

docker-bake.hcl
```diff
@@ -62,7 +62,7 @@ target "runc-azlinux" {
         "DALEC_DISABLE_DIFF_MERGE" = DALEC_DISABLE_DIFF_MERGE
     }
     contexts = {
-      "dalec_frontend" = "target:frontend"
+        "dalec_frontend" = "target:frontend"
```
cpuguy83, line 65: Weird that the spacing was messed up.

packaging/linux/deb/pkg.go
```diff
@@ -129,13 +139,44 @@ func BuildDeb(worker llb.State, spec *dalec.Spec, srcPkg llb.State, distroVersio
...
+					mounthPath = filepath.Join("/src", key, key)
+				}
+
+				llb.AddMount(mounthPath, state, mountOpts...).SetRunOption(ei)
+				if len(srcStates) == 1 {
+					break
+				}
+			}
+		}),
+		dalec.ShArgs("set -x ; tar -C /src -cvzf /tmp/st ."),
```
cpuguy83, line 174: Can remove `set -x`

test/fixtures/deb.yml
```diff
@@ -0,0 +1,78 @@
+# syntax=local/dalec/frontend:latest
```
cpuguy83, line 1: Can you remove this file?

packaging/linux/deb/debroot.go
```diff
@@ -136,20 +131,12 @@ func Debroot(ctx context.Context, sOpt dalec.SourceOpts, spec *dalec.Spec, worke
...
 		File(llb.Mkdir(filepath.Join(dir, "source"), 0o755), opts...).
-		With(func(in llb.State) llb.State {
-			if len(spec.Sources) == 0 {
-				return in
-			}
-			return in.
-				File(llb.Mkfile(filepath.Join(dir, "source/format"), 0o640, []byte(sourceFormat)), opts...).
-				File(llb.Mkfile(filepath.Join(dir, "source/options"), 0o640, []byte(sourceOptions)), opts...)
-		}).
+		File(llb.Mkfile(filepath.Join(dir, "source/format"), 0o640, []byte(sourceFormat)), opts...).
```
cpuguy83, line 139: Can we make sure there is an integration test case where there are no sources?

cpuguy83 (changes requested):

test/fixtures/deb.yml
DannyBrito: done

packaging/linux/deb/debroot.go
DannyBrito: There is one test at least that runs without sources. I encounter this as previously PR was failing on not having the upstream tar for dpkg, as we were skipping the taring if not sources where there. So I assumed that case was cover.
ref to test wtihout sources: https://github.com/Azure/dalec/blob/main/test/linux_target_test.go#L150

docker-bake.hcl
DannyBrito: just some nit to move spacing at same level as the other objects in this file

cpuguy83 (approved):

PR #558, opened 2025-03-07
Validate licenses included in artifacts

### LINKED ISSUE #547: [REQ] validate licenses are copied if a licence is specified in the source

~~Dalec should create licence files in /usr/share/licenses for known licenses from specs (not dependencies)~~

warning that says:
"source <name> has LICENSE file but is not included in artifacts"

example:
https://oci.dag.dev/layers/mcr.microsoft.com/azurelinux/base/nginx@sha256:f203d7e49ce778f8464f403d2558c5d7162b1b9189657c6b32d4f70a99e0fe83/usr/share/licenses/

bchuo:

Fixes #547 

### DIFF
--- a/load.go
+++ b/load.go
@@ -538,6 +538,10 @@ func (s Spec) Validate() error {
 		}
 	}
 
+	if s.Artifacts.Licenses == nil {
+		errs = append(errs, errMissingLicense)
+	}
+
 	if err := s.Image.validate(); err != nil {
 		errs = append(errs, errors.Wrap(err, "image"))
 	}
--- a/artifacts.go
+++ b/artifacts.go
@@ -4,6 +4,12 @@ import (
 	"io/fs"
 	"maps"
 	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+var (
+	errMissingLicense = errors.New("missing required license in artifacts")
 )
 
 // Artifacts describes all the artifacts to include in the package.


load.go
```diff
@@ -538,6 +538,10 @@ func (s Spec) Validate() error {
 		}
 	}
 
+	if s.Artifacts.Licenses == nil {
```
sozercan, line 541: i think we want to check if source includes LICENSE or LICENSE.md

bchuo: From what I gathered on build-defs, I don't see where a license is included explicitly. Do you mean searching through the aforementioned source to verify the path is correct?

cpuguy83: I don't think we can really do this validation here.
At best we can provide a warning message back to clients, but we can't do that from here.

bchuo: What's prohibiting us from doing that validation? Is it that we don't have visibility to the file system at this point?

bchuo: I guess in the same vein, what exactly constitutes a license and do we require it for every spec/does it exist for every project

cpuguy83: I think dalec itself is just not the right place to error out on a missing license file.

bchuo: Sounds like we should add this check to build-defs then?

cpuguy83: Yeah, I think so.

PR #564, opened 2025-03-11
Breaking: Do not include package name in lib artifact paths

### LINKED ISSUE #548: [REQ] Delete `ld.so.cache` in containers and dynamic linking woes

None

Right now, *.so files are places under /usr/lib/<package>. This is not ideal because AFAIK, `ld` will not look there by default. It will only look under /usr/lib and /lib and /usr/local/lib. We can modify this behavior, by adding a config file into `/etc/ld.so.conf.d/` containing

```
/usr/lib/<package>
```

so that `ld` looks in the right place (TBH, we should probably do this automatically if a spec contains libraries).
Despite this, the `ld` still doesn't find shared objects in `/usr/lib/<package>` because its cache `ld.so.cache` does not get updated/refreshed after updating the config.

The only way around this (AFAIK) is to create a bunch of symlinks under `/usr/lib` pointing to the files in `/usr/lib/<package>`

cpuguy83:
Libs should be able to be installed directly into `/usr/lib` rather than forcing everything to go into `/usr/lib/<package name>`.

Fixes #548


adamperlin (approved):
This LGTM

PR #575, opened 2025-03-17
windows: add ltsc2019 to list of default base images

cpuguy83:


targets/windows/handle_container.go
```diff
@@ -22,10 +22,21 @@ import (
 )
 
 const (
-	defaultBaseImage = "mcr.microsoft.com/windows/nanoserver:1809"
+	winltsc2019 = "mcr.microsoft.com/windows/nanoserver:ltsc2019"
```
cpuguy83, line 25: Note: ltsc2019 and 1809 are the same images.

DannyBrito: maybe add a comment about this?

adamperlin (commented):
@cpuguy83 it looks like there are some CI failures, possibly to do with an image config issue?

cpuguy83:
> @cpuguy83 it looks like there are some CI failures, possibly to do with an image config issue?

Failures are due to using `SingleRef()` in tests where we are no longer producing just a single ref since there are multiple images in the result.

PR #584, opened 2025-03-25
azlinux: use weak dependency for prebuilt-ca-certificates

cpuguy83:
As it turns out, the previous "fix" was causing dnf to *always* choose `ca-certificates`, presumably because it is what it sees first in the package repo.

By using a weak dependency we can allow the built package to depend on ca-certificates and ignore the prebuilt-ca-certificates but by default install the prebuilt ones.

The end result of the previous case is we end up with bash and some other tools int he final image, which is undesirable.

ref: https://docs.fedoraproject.org/en-US/packaging-guidelines/WeakDependencies/

### DIFF
--- a/packaging/linux/rpm/template.go
+++ b/packaging/linux/rpm/template.go
@@ -35,6 +35,7 @@ BuildArch: noarch
 {{- .Provides -}}
 {{- .Replaces -}}
 {{- .Requires -}}
+{{- .Recommends -}}
 
 %description
 {{.Description}}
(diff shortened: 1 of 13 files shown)


MorrisLaw (approved):

cpuguy83:
I updated this to skip the test that checks if `ca-certificates` can be installed for mariner2 since mariner2 does not support weak dependencies.
We just can't support it there.

cpuguy83:
If it becomes a problem, and I doubt it would, we can do more like manually walking the dependency tree to see if we should install the prebuilt certs or not.

PR #585, opened 2025-03-25
Initial commit to use srpms to install build deps

### LINKED ISSUE #570: Simplify rpm build deps install
Right now when we build an rpm we first generate an RPM that has the build deps of the spec as runtime deps and then we install that package which triggers all the deps to be installed.

As it turns out we can generate a source package and use `dnf install --builddeps <srcpkg>.rpm`
This would in theory save a bit of time since it can skip generating the intermediate package (and intermediate source package!) and should make errors easier to track down.

bchuo:
**What this PR does / why we need it**: Currently the process to install build deps involves extra hoops to jump through. We can simplify it using dnf install with an srpm instead of a rpm

Fixes #570 


bchuo:
Closing because this solution doesn't seem to be feasible as per offline discussion

PR #598, opened 2025-04-09
Add support for cargo dep generation

### LINKED ISSUE #587: [REQ] Add support for cargo home generator

None

Similar to go module's, we need to be able to generate sources suitable for CARGO_HOME to simplify doing non-networked builds with rust+cargo

bchuo:

Fixes #587 


cpuguy83:
It looks like the issue on Ubuntu is related to how we resolve dependencies.
Mainly that we do `dpkg -i` expecting it to fail and then use `aptitude` to resolve all failed dependencies.
This is causing `rust-all` to be turned into `rust-1.74-all` (on Noble) which does not include the changes to add rust binaries to /usr/bin.

This change looks like it resolves the issue:


```diff
diff --git a/targets/linux/deb/distro/install.go b/targets/linux/deb/distro/install.go
index cca4cda..d3076d5 100644
--- a/targets/linux/deb/distro/install.go
+++ b/targets/linux/deb/distro/install.go
@@ -80,6 +80,10 @@ apt autoclean -y
rm -rf /var/lib/apt/lists/partial/*
apt update
+if apt install -y ${1}; then
+	exit 0
+fi
+
if ! command -v aptitude > /dev/null; then
	needs_cleanup=1
	apt install -y aptitude
```

fyi @adamperlin
This seems like an OK change.
The reason we do the aptitude dance is so that constraints can be processed as expected instead of erroring like `apt`.
In our case here apt doesn't error because there are no constraints.

There is still the case where there are constraints involved and the need to fallback to aptitude that this does not deal with.

cpuguy83 (approved):

PR #611, opened 2025-04-26
deb: Upgrade before install container packages

cpuguy83:
This makes sure the container is fully up to
date in case the container image is stale.

### DIFF
--- a/targets/linux/deb/distro/install.go
+++ b/targets/linux/deb/distro/install.go
@@ -3,6 +3,7 @@ package distro
 import (
 	"context"
 	"path/filepath"
+	"strconv"
 
 	"github.com/Azure/dalec"
 	"github.com/Azure/dalec/packaging/linux/deb"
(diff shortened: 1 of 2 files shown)


targets/linux/deb/distro/install.go
```diff
@@ -80,6 +81,10 @@ apt autoclean -y
 rm -rf /var/lib/apt/lists/partial/*
 apt update
 
+if [ "${DALEC_UPGRADE}" = "true" ]; then
+		apt dist-upgrade -y
```
adamperlin, line 85:
```suggestion
	apt dist-upgrade -y
```

targets/linux/deb/distro/install.go
```diff
@@ -164,7 +170,7 @@ func (d *Config) InstallBuildDeps(sOpt dalec.SourceOpts, spec *dalec.Spec, targe
 			return in.Run(
 				dalec.WithConstraints(opts...),
 				customRepos,
-				InstallLocalPkg(pkg, opts...),
+				InstallLocalPkg(pkg, false, opts...),
```
adamperlin, line 173: When are we expecting `upgrade=true` to be used?

cpuguy83: Creating the container.

targets/linux/deb/distro/install.go
cpuguy83: Added.

targets/linux/deb/distro/install.go
adamperlin: Ok, that makes sense!

adamperlin: Oh I see, I'd missed the diff in `container.go`

adamperlin (approved):

PR #613, opened 2025-04-28
Add chown support for symlinks

### LINKED ISSUE #599: [REQ] Add chown support for symlinks

None

Symlinks support setting uid/gid, we should support this in the spec.

bchuo:

Fixes #599 


bchuo:
Not sure where exactly to add tests here, just realized I don't have any in the PR

cpuguy83 (changes requested):
I think we also need this for the symlink artifacts type that we have.
That would need to go into the rpm and deb implementations.

Can you also add docs and tests?

cpuguy83:
> Not sure where exactly to add tests here, just realized I don't have any in the PR

Under tests/linux_target_test.go there should be a test like `t.Run("container", ...)`
I think that's an appropriate place.
It could go into the "tests" section of the spec (along with any relevant symlink additions to `artifacts` and `image`).

Since we can't use the buildkit API to inspect symlinks (the API follows the symlink) you'll need to use a bash command in the tests section to actually look at the symlink.

packaging/linux/deb/debroot.go
```diff
@@ -185,13 +185,42 @@ func Debroot(ctx context.Context, sOpt dalec.SourceOpts, spec *dalec.Spec, worke
 
 	if len(artifacts.Links) > 0 {
 		buf := bytes.NewBuffer(nil)
+		allLinksBuf := bytes.NewBuffer(nil)
```
cpuguy83, line 188: I'm thinking because the links file doesn't seem to support setting ownership, we probably need to move things that need ownership changes into the generic install file where we can create the link then chown in a shell script.

packaging/linux/deb/template_rules.go
```diff
@@ -108,7 +118,7 @@ func (w *rulesWrapper) OverridePerms() fmt.Stringer {
 		// our extra script.
 		b.WriteString("override_dh_fixperms:\n")
 		b.WriteString("\tdh_fixperms\n")
-		b.WriteString("\tdebian/dalec/fix_perms.sh\n\n")
+		b.WriteString("\tDESTDIR=debian/$(shell dh_listpackages) debian/dalec/fix_perms.sh\n\n")
```
cpuguy83, line 121: Can you explain this change?

packaging/linux/deb/debroot.go
```diff
@@ -277,6 +278,26 @@ func fixupArtifactPerms(spec *dalec.Spec, target string, cfg *SourcePkgConfig) [
...
+	if len(artifacts.Links) > 0 {
+		fmt.Fprintf(buf, "# Set ownership for artifact symlinks\n")
+		for _, link := range artifacts.Links {
+			if link.UID != 0 || link.GID != 0 {
+				fmt.Fprintf(buf, "chown -h %d:%d \"$DESTDIR%s\"\n", link.UID, link.GID, link.Dest)
+			}
+		}
+	}
+
+	if spec.Image != nil && spec.Image.Post != nil {
```
cpuguy83, line 290: This doesn't belong here.
The image post stuff is not part of the packaging and is instead used to modify the container after it is built.

cpuguy83 (changes requested):

packaging/linux/deb/template_rules.go
bchuo: Copilot suggested this change since it provides absolute path resolution with `destdir`, and the `shell dh_listpackages` command dynaically retrieves the actual binary package name, so it would work even if the package name is different from the source package name, or if multiple binary packages are produced from one source package

packaging/linux/rpm/template.go
```diff
@@ -829,7 +865,13 @@ func (w *specWrapper) Files() fmt.Stringer {
 	}
 
 	for _, l := range artifacts.Links {
-		fmt.Fprintln(b, l.Dest)
+		if l.UID != 0 || l.GID != 0 {
+			// Use %attr to specify ownership, with "-" for mode to preserve symlink's mode
+			fmt.Fprintf(b, "%%attr(-, %d, %d) %s\n", l.UID, l.GID, l.Dest)
```
cpuguy83, line 870: `%attr` does not support numeric id's for user and group.
I think we'll need to make these strings (in the spec) and expect people to put names in rather than ID's.

packaging/linux/rpm/template.go
```diff
@@ -689,6 +722,9 @@ func (w *specWrapper) Install() fmt.Stringer {
 	for _, l := range artifacts.Links {
 		fmt.Fprintln(b, "mkdir -p", filepath.Dir(filepath.Join("%{buildroot}", l.Dest)))
 		fmt.Fprintln(b, "ln -sf", l.Source, "%{buildroot}/"+l.Dest)
+		if l.UID != 0 || l.GID != 0 {
+			fmt.Fprintf(b, "chown -h %d:%d %q\n", l.UID, l.GID, "%{buildroot}/"+l.Dest)
```
cpuguy83, line 726: `%attr` under the `%files` section would take care of setting ownership.

cpuguy83 (changes requested):

helpers.go
```diff
@@ -367,12 +367,18 @@ func InstallPostSymlinks(post *PostInstall, rootfsPath string) llb.RunOption {
...
-			newpaths := post.Symlinks[oldpath].Paths
+			cfg := post.Symlinks[oldpath]
+			newpaths := cfg.Paths
 			sort.Strings(newpaths)
 
 			for _, newpath := range newpaths {
 				fmt.Fprintf(buf, "mkdir -p %q\n", filepath.Join(rootfsPath, filepath.Dir(newpath)))
 				fmt.Fprintf(buf, "ln -s %q %q\n", oldpath, filepath.Join(rootfsPath, newpath))
+				if cfg.UID != "" || cfg.GID != "" {
+					llb.AddMount("/etc/passwd", ei.State, llb.SourcePath(filepath.Join(rootfsPath, "/etc/passwd"))).SetRunOption(ei)
```
cpuguy83, line 378: Looking at this function (`InstallPostSymlinks`), it is being used as an option to`llb.Run`, which is mounting in the new rootfs to `rootfsPath`.
Unfortunately this means these mounts won't work because those paths don't exist until after `llb.Run` is actually executed.

Intead of having `InstallPostSymlinks` return an `llb.RunOption`, I think it should return an `llb.StateOption`.

An `llb.StateOption` takes as input an `llb.State` and outputs a new `llb.State`.
So in this case the input would be just the `llb.State` that represents the container filesystem that we are adding symlinks to.
We'd still need the worker image, though, so that we can make sure we have access to `ln` and `chown`.

So basically we need to change the callers to do something like `container.With(dalec.InstallPostSymlinks(post, worker, opts...)` where:

1. `container` is the `llb.State` we are adding symlinks to
2. `worker` is the `llb.State` of the worker image
3. `opts...` are whatever `llb.ConstraintsOpt`s that need to be passed through
4. `InstallPostSymlinks would return an `llb.StateOption`, which is what the `With()` function accepts.

Then we can build up our own `llb.Run` internally here, something like:

```go
  // other setup...
  script := llb.File(llb.Mkfile(...)) // create the script file from buf
  return llb.Run(
    ShArgs("/tmp/add_symlink.sh"),
    llb.AddMount(script, "/tmp/add_symlink.sh", llb.SourcePath(...)),
    llb.AddMount("/etc/group", in, llb.SourcePath("/etc/group"),
    // Same for /etc/passwd
  ).AddMount("/tmp/rootfs", in)
}
```

cpuguy83 (changes requested):

packaging/linux/deb/template_rules.go
```diff
@@ -99,7 +108,8 @@ func (w *rulesWrapper) OverridePerms() fmt.Stringer {
 		checkPerms(artifacts.Libexec) ||
 		checkPerms(artifacts.DataDirs) ||
 		checkDirPerms(artifacts.Directories.GetConfig()) ||
-		checkDirPerms(artifacts.Directories.GetState())
+		checkDirPerms(artifacts.Directories.GetState()) ||
+		checkSymlinkOwnership(artifacts.Links)
```
cpuguy83, line 112: Looks like this (and the function being called) is leftover from a previous iteration.

```suggestion
		checkDirPerms(artifacts.Directories.GetState())
```

packaging/linux/deb/template_rules.go
```diff
@@ -89,6 +89,15 @@ func (w *rulesWrapper) OverridePerms() fmt.Stringer {
...
 
+	checkSymlinkOwnership := func(links []dalec.ArtifactSymlinkConfig) bool {
+		for _, link := range links {
+			if link.User != "" || link.Group != "" {
+				return true
+			}
+		}
+		return false
+	}
+
```
cpuguy83, line 100:
```suggestion
```

cpuguy83 (approved):
LGTM

I pushed 2 new commits:

1. Add an additional test case for users not being created as part of the spec
2. Update docs to reflect UID/GID is now user/group.

Can we take as a follow-up to document the decisions for the specifics in the rpm/deb implementations?

PR #615, opened 2025-04-29
Fix issue with rpm spec provides/replaces

cpuguy83:
The rpm spec was using the wrong field for "provides" and the wrong rpm spec tag for "replaces".
This addresses that.
The 2nd commit also handles the case where version constraints are provided with args.


packaging/linux/rpm/template_test.go
```diff
@@ -890,3 +890,43 @@ func TestTemplate_DisableStrip(t *testing.T) {
...
+func TestTemplate_Provides(t *testing.T) {
+	spec := &dalec.Spec{
+		Provides: map[string]dalec.PackageConstraints{
+			"test-provides": {},
+		},
+	}
+
+	w := &specWrapper{Spec: spec}
+	got := w.Provides().String()
+	want := "Provides: test-provides\n\n"
```
pmengelbert, line 903: I love these variable names

pmengelbert (approved):
LGTM

PR #617, opened 2025-04-29
Add workflow to retag images used in CI

cpuguy83:
This should help prevent CI failures due to DockerHub rate limits.
But after this is merged will require a follow-up to change these refs over.


pmengelbert (approved):
Great idea, LGTM

PR #618, opened 2025-04-29
Add custom marshalling for changelog date

cpuguy83:
This is mostly so that when marshaling the
changelog, the date is formatted in a way that we
are expecting and not a full RFC 3339 date-time.

### DIFF
--- a/spec_test.go
+++ b/spec_test.go
@@ -16,7 +16,6 @@ func TestDate(t *testing.T) {
 	assert.NilError(t, err)
 
 	d := Date{Time: expectTime}
-	assert.NilError(t, err)
 	assert.Check(t, cmp.Equal(d.Format(time.DateOnly), expect))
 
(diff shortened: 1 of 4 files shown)


spec_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"gotest.tools/v3/assert/cmp"
+)
+
+func TestDate(t *testing.T) {
+	expect := "2023-10-01"
+	expectTime, err := time.Parse(time.DateOnly, expect)
+	assert.NilError(t, err)
+
+	d := Date{Time: expectTime}
+	assert.NilError(t, err)
```
DannyBrito, line 19: Is this assertion needed? I don't see we are recapturing `err` after last assertion.
```suggestion
```

cpuguy83: Fixed.

DannyBrito (approved):

PR #622, opened 2025-05-01
windowscross: set GOOS=windows by default

cpuguy83:


website/docs/targets.md
```diff
@@ -175,3 +175,70 @@ targets:
...
+
+```yaml
+build:
+  steps:
+    - command: |
+        go build -o _output/bin/dalec_example.exe
+```
+
+You can use the built-in `TARGETOS` build-arg to determine if the build is targeting Windows or not.
+Alaternatively you can use the built-in `DALEC_TARGET` build-arg to determine the target being built.
```
adamperlin, line 193:
```suggestion
Alternatively you can use the built-in `DALEC_TARGET` build-arg to determine the target being built.
```

website/docs/targets.md
```diff
@@ -175,3 +175,70 @@ targets:
...
+          go build -o _output/bin/dalec_example
+        fi
+```
+
+Since `windowscross` is intended for cross-compilation, the environment has the
+following env vars set by default:
+
+- `GOOS=windows` - ensures that by default `go build` produces a Windows binary
+
+Theis can be overriden in your spec by either setting them in the `env` section
```
adamperlin, line 226:
```suggestion
This can be overridden in your spec by either setting them in the `env` section
```

adamperlin (approved):
Just a few comments on docs typos

website/docs/targets.md
cpuguy83: Fixed.

website/docs/targets.md
cpuguy83: Fixed.

PR #624, opened 2025-05-05
Allow target-specific provides/conflicts/replaces

cpuguy83:
Packages in different targets may have different names or possibly versions. This allows the user to specify a target-specific provides/conflicts/replaces.

### DIFF
--- a/website/docs/spec.md
+++ b/website/docs/spec.md
@@ -101,7 +101,7 @@ conflicts:
 
 #### provides
 
-The `provides` field is used to define the packages that are provided by this package package. This is an optional field.
+The `provides` field is used to define the packages that are provided by this package. This is an optional field.
 This is useful when the package name may not align with a more common name for the package or when the package is a
 replacement for another package.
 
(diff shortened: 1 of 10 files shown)


DannyBrito:
Do we have this on the docs? if not could you please add examples of the usage of these?

cpuguy83:
Added docs, apparently we didn't have these in the docs at all.

website/docs/spec.md
```diff
@@ -81,6 +81,54 @@ website: https://github.com/foo/bar
...
+  bar:
+    # Add version constraints to the conflicting package
+    version:
+      - ">=1.0.0"
+      - "<2.0.0"
+```
+
+#### provides
+
+The `provides` field is used to define the packages that are provided by this package package. This is an optional field.
```
DannyBrito, line 104: double package?
```suggestion
The `provides` field is used to define the packages that are provided by this package. This is an optional field.
```

cpuguy83: Fixed.

DannyBrito (approved):
lgtm

PR #625, opened 2025-05-05
Add persistent go module cache

cpuguy83:
This keeps a shared persistent go module cache across all builds. We still extract just the modules we need for the individual build but but now we can re-use already fetched modules just like one would expect from a go module cache.

Unfortunately there is no good way to automate testing this reliably due to buildkit caching.
After #621 is in, we can address this potentially by adding an ignore context for the gomod fetching so that a test will run that piece of code regardless of buildkit caching.

### DIFF
--- a/generator_gomod.go
+++ b/generator_gomod.go
@@ -9,6 +9,9 @@ import (
 
 const (
 	gomodCacheDir = "/go/pkg/mod"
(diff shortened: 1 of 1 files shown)


MorrisLaw (approved):
💯

PR #629, opened 2025-05-13
Add support for bazel caches similar to the gobuild cache.

cpuguy83:
This allows bazel users to keep a cache of bazel artifacts that
persists between builds and can even be shared between different
projects.

One slight difference here is we are not automatically injecting a bazel
cache at any point, it must always be manually added.
We may change this in the future but as I understand it, it will
actually be difficult to detect bazel anyway in most builds due to bazel
either not being available in the package manager or the *right* version
of bazel not being available.

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -29,11 +29,14 @@ type CacheConfig struct {
 	// GoBuild specifies a cache for Go's incremental build artifacts.
 	// This should speed up repeated builds of Go projects.
 	GoBuild *GoBuildCache `json:"gobuild,omitempty" yaml:"gobuild,omitempty" jsonschema:"oneof_required=gobuild"`
+	// Bazel specifies a cache for bazel builds.
+	Bazel *BazelCache `json:"bazel,omitempty" yaml:"bazel,omitempty" jsonschema:"oneof_required=bazel-local"`
 }
 
 type CacheInfo struct {
 	DirInfo CacheDirInfo
 	GoBuild GoBuildCacheInfo
+	Bazel   BazelCacheInfo
 }
 
 type CacheDirInfo struct {
@@ -93,6 +96,16 @@ func (c *CacheConfig) ToRunOption(distroKey string, opts ...CacheConfigOption) l
 		}))
 	}
 
+	if c.Bazel != nil {
+		return c.Bazel.ToRunOption(distroKey, BazelCacheOptionFunc(func(info *BazelCacheInfo) {
+			var cacheInfo CacheInfo
+			for _, opt := range opts {
+				opt.SetCacheConfigOption(&cacheInfo)
+			}
+			*info = cacheInfo.Bazel
+		}))
+	}
+
 	// Should not reach this point
 	panic("invalid cache config")
 }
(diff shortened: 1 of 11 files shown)


helpers.go
```diff
@@ -628,3 +633,31 @@ func HasGolang(spec *Spec, targetKey string) bool {
 	}
 	return false
 }
+
+var bzlVerRegex = sync.OnceValue(func() *regexp.Regexp {
+	return regexp.MustCompile(`^bazel[0-9]+`)
+})
+
+func HasBazel(spec *Spec, targetKey string) bool {
```
krinkinmu, line 641: We install Bazelisk as part of the build script instead of installing Bazel from Azure Linux repos. That's because installing Bazel from Azure Linux repos is fundamentally broken. Bazel-based projects are often pinned to a particular version of Bazel (and Envoy is one of such projects), Azure Linux only has a fixed and quite old version of Bazel (6.5).

Bazelisk, on the other hand, looks at what version of Bazel the project needs and downloads it in runtime. Bazelisk is also not available from Azure Linux repos, so we have to download it at runtime ourselves. That's also a rather typical setup for Bazel where you just download Bazelisk binary and install it manually instead of relying on packet managers.

We tried to reach out to the Azure Linux team both by email and via a ticket in their repo to improve the situation a bit and either improve their version support or to add Bazelisk, but got no response so far (like at all, it almost seems that we are reaching them in a wrong way somehow, because we never actrually got any responses from them).

As a result, you can't really depend on dependencies to detect Bazel. I would also argue that you shouldn't auto-detect and auto-inject Bazel cache IMO.

We could of course add bazelisk to thew Dalec specs, like we did with iptables, but we didn't get around it and in general supporting Bazel is somewhat outside of our primary area of expertise.

cpuguy83: Thanks.
I've removed this.

I've also pulled out the socket stuff for now as there are some issues that need to be addressed.

test/cache_test.go
```diff
@@ -71,6 +74,11 @@ func testArtifactBuildCacheDir(ctx context.Context, t *testing.T, cfg targetConf
 		if c.GoBuild != nil {
 			return "${GOCACHE}"
 		}
+		if c.Bazel != nil {
+			// There is no good way to determine the bazel cache dir
+			// So just hardcode this for now
+			return "/tmp/dalec/bazel-local-cache"
```
DannyBrito, line 80: not sure if better but maybe we could a helper function in `helpers.go` or other place that resolves to this value? so can be set one and reused also in `cache.go`?

cpuguy83: I honestly just didn't want to expose it as a public API.

test/target_almalinux_test.go
```diff
@@ -21,7 +21,8 @@ func TestAlmalinux9(t *testing.T) {
 			},
 			ListExpectedSignFiles: azlinuxListSignFiles("el9"),
 			PackageOverrides: map[string]string{
-				"rust": "rust cargo",
+				"rust":  "rust cargo",
+				"bazel": "",
```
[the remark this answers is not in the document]
cpuguy83: The way the code works, if there is no match (ie the package is not in the map) it assumes it can just use the requested package (ie "bazel").
If there is a match and the match is empty it signals that the package is not available.

cache.go
```diff
@@ -102,8 +115,19 @@ func (c *CacheConfig) validate() error {
...
+	}
+	if c.GoBuild != nil {
+		count++
+	}
+	if c.Bazel != nil {
+		count++
+	}
+
+	if count != 1 {
+		return fmt.Errorf("invalid cache config: exactly one of (dir, gobuild, bazel) must be set")
```
DannyBrito, line 130: question: this message sounds like you can only use one cache for spec but I think that is not the case, right? It is really referring that on entry is specific to one type of cache? if so, could we add some sort not in the docs and maybe example of multiple cache usage?

cpuguy83: Note this is for a single cache config instance.
The higher level validation wraps this error.

cpuguy83: Updated docs to make this more clear.

test/target_almalinux_test.go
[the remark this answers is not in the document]
cpuguy83: Changed this to use a constant to make this more clear.

DannyBrito (approved):
lgtm at least on syntax and logic.

PR #634, opened 2025-05-17
Allow container builds to use pre-built pkgs defined by target

### LINKED ISSUE #597: [REQ] Add targets to create a container from dir of packages

New feature

Building a container without having to build a package first (because we already have it in the context) will allow users split package builds from container builds.
Internally, this is kind of already how we build the container.
First we build the package then we hand the container builder the llb state with the packages.
We can just expose this functionality as a target.

MorrisLaw:
This PR adds functionality that allows a user to pass in a pre-built package via a target and then uses that same package when building the container. Allows for users to separate package builds from container builds.

Fixes #597

### DIFF
--- a/targets/linux/deb/distro/install.go
+++ b/targets/linux/deb/distro/install.go
@@ -8,6 +8,7 @@ import (
 	"github.com/Azure/dalec"
 	"github.com/Azure/dalec/packaging/linux/deb"
 	"github.com/moby/buildkit/client/llb"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/pkg/errors"
 )
 
@@ -127,7 +128,11 @@ aptitude install -y -f -o "Aptitude::ProblemResolver::Hints::=reject ${pkg_name}
 	})
 }
 
-func (d *Config) InstallBuildDeps(sOpt dalec.SourceOpts, spec *dalec.Spec, targetKey string, opts ...llb.ConstraintsOpt) llb.StateOption {
+func (d *Config) InstallBuildDeps(ctx context.Context, client gwclient.Client, spec *dalec.Spec, sOpt dalec.SourceOpts, targetKey string, opts ...llb.ConstraintsOpt) llb.StateOption {
+	return d.installBuildDeps(sOpt, spec, targetKey, opts...)
+}
+
+func (d *Config) installBuildDeps(sOpt dalec.SourceOpts, spec *dalec.Spec, targetKey string, opts ...llb.ConstraintsOpt) llb.StateOption {
 	return func(in llb.State) llb.State {
 		buildDeps := spec.GetBuildDeps(targetKey)
 		if len(buildDeps) == 0 {
--- a/targets/linux/distro_handler.go
+++ b/targets/linux/distro_handler.go
@@ -43,6 +43,9 @@ type DistroConfig interface {
 	// Some distros may need to pass in a separate worker before mounting the target container.
 	RunTests(ctx context.Context, client gwclient.Client, worker llb.State, spec *dalec.Spec, sOpt dalec.SourceOpts, ctr llb.State,
 		targetKey string, opts ...llb.ConstraintsOpt) (gwclient.Reference, error)
+
+	// InstallBuildDeps installs the build dependencies for a given spec.
+	InstallBuildDeps(ctx context.Context, client gwclient.Client, spec *dalec.Spec, sOpt dalec.SourceOpts, targetKey string, opts ...llb.ConstraintsOpt) llb.StateOption
 }
 
 func BuildImageConfig(ctx context.Context, sOpt dalec.SourceOpts, spec *dalec.Spec, platform *ocispecs.Platform, targetKey string) (*dalec.DockerImageSpec, error) {
@@ -90,17 +93,35 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
 				return nil, nil, err
 			}
 
-			pg := dalec.ProgressGroup(spec.Name)
-			pc := dalec.Platform(platform)
+			var opts []llb.ConstraintsOpt
+			opts = append(opts, dalec.ProgressGroup(spec.Name))
+			opts = append(opts, dalec.Platform(platform))
 
-			worker, err := c.Worker(sOpt, pg, pc)
+			worker, err := c.Worker(sOpt, opts...)
 			if err != nil {
 				return nil, nil, err
 			}
 
-			deb, err := c.BuildPkg(ctx, client, worker, sOpt, spec, targetKey, pg, pc)
+			// We need to get the build context from the client.
+			inputs, err := client.Inputs(ctx)
 			if err != nil {
-				return nil, nil, err
+				return nil, nil, fmt.Errorf("error getting llb state from client with provided build context: %w", err)
+			}
+
+			// If we have a build context named after the target, then we should use that pre-built
+			// package when building the container.
+			var pkgSt llb.State
+			if input, ok := inputs[targetKey]; ok {
+				pkgSt = input
+				opts = append(opts, dalec.ProgressGroup(fmt.Sprintf("Using pre-built package from %s context", targetKey)))
+				// ensure that worker state has the correct dependencies installed
+				worker = worker.With(c.InstallBuildDeps(ctx, client, spec, sOpt, targetKey, opts...))
+			} else {
+				var err error
+				pkgSt, err = c.BuildPkg(ctx, client, worker, sOpt, spec, targetKey, opts...)
+				if err != nil {
+					return nil, nil, err
+				}
 			}
 
 			img, err := BuildImageConfig(ctx, sOpt, spec, platform, targetKey)
@@ -108,12 +129,12 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
 				return nil, nil, err
 			}
 
-			ctr, err := c.BuildContainer(ctx, client, worker, sOpt, spec, targetKey, deb, pg, pc)
+			ctr, err := c.BuildContainer(ctx, client, worker, sOpt, spec, targetKey, pkgSt, opts...)
 			if err != nil {
 				return nil, nil, err
 			}
 
-			ref, err := c.RunTests(ctx, client, worker, spec, sOpt, ctr, targetKey, pg, pc)
+			ref, err := c.RunTests(ctx, client, worker, spec, sOpt, ctr, targetKey, opts...)
 			return ref, img, err
 		})
 	}
--- a/test/linux_target_test.go
+++ b/test/linux_target_test.go
@@ -164,6 +164,12 @@ func testLinuxDistro(ctx context.Context, t *testing.T, testConfig testLinuxConf
 		})
 	})
 
+	t.Run("prebuilt packages", func(t *testing.T) {
+		t.Parallel()
+		ctx := startTestSpan(ctx, t)
+		testPrebuiltPackages(ctx, t, testConfig)
+	})
+
 	t.Run("test-dalec-empty-artifacts", func(t *testing.T) {
 		t.Parallel()
 		ctx := startTestSpan(ctx, t)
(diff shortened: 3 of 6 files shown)


targets/linux/distro_handler.go
```diff
@@ -90,30 +93,48 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
...
+			inputs, err := client.Inputs(ctx)
 			if err != nil {
-				return nil, nil, err
+				return nil, nil, fmt.Errorf("error getting llb state from client with provided build context: %w", err)
+			}
+
+			// If we have a build context named after the target, then we should use that pre-built
+			// package when building the container.
+			var pkgSt llb.State
+			if input, ok := inputs[targetKey]; ok {
```
cpuguy83, line 114: `sOpt.GetContext`
This makes sure it obeys include/exclude filters.
Note this can return a `nil` state, which means no such context was set.

cpuguy83: I think `targetKey` is not the best name to use here.
Maybe something like `pkg-dir`?
Maybe it should include the `targetKey` in there, but then it gets more complicated to use it since the client needs to derive the target to setup the build.

Or perhaps either `targetKey+"-pkg-dir"` OR `pkg-dir`, with dalec preferring the former if it is set.

Another route would be to set it to the name of the build target (e.g. azlinux3/rpm) that produced the output.
Each target (e.g. `azlinux3`) can specify the name they are expecting to lookup from the inputs.

Anyway, naming is hard, just some suggestions here.

targets/linux/deb/distro/install.go
```diff
@@ -127,7 +128,11 @@ aptitude install -y -f -o "Aptitude::ProblemResolver::Hints::=reject ${pkg_name}
 	})
 }
 
-func (d *Config) InstallBuildDeps(sOpt dalec.SourceOpts, spec *dalec.Spec, targetKey string, opts ...llb.ConstraintsOpt) llb.StateOption {
+func (d *Config) InstallBuildDeps(ctx context.Context, client gwclient.Client, spec *dalec.Spec, sOpt dalec.SourceOpts, targetKey string, opts ...llb.ConstraintsOpt) llb.StateOption {
```
cpuguy83, line 131: I'm not sure I understand these changes to `InstallBuildDeps`, can you explain it?

cpuguy83 (changes requested):

targets/linux/deb/distro/install.go
MorrisLaw: In this new code, I am skipping `BuildPkg` if it's determined that a package already exists. By skipping that logic though, I am not installing the build deps. But now that I think about it, are the build deps that get installed still necessary if we're re-using an existing package?

Maybe I actually don't need the `InstallBuildSteps` and can skip that particular change altogether... But I assumed that the container might still need the build dependencies (unless that's only specific to building the pkg? which might actually be the case).

So maybe I should assume that the existing pkg already has the build dependencies it needs? If so, then the InstallBuildDeps changes here will go away and we'd just use whatever the worker image is for that particular distro.

e.g.

```go
worker, err := c.Worker(sOpt, pg, pc)
	if err != nil {
	return nil, nil, err	
```

cpuguy83: Yeah you should only need to install build deps when actually building the package.

DannyBrito:
how would this look like when calling for the build?
something like `docker build -f <spec> --target="<path-to-rpm>" .`?

MorrisLaw:
> how would this look like when calling for the build? something like `docker build -f <spec> --target="<path-to-rpm>" .`?

Good question. I think something like this would be the workflow:

1. Build the pkg first:
`docker buildx build --file=spec.yaml --target=rpm --output=type=local,dest=./pkgs .`

2. Build the container using the prebuilt pkg:
`docker buildx build --file=spec.yaml --target=container --build-context=rpm-pkg=./pkgs .`

If you look in the test, it's also how I'm including it: `withBuildContext`. Here are the dalec docs as well: https://azure.github.io/dalec/targets/#named-build-contexts

cc/ @cpuguy83 if I'm incorrect or missing anything

targets/linux/distro_handler.go
```diff
@@ -90,30 +90,66 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
...
-			deb, err := c.BuildPkg(ctx, client, worker, sOpt, spec, targetKey, pg, pc)
+			// We need to get the build context from the client.
+			inputs, err := client.Inputs(ctx)
 			if err != nil {
-				return nil, nil, err
+				return nil, nil, fmt.Errorf("error getting llb state from client with provided build context: %w", err)
+			}
+
+			// This makes sure it obeys include/exclude filters when called later on by the image resolver logic.
+			sOpt.GetContext = func(name string, opts ...llb.LocalOption) (*llb.State, error) {
```
cpuguy83, line 108: This should already be set on `sOpt`, no need to change anything.
Should definitely change this to an interface at some point rather than a struct with closures.

cpuguy83: What we want is to not look at `inputs` lower down directly and instead just call `sOpt.GetContext(<name>, opts...)`.
As mentioned above, `sOpt` should already be fully setup.

cpuguy83 (changes requested):

targets/linux/distro_handler.go
MorrisLaw: The opts we have here is of type:
```
type ConstraintsOpt interface {
	SetConstraintsOption(*Constraints)
	RunOption
	LocalOption
	HTTPOption
	ImageOption
	GitOption
	OCILayoutOption
}
```

And the one that sOpt.GetContext expects is:
```
type LocalOption interface {
	SetLocalOption(*LocalInfo)
}
```

Should I be initializing a diff opts separate from the one that we're working with in the function or is there a way to easily translate it over to the local version?

cpuguy83: All constraints opts implement LocalOption, no need to filter them out.

adamperlin: You can do `sOpt.GetContext(<name>, WithConstraints(opts...))`. `WithConstraintsOpts` combines multiple constraints opts into one, and that that constraints opt can be downcast into a `LocalOption`

target.go
```diff
@@ -7,6 +7,16 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// PreBuiltPkgSuffix is what is expected to be appended to a targetKey when it's
+	// meant to be a target specific package. (e.g. rpm-pkg, deb-pkg). When this is
+	// provided and used to build a container, it will take precedence over GenericPkg.
+	PreBuiltPkgSuffix = "-pkg"
```
cpuguy83, line 15: Did you have a use-case in mind for this?
We can only ever build one thing at a time, so I'm not sure where this would be used.

test/linux_target_test.go
```diff
@@ -3091,3 +3103,179 @@ int main() {
...
+
+			pkgSt, _ := pkgRef.ToState()
+			targetKey := strings.Split(testConfig.Target.Package, "/")[1]
+			targetSpecificName := targetKey + "-pkg"
+
+			// Use the built package directly as a named build context.
+			// This is how the prebuilt package gets picked up and used
+			// for the container build.
+			containerSr := newSolveRequest(
+				withSpec(ctx, t, spec),
```
cpuguy83: Should this use a new spec that doesn't have the build step in it?
Maybe just a test step that validates the content of the marker file?

target.go
MorrisLaw: You left a comment earlier around naming ideas:

```
Or perhaps either targetKey+"-pkg-dir" OR pkg-dir, with dalec preferring the former if it is set.

Another route would be to set it to the name of the build target (e.g. azlinux3/rpm) that produced the output.
Each target (e.g. azlinux3) can specify the name they are expecting to lookup from the inputs.
```

The former seemed the most straightforward.

My understanding was that this was a way to force a naming convention so that we could explicitly check for `pkg` passed into the build context. Which would indicate that the caller intended for a prebuilt pkg to be used.

So either `rpm-pkg` or just `pkg` would be considered valid.

Are you saying that a target specific package wouldn't make sense here?

test/linux_target_test.go
MorrisLaw: Logically it seemed like building the container with the proposed prebuilt package and then reading the contents of the file after the fact, was a good way to make sure it flowed through correctly and the prebuilt pkg used to build the container, built successfully.

cpuguy83: But then shouldn't this code work exactly the same before these changes are in place?
Right now it seems to me that we are testing that what worked before continues to work (which we already have tests for), but not explicitly ensuring that the new behavior is exercised because dalec could just be building the same exact spec again.

target.go
cpuguy83: Here I think I was referring to something like `mariner2-pkg` or even `mariner/rpm` as the potential context name rather than a generic `rpm` or `deb`.
`targetKey` there refers to the name of the target distro.

test/linux_target_test.go
MorrisLaw: My intention was to test that a pre-built package, when it exists, is being used by the container instead of a new one getting built in the process of building the container (the existing flow).

But maybe the test wasn't constructed in a way that actually tests that. I updated that. Let me know if the test code makes sense now. IIUC, the build step will only execute if the container rebuilds the package during the container solve. Since we're passing it into the `withBuildContext` function. It shouldn't rebuild that package.

Then we can confirm the "reuse prebuilt pkg" functionality is exercised by validating the contents/existence of the files.

target.go
MorrisLaw: Hm, I'll have to look into how to get the right target string combination. Will address this.

Just tbc, you're saying that the targetKey I'm using here in the code is the distro and not a package? Or are you just referring to the comment and that the definition of targetKey is incorrect?

test/linux_target_test.go
MorrisLaw: Hm, seems it failing. So I'm doing something wrong. Will look at this again after lunch.

target.go
MorrisLaw: Updated the comment.

targets/linux/distro_handler.go
```diff
@@ -90,30 +90,62 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
...
+			var foundPrebuiltPkg bool
+
+			// Pre-built packages can be provided via the build context by providing one of these package name formats:
+			// 1. {targetKey}-pkg - Target specific package context (e.g. rpm-pkg, deb-pkg).
+			// 2. pkg - Generic package for any package type.
+			//
+			// Target specific package will take precedence over the generic package.
+
+			// Try checking for target specific package first.
+			targetSpecificName := targetKey + dalec.PreBuiltPkgSuffix
```
cpuguy83: Can you extract this logic into a separate function?

test/linux_target_test.go
```diff
@@ -3091,3 +3095,89 @@ int main() {
...
+							},
+						},
+					},
+				},
+			},
+		}
+
+		testEnv.RunTest(ctx, t, func(ctx context.Context, client gwclient.Client) {
+			// This rebuilt.txt marker file will only be created if the package is rebuilt.
+			preBuiltSpec.Build.Steps = []dalec.BuildStep{
```
cpuguy83: I'm a little confused by the test here.
It looks like we build the package and then build the same definition again but with the original build as the passed in package state.
Wouldn't this test pass even before your changes from the PR?

I think we should build the package without the marker, then add the marker and validate that the marker does *not* exist since we aren't using the updated spec to build the package but instead are using the pre-built package.

targets/linux/distro_handler.go
```diff
@@ -90,30 +90,62 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
...
+				pkgSt = *targetPkgSt
+				foundPrebuiltPkg = true
+				opts = append(opts, dalec.ProgressGroup(fmt.Sprintf("Using pre-built package from %s context", targetSpecificName)))
+			} else {
+				// Try generic package.
+				genericPkgSt, err := sOpt.GetContext(dalec.GenericPkg, dalec.WithConstraints(opts...))
+				if err == nil && genericPkgSt != nil {
+					pkgSt = *genericPkgSt
+					foundPrebuiltPkg = true
+					opts = append(opts, dalec.ProgressGroup(fmt.Sprintf("Using pre-built package from %s context", dalec.GenericPkg)))
```
cpuguy83: I think these should be warnings we write to the client instead of progress groups.
`client` has a `Warn` function and, IIRC we have a helper method in the `frontend` package to simplify that a little.

cpuguy83 (changes requested):

targets/linux/distro_handler.go
MorrisLaw: done

test/linux_target_test.go
MorrisLaw: done

targets/linux/distro_handler.go
MorrisLaw: done

MorrisLaw: Thanks for highlighting this btw, I did not know about this function.

targets/linux/distro_handler.go
```diff
@@ -90,35 +90,71 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
...
+	var pkgSt llb.State
+	var foundPrebuiltPkg bool
+
+	targetSpecificName := targetKey + dalec.PreBuiltPkgSuffix
+	targetPkgSt, err := sOpt.GetContext(targetSpecificName, dalec.WithConstraints(opts...))
+
+	if err == nil && targetPkgSt != nil {
+		pkgSt = *targetPkgSt
+		foundPrebuiltPkg = true
+		frontend.Warn(ctx, client, pkgSt, fmt.Sprintf("Using pre-built package from %s context", targetSpecificName))
```
cpuguy83: This can return early and remove the `else`

targets/linux/distro_handler.go
```diff
@@ -90,35 +90,71 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
...
 }
 
+func getPrebuiltPackageName(ctx context.Context, targetKey string, client gwclient.Client, opts []llb.ConstraintsOpt, sOpt dalec.SourceOpts) (llb.State, bool) {
+	var pkgSt llb.State
+	var foundPrebuiltPkg bool
+
+	targetSpecificName := targetKey + dalec.PreBuiltPkgSuffix
+	targetPkgSt, err := sOpt.GetContext(targetSpecificName, dalec.WithConstraints(opts...))
+
+	if err == nil && targetPkgSt != nil {
```
cpuguy83: I think we need to handle these errors.
If there is an error then something is wrong in this case.

However, instead of having to return an error this can use `llb.Scratch().Async(<closure>)` which will capture the error and surface it when the state gets marshalled.

targets/linux/distro_handler.go
```diff
@@ -90,35 +90,71 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
...
-			worker, err := c.Worker(sOpt, pg, pc)
+			worker, err := c.Worker(sOpt, opts...)
 			if err != nil {
 				return nil, nil, err
 			}
 
-			deb, err := c.BuildPkg(ctx, client, worker, sOpt, spec, targetKey, pg, pc)
-			if err != nil {
-				return nil, nil, err
+			// Pre-built packages can be provided via the build context by providing one of these package name formats:
```
cpuguy83: And I guess these docs can move down into `getPrebuiltPackageName`

targets/linux/distro_handler.go
```diff
@@ -90,35 +90,71 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
...
 			}
 
-			ref, err := c.RunTests(ctx, client, worker, spec, sOpt, ctr, targetKey, pg, pc)
+			ref, err := c.RunTests(ctx, client, worker, spec, sOpt, ctr, targetKey, opts...)
 			return ref, img, err
 		})
 	}
 }
 
+func getPrebuiltPackageName(ctx context.Context, targetKey string, client gwclient.Client, opts []llb.ConstraintsOpt, sOpt dalec.SourceOpts) (llb.State, bool) {
```
cpuguy83:
```suggestion
func getPrebuiltPackage(ctx context.Context, targetKey string, client gwclient.Client, opts []llb.ConstraintsOpt, sOpt dalec.SourceOpts) (llb.State, bool) {
```

MorrisLaw:
Should be good now after addressing the timeout issue: https://github.com/Azure/dalec/pull/671

cpuguy83 (approved):

PR #638, opened 2025-05-22
Add support for disabling automatic runtime dependency injection

cpuguy83:
This commit adds a new configuration option to the `artifacts` section `disable_automatic_requires` which, when set to true, disables the automatic injection of runtime dependencies into the generated packages.

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -57,8 +57,8 @@ type Artifacts struct {
 	// Some tooling, such as `rpmbuild`, will look at all artifacts and
 	// automatically inject missing dependencies into the package metadata.
 	// For instance, if you include a `.sh` script, rpmbuild with automatically
-	// add `bash` as a depdendency for the package.
-	// It also does this for librabries being linked against.
+	// add `bash` as a dependency for the package.
+	// It also does this for libraries being linked against.
 	//
 	// This is useful if you want to have more control over the dependencies
 	// that are included in the package.
--- a/artifacts.go
+++ b/artifacts.go
@@ -50,6 +50,20 @@ type Artifacts struct {
 
 	// DisableStrip is used to disable stripping of artifacts.
 	DisableStrip bool `yaml:"disable_strip,omitempty" json:"disable_strip,omitempty"`
+
+	// DisableAutoRequires is used to disable automatic dependency discovery for
+	// the produced package.
+	//
+	// Some tooling, such as `rpmbuild`, will look at all artifacts and
+	// automatically inject missing dependencies into the package metadata.
+	// For instance, if you include a `.sh` script, rpmbuild with automatically
+	// add `bash` as a depdendency for the package.
+	// It also does this for librabries being linked against.
+	//
+	// This is useful if you want to have more control over the dependencies
+	// that are included in the package.
+	// However, you must be careful to manually include all dependencies that are required.
+	DisableAutoRequires bool `yaml:"disable_auto_requires,omitempty" json:"disable_auto_requires,omitempty"`
 }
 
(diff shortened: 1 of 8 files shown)


artifacts.go
```diff
@@ -50,6 +50,20 @@ type Artifacts struct {
...
 	// DisableStrip is used to disable stripping of artifacts.
 	DisableStrip bool `yaml:"disable_strip,omitempty" json:"disable_strip,omitempty"`
+
+	// DisableAutoRequires is used to disable automatic dependency discovery for
+	// the produced package.
+	//
+	// Some tooling, such as `rpmbuild`, will look at all artifacts and
+	// automatically inject missing dependencies into the package metadata.
+	// For instance, if you include a `.sh` script, rpmbuild with automatically
+	// add `bash` as a depdendency for the package.
```
DannyBrito, line 60:
```suggestion
	// add `bash` as a dependency for the package.
```

artifacts.go
```diff
@@ -50,6 +50,20 @@ type Artifacts struct {
...
+	// It also does this for librabries being linked against.
```
DannyBrito, line 61:
```suggestion
	// It also does this for libraries being linked against.
```

packaging/linux/deb/template_control.go
```diff
@@ -113,19 +113,29 @@ func (w *controlWrapper) depends(buf *strings.Builder, depsSpec *dalec.PackageDe
...
+				rtDeps = maps.Clone(rtDeps)
+				needsClone = false
+			}
 
-		rtDeps[shlibsDeps] = dalec.PackageConstraints{}
+			rtDeps[shlibsDeps] = dalec.PackageConstraints{}
+		}
 	}
 
+	// We _must_ add miscDeps regardless of `DisableAutoRequires` because
```
DannyBrito, line 136: is the underscore intended? 
```suggestion
	// We must add miscDeps regardless of `DisableAutoRequires` because
```

cpuguy83: Fixed.

artifacts.go
cpuguy83: Fixed.

website/docs/artifacts.md
```diff
@@ -332,3 +332,28 @@ This is a global setting that applies to all artifacts only.
 
 If you want some binaries stripped and others not, you will need to manually
 strip them in the build phase.
+
+## Automatic Dependency Resolution
+
+Some package tooling, such as `rpmbuild` or `debbuild` both used in core DALEC,
+will attempt to automatically resolve runtime dependencies for you based on
+the artifacts you have specified.
+This can be linked libraries or even detecting you've included a shell script
```
DannyBrito, line 341: The wording for this line a bit confusing
```suggestion
This can be linked libraries or even detecting you've included a shell script
```

cpuguy83: I'm not seeing a diff in that suggestion.

DannyBrito: yeah sorry, I don't have a suggested fix but I did want to point out that is a bit confusing

DannyBrito: maybe if I understand correctly
```suggestion
This can be from linking libraries or even detecting you've included a shell script
```

DannyBrito:
will this need to be handled for windows targets?

cpuguy83:
> will this need to be handled for windows targets?

No, since this affects the produced package only and we don't produce any kind of windows package.

DannyBrito (approved):
besides the nit of the doc, lgtm

PR #645, opened 2025-05-23
Add support for mounting sources per command

### LINKED ISSUE #437: [REQ] Add support for per-command mounts on image+cmd type

None

Right now you can add mounts to image sources like:

```yaml
sources:
  foo:
    image:
      cmd:
        mounts:
          # insert mounts here
```

The mounts get added to all steps defined in `cmd`.
The trouble here is now any command you run in the image is going to have its cache broken by any change to the mount content.
This means steps like installing some packages will always have to be re-run if the mount data changes even though the mount data does not affect that.

I propose that we can add support for adding mounts per cmd step.

cpuguy83:
The docker image source currently allows you to mount other sources in for *all* commands.
This change allows you to specify a mount for a specific command.

This can help improve performance by only mounting the sources that are needed for a specific command by improving the cache hit rate.

Closes #437

### DIFF
--- a/source.go
+++ b/source.go
@@ -354,6 +354,31 @@ func generateSourceFromImage(st llb.State, cmd *Command, sOpts SourceOpts, subPa
 			rOpts = append(rOpts, llb.AddEnv(k, v))
 		}
 
+		for _, m := range step.Mounts {
+			if err := m.validate(subPath); err != nil {
+				return llb.Scratch(), err
+			}
+
+			srcSt, err := m.Spec.AsMount(internalMountSourceName, sOpts, opts...)
+			if err != nil {
+				return llb.Scratch(), err
+			}
+
+			var mountOpt []llb.MountOption
+			// This handles the case where we are mounting a source with a target extract path and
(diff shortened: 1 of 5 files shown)


source.go
```diff
@@ -354,6 +354,31 @@ func generateSourceFromImage(st llb.State, cmd *Command, sOpts SourceOpts, subPa
...
 		}
 
+		for _, m := range step.Mounts {
+			if err := m.validate(subPath); err != nil {
+				return llb.Scratch(), err
+			}
+
+			srcSt, err := m.Spec.AsMount(internalMountSourceName, sOpts, opts...)
+			if err != nil {
+				return llb.Scratch(), err
```
MorrisLaw, line 364: Just for my understanding. How do we determine whether or not we should return the error and presumably end early? vs. propagating the error up to when llb state is marshaled?

cpuguy83: If I were to rewrite this (And I am rewriting it on a different branch :) ) I would not use an error return.

PR #653, opened 2025-06-04
test2jsongha: fix case where test timeouts are not reported

cpuguy83:
The test2json2gha tool was not reporting test timeouts correctly. It would still exit non-zero, however none of the data was reported to the console or error annotations so it was difficult to see what happened.

### DIFF
--- a/cmd/test2json2gha/event.go
+++ b/cmd/test2json2gha/event.go
@@ -78,6 +78,19 @@ func (h *resultsHandler) getOutputStream(te *TestEvent) (*TestResult, error) {
 	return tr, nil
 }
 
+// markUnfinishedAsTimeout marks all tests that have not been marked as pass, skip, or fail
+// as a timeout.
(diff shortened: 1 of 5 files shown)


cmd/test2json2gha/event.go
```diff
@@ -89,6 +102,8 @@ func (h *resultsHandler) HandleEvent(te *TestEvent) error {
 		tr.failed = true
 	case skip:
 		tr.skipped = true
+	case pass:
+		tr.pass = true
 	}
```
DannyBrito, line 107: wouldn't you want to also have a default or case for timeout?

cpuguy83: There is no case for a timeout.
All we get is "output" events with the test panic trace.
So we determine if a timeout happened based on if the test did not get a pass/fail/skip event, so it is effectively a test that did not finish running.

DannyBrito (approved):

PR #655, opened 2025-06-04
Prebuild worker images

cpuguy83:
Adds pre-built worker images to ghcr repo.
These get pushed on tag, but also create a GHA cache on merge.

Builders can either use `--cache-from` with the image ref for the versioned worker to use it as a cache for building the worker image OR use `--build-context <worker-name>=<worker image ref>` just use the worker image outright.


pmengelbert:
Can you explain how the containerd plugin system is being used here, and what it does generally? What does it mean to "load" a route/build target via a plugin?

internal/plugins/types.go
```diff
@@ -0,0 +1,23 @@
+package plugins
+
+import (
+	"context"
+
+	"github.com/moby/buildkit/frontend/gateway/client"
```
pmengelbert, line 6: I would alias this to `gwclient` since we do that almost everywhere else. Makes it clearer faster which client is being used

cmd/worker-img-matrix/main.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		defer outF.Close()
+	}
+
+	filter := func(r *plugins.Registration) bool {
+		return r.Type != plugins.TypeBuildTarget
+	}
+
+	var out []Info
+	for _, r := range plugins.Graph(filter) {
```
pmengelbert, line 47: For clarity, the target graph initially has only the top-level build-targets, which mostly correspond to the distro names. So in this case we are iterating over "mariner2", "azlinux3", "almalinux9", and so on.

Everything before the first slash in the build targets. If i understand correctly, the handlers for the remaining targets do not get loaded until later.

cmd/worker-img-matrix/main.go
```diff
@@ -0,0 +1,63 @@
...
+		var i Info
+		i.Target = path.Join(r.ID, "worker")
+		out = append(out, i)
```
pmengelbert, line 50: Should we `continue` on `debug/worker`? I'm pretty sure that's not a build target

pmengelbert: I'm misunderstanding. debug isn't ever returned by plugins.Graph, so no need to continue.

@cpuguy83, why doesn't the debug top-level target show up in the graph?

.github/workflows/worker-images.yml
```diff
@@ -0,0 +1,117 @@
...
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Git reference to build worker images for'
+        required: true
+        default: ''
+        type: string
+
+jobs:
+  generate-matrix:
```
pmengelbert: nit: it appears nothing is being generated here. Maybe call it `load-matrix:`

.github/workflows/worker-images.yml
```diff
@@ -0,0 +1,117 @@
...
+    branches:
+      - main
+      - 'releases/*'
+    tags:
+      - 'v[0-9]+.[0-9]+.*'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - '.github/workflows/worker-images/matrix.json'
```
pmengelbert, line 14: Where/how do the matrix files get generated and checked in?

pmengelbert: If manual, can we automate it?

cpuguy83: `go generate ./...`

CI validates that generated files are updated.

cmd/worker-img-matrix/main.go
cpuguy83: It only ever has those top-level build targets, those build targets have their own target routing (frontend.BuildMux), but do not get registered to the plugin graph since there is no need to.

cmd/worker-img-matrix/main.go
cpuguy83: Right now the code is just loading that directly in the main function.

PR #656, opened 2025-06-05
Allow SOURCE_DATE_EPOCH to be passed through without declaration

### LINKED ISSUE #447: [REQ] Allow SOURCE_DATE_EPOCH to be passed through without explicitly declaring it

None

`SOURCE_DATE_EPOCH` is a common variable set to improve reproducibility of builds.
We do have an explicit call-out for it in the spec, but it requires people explicitly declare the variable in the spec.

This means a builder who wants to use `SOURCE_DATE_EPOCH` must modify a spec to use it.
I don't think this should be a requirement.

bchuo:

Fixes #447 


load.go
```diff
@@ -435,6 +440,19 @@ func (s Spec) MarshalYAML() ([]byte, error) {
 
 func (s *BuildStep) processBuildArgs(lex *shell.Lex, args map[string]string, allowArg func(string) bool) error {
 	var errs []error
+
+	// Process the command field with a permissive allowArg function
+	// Commands can contain shell variables that aren't build arguments
+	if s.Command != "" {
```
cpuguy83: We don't want to modify the command, this would get taken care of by the shell.
What we would want to do is set an env var for the build if a value exists for SOURCE_DATE_EPOCH.

load.go
```diff
@@ -140,20 +140,35 @@ func (s *Spec) SubstituteArgs(env map[string]string, opts ...SubstituteOpt) erro
...
 		args[k] = v
 	}
+
 	for k, v := range env {
+		if k == "SOURCE_DATE_EPOCH" {
+			args[k] = v
+		}
+
 		if _, ok := args[k]; !ok {
 			if !knownArg(k) && !cfg.AllowArg(k) {
```
cpuguy83, line 149: Should the epoch stuff go into knownArg or cfg.allowArg?

bchuo: `SOURCE_DATE_EPOCH` already is in the knownArg section, but not sure what you mean about allowArg

cpuguy83: I'm just going by the diff here.
If that's the case then shouldn't L145 and L158 be redundant?

bchuo: Yes, you're right it'll add to the map twice. We still need the conditional there because we want to always pass the value through regardless if it was added to the args block. I can add a continue to bypass the rest

bchuo: Now that I think about it, we probably don't need it to be in knownargs anymore since it no longer requires opt in.

PR #658, opened 2025-06-12
node_modules caching via npm

DannyBrito:

poc for node_modules caching via npm.
previously including yarn, however getting yarn into build stage is not straight forward 

Fixes #

### DIFF
--- a/generator_nodemodules.go
+++ b/generator_nodemodules.go
@@ -26,26 +26,27 @@ func (s *Spec) HasNodeMods() bool {
 	return false
 }
 
-func withNodeMod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.ConstraintsOpt) func(llb.State) llb.State {
-	return func(in llb.State) llb.State {
-		workDir := "/work/src"
-		joinedWorkDir := filepath.Join(workDir, g.Subpath)
-		srcMount := llb.AddMount(workDir, srcSt)
-		installCmd := "npm install"
-		// Run install in a new state, mounting the source
-		installed := worker.Run(
+func withNodeMod(g *SourceGenerator, srcSt, worker llb.State, opts ...llb.ConstraintsOpt) llb.State {
+	workDir := "/work/src"
+	joinedWorkDir := filepath.Join(workDir, g.Subpath)
+	srcMount := llb.AddMount(workDir, srcSt)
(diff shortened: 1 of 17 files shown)


test/linux_target_test.go
```diff
@@ -1069,6 +1069,75 @@ Environment="KUBELET_KUBECONFIG_ARGS=--bootstrap-kubeconfig=/etc/kubernetes/boot
 		})
 	})
 
+	t.Run("node npm generator", func(t *testing.T) {
```
cpuguy83, line 1072: Can you move this into a separate test function?
Then you can re-use that test function in the windows side.

From here (assuming this is the distro test function?) you can to your `t.Run()` and call that function.
There are some examples of this down towards the end of this function.

generator_nodemodules.go
```diff
@@ -0,0 +1,95 @@
...
+		_, ok := sources[name]
+		return ok
+	}, opts...)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get patched sources")
+	}
+
+	result := make(map[string]llb.State)
+	sorted := SortMapKeys(patched)
+	opts = append(opts, ProgressGroup("Fetch node module dependencies for source: all before: "))
```
cpuguy83: Is this progress group name right?

generator_nodemodules.go
```diff
@@ -0,0 +1,95 @@
...
+		joinedWorkDir := filepath.Join(workDir, g.Subpath)
+		srcMount := llb.AddMount(workDir, srcSt)
+		installCmd := "npm install"
+		// Run install in a new state, mounting the source
+		installed := worker.Run(
+			ShArgs(installCmd),
+			llb.Dir(joinedWorkDir),
+			srcMount,
+			WithConstraints(opts...),
+		).AddMount(filepath.Join(joinedWorkDir, "node_modules"), in)
```
cpuguy83: What is "in" in this case?

I feel like we should be able to `.AddMount(joinedWorkDir, srcSt)` and avoid the copy operation below.

cpuguy83 (changes requested):

generator_nodemodules.go
DannyBrito: fixed

PR #660, opened 2025-06-16
Add pip dep generation support and caching

### LINKED ISSUE #628: [REQ] Add support for python pip requirements

None

I think this would just be treated like another generator.
Basically a python project would have a requirements.txt which one would typically use `pip` (or `pip3` as it were) to install.

One headache here is that pip dependencies can often rely pre-compiled "wheels" which are arch specific.
pip does have a [--no-binary](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-no-binary) flag, which could be an option.

bchuo:

Fixes #628


docs/spec.schema.json
```diff
@@ -1384,7 +1431,7 @@
...
+					"description": "Generate is the list generators to run on the source.\n\nGenerators are used to generate additional sources from this source.\nAs an example the `gomod` generator can be used to generate a go module cache from a go source.\nHow a generator operates is dependent on the actual generator.\nGenerators may also cauuse modifications to the build environment.\n\nCurrently supported generators are: \"gomod\", \"cargohome\", and \"pip\".\nThe \"gomod\" generator will generate a go module cache from the source.\nThe \"cargohome\" generator will generate a cargo home from the source.\nThe \"pip\" generator will generate a pip cache from the source."
```
MorrisLaw:
```suggestion
					"description": "Generate is the list of generators to run on the source.\n\nGenerators are used to generate additional sources from this source.\nAs an example the `gomod` generator can be used to generate a go module cache from a go source.\nHow a generator operates is dependent on the actual generator.\nGenerators may also cause modifications to the build environment.\n\nCurrently supported generators are: \"gomod\", \"cargohome\", and \"pip\".\nThe \"gomod\" generator will generate a go module cache from the source.\nThe \"cargohome\" generator will generate a cargo home from the source.\nThe \"pip\" generator will generate a pip cache from the source."
```

packaging/linux/deb/debroot.go
```diff
@@ -296,6 +296,16 @@ func fixupGenerators(spec *dalec.Spec, cfg *SourcePkgConfig) []byte {
 		fmt.Fprint(buf, "\n")
 	}
 
+	if spec.HasPips() {
+		// Set PIP environment variables to point to our prepared pip packages
+		// Use --break-system-packages since we're in an isolated build container
```
adamperlin: Was there a specific reason you needed to set `break-system-packages` here? In this case our "system" would be the isolated build container, it's true, but I'm not sure that necessarily means we couldn't run into conflicts. I'm wondering about the case where someone has a build dependency (installed via the package manager) that depends on python + python packages. If they then use the pip generator, could that cause issues?

website/docs/examples/targets.md
```diff
@@ -20,6 +20,7 @@ bullseye/testing/container       Builds a container image for testing purposes o
 bullseye/worker                  Builds the worker image.
 debug/gomods                     Outputs all the gomodule dependencies for the spec
 debug/cargohome                  Outputs all the cargohome dependencies for the spec
+debug/pip                        Outputs all the cargohome dependencies for the spec
```
adamperlin:
```suggestion
debug/pip                        Outputs all the pip dependencies for the spec
```

generator_pip.go
```diff
@@ -0,0 +1,130 @@
...
+	for _, gen := range s.Generate {
+		if gen.Pip != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func (s *Spec) HasPips() bool {
+	for _, src := range s.Sources {
```
MorrisLaw, line 20: I wonder if we can cache this instead of doing O(n) lookup every time.

What if we exported the `pipSources` function (e.g. `UpdatePipSources`) and that can populate a `PipSources` field on the spec the first time it's needed. Then, in `HasPips()` we could do:

```go
func (s *Spec) HasPips() bool {
    return len(s.PipSources) > 0
}
```

This would give us O(1) lookups after that first initialization instead of redoing the loop every time we call `HasPips()`. Do you think that would work or is there a reason I'm not seeing that requires re-calculating every time?

packaging/linux/deb/debroot.go
bchuo: So without the break-system-packages being set to 1, pip would refuse to install packages on all the distros, with an error along the lines of

```
error: externally-managed-environment

× This environment is externally managed
╰─ To install Python packages system-wide, try apt install
   python3-xyz, where xyz is the package you are trying to
   install.
   
   If you wish to install a non-Debian-packaged Python package,
   create a virtual environment using python3 -m venv path/to/venv.
   Then use path/to/venv/bin/pip install xyz.
```

In terms of potentially breaking something, the flag only applies to the build container, which is discarded after buildtime. I couldn't find another workaround to avoid this other than using a virtual environment, but that introduced its own set of problems, such as some distros not having the venv package available or the above error still persisting.

generator_pip.go
cpuguy83: I don't think we want to expose anything like `PipSources` on the spec type.
This also really shouldn't be a bottleneck, even with hundreds of sources.

bchuo: We only call HasPips once per build so I'm not sure if optimizing here gives us added benefits

MorrisLaw: > This also really shouldn't be a bottleneck, even with hundreds of sources.

That's fair. Thanks for clarifying!

docs/spec.schema.json
```diff
@@ -1384,7 +1431,7 @@
...
+					"description": "Generate is the list generators to run on the source.\n\nGenerators are used to generate additional sources from this source.\nAs an example the `gomod` generator can be used to generate a go module cache from a go source.\nHow a generator operates is dependent on the actual generator.\nGenerators may also cause modifications to the build environment.\n\nCurrently supported generators are: \"gomod\", \"cargohome\", and \"pip\".\nThe \"gomod\" generator will generate a go module cache from the source.\nThe \"cargohome\" generator will generate a cargo home from the source.\nThe \"pip\" generator will generate a pip cache from the source."
```
MorrisLaw: I think my previous suggestion may have been missed. I believe there is opportunity to make this documentation clearer.

What are we trying to convey here:
```
Generate is the list generators to run on the source.
```

For example, instead of `“Generate is the list generators to run on the source,”` maybe `“Generate specifies a list of generators to apply to the source.”`

Even small improvements in documentation clarity can make a big difference for readers.

packaging/linux/deb/debroot.go
```diff
@@ -295,6 +295,14 @@ func fixupGenerators(spec *dalec.Spec, cfg *SourcePkgConfig) []byte {
 		// Above command does not have a newline due to quoting issues, so add that here.
 		fmt.Fprint(buf, "\n")
 	}
+
+	if spec.HasPips() {
```
cpuguy83: I think maybe let's let the spec author deal with this.

targets/windows/handle_zip.go
```diff
@@ -240,6 +250,13 @@ func createBuildScript(spec *dalec.Spec, opts ...llb.ConstraintsOpt) llb.State {
 		fmt.Fprintln(buf, "export CARGO_HOME=\"$(pwd)/"+cargohomeName+"\"")
 	}
 
+	if spec.HasPips() {
```
cpuguy83: Same as other comment, let's have the spec author deal with this.

test/fixtures/phony.go
```diff
@@ -15,6 +15,7 @@ import (
 var (
 	goModCache   = llb.AddMount("/go/pkg/mod", llb.Scratch(), llb.AsPersistentCacheDir("/go/pkg/mod", llb.CacheMountShared))
 	goBuildCache = llb.AddMount("/root/.cache/go-build", llb.Scratch(), llb.AsPersistentCacheDir("/root/.cache/go-build", llb.CacheMountShared))
+	pipCache     = llb.AddMount("/root/.cache/pip", llb.Scratch(), llb.AsPersistentCacheDir("/root/.cache/pip", llb.CacheMountShared))
```
cpuguy83: Where is this being used?

packaging/linux/rpm/sources.go
```diff
@@ -47,6 +47,13 @@ func buildScript(spec *dalec.Spec) string {
 		fmt.Fprintln(b, "export CARGO_HOME=\"$(pwd)/"+cargohomeName+"\"")
 	}
 
+	if spec.HasPips() {
```
cpuguy83: Same as other comment, let's have the spec author deal with this.

packaging/linux/deb/debroot.go
```diff
@@ -340,6 +348,31 @@ func createBuildScript(spec *dalec.Spec, cfg *SourcePkgConfig) []byte {
 	buf := bytes.NewBuffer(nil)
 	writeScriptHeader(buf, cfg)
 
+	// Add environment setup for generators (similar to RPM buildScript)
+	if spec.HasGomods() {
```
cpuguy83: Did this move from somewhere else?
This should all have already been taken care of.

packaging/linux/deb/debroot.go
```diff
@@ -340,6 +348,31 @@ func createBuildScript(spec *dalec.Spec, cfg *SourcePkgConfig) []byte {
...
+
+		fmt.Fprintln(buf, "export GOMODCACHE=\"$(pwd)/"+gomodsName+"\"")
+	}
+
+	if spec.HasCargohomes() {
+		// Set CARGO_HOME to point to our prepared cargo cache
+		fmt.Fprintln(buf, "export CARGO_HOME=\"$(pwd)/"+cargohomeName+"\"")
+	}
+
+	if spec.HasPips() {
```
cpuguy83: Same as other comment, let's have the spec author deal with this.

test/fixtures/phony.go
bchuo: I think this was some residual from when I initially tried to a build cache for python builds too. I'll remove it

packaging/linux/deb/debroot.go
bchuo: It seems like this might've been removed in a previous PR? Perhaps it was an accident

https://github.com/Azure/dalec/blob/main/packaging/linux/deb/debroot.go

cpuguy83: Oh that's done here:https://github.com/Azure/dalec/blob/f87c3c06daf67eb205df100785cf3295693c3da9/packaging/linux/deb/template_rules.go#L50-L66

website/docs/sources.md
```diff
@@ -433,6 +433,64 @@ sources:
...
+### Pip
+
+The `pip` generator manages a single pip cache for all sources that specify it in the spec. It is expected that the build dependencies include a Python toolchain with pip suitable for fetching Python package dependencies.
+
+Adding a pip generator to 1 or more sources causes the following to occur automatically:
+
+1. Fetch all pip dependencies for *all* sources in the spec that specify the generator
+2. Keeps a single pip cache directory for all Python package deps.
+3. Adds the pip cache directory as a source which gets included in source packages like a normal source.
+4. Adds the `PIP_CACHE_DIR` environment variable to the build environment.
```
cpuguy83: I think these steps aren't right anymore?

website/docs/sources.md
```diff
@@ -337,7 +337,7 @@ Build sources are considered to be "directory" sources.
 ## Generators
 
 Generators are used to generate a source from another source.
-Currently the only generators supported are `gomod` and `cargohome`.
+Currently the generators supported are `gomod`, `cargohome`, and `pip`.
```
cpuguy83: Can you add `nodemod` in there?

cpuguy83 (approved):

PR #667, opened 2025-06-24
Fix buildx version check

cpuguy83:
- Use feature check instead of version check for buildx dial-stdio
- Fix subprocess handling for buildx dial-stdio

### DIFF
--- a/test/testenv/buildx.go
+++ b/test/testenv/buildx.go
@@ -11,7 +11,6 @@ import (
 	"net"
 	"os"
 	"os/exec"
-	"strconv"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


test/testenv/buildx.go
```diff
@@ -59,59 +58,48 @@ func (b *BuildxEnv) Load(ctx context.Context, id string, f gwclient.BuildFunc) e
...
 	if err != nil {
-		return "", pkgerrors.Wrap(err, string(out))
+		return false, pkgerrors.Wrap(err, string(out))
 	}
 
-	fields := strings.Fields(string(out))
-
-	if len(fields) != 3 {
-		return "", errors.New("could not determine buildx version")
+	if !strings.Contains(string(out), "dial-stdio") {
```
adamperlin: Maybe add a comment about the logic here:
`buildx --help` includes `dial-stdio` => `dial-stdio` supported

adamperlin (approved):
Small comment nit but otherwise looks good

test/testenv/buildx.go
cpuguy83: Sure, updated.
Also cleaned up the logic a little bit after reading more of the code.

DannyBrito (approved):

PR #673, opened 2025-06-27
deb: Make sure base packages are installed first

cpuguy83:
The base packages need to be installed before we try to install the main package because the install may require those base packages (e.g. the main use-case here is ca-certificates for importing repos).


cpuguy83:
I'm not positive how to test the actual use-case here outside of manual testing (which I did) since it requires certs that are trusted by ca-certificates.

adamperlin (approved):
I'm glad you caught this!

targets/linux/deb/distro/container.go
```diff
@@ -48,6 +48,7 @@ func (c *Config) BuildContainer(ctx context.Context, client gwclient.Client, wor
 
 	// If we have base packages to install, create a meta-package to install them.
 	if len(c.BasePackages) > 0 {
+		frontend.Warnf(ctx, client, debSt, "Installing base packages into container image %v", c.BasePackages)
```
cpuguy83: Oops, didn't mean to include this.

cpuguy83:
Argh... now git-worktree issue... one sec.

adamperlin:
This looks good to me pending CI!

PR #677, opened 2025-07-01
Update some modules:

cpuguy83:
1. go-jsonschema
2. go-yaml
3. buildkit

These of course all brought their own transitive updates.


load.go
```diff
@@ -361,7 +361,15 @@ func (s *Spec) UnmarshalYAML(dt []byte) error {
 
 	// Remove extension nodes from the main AST and store them so we can unmarshal
 	// them separately.
-	body := parsed.Docs[0].Body.(*ast.MappingNode)
+	b := parsed.Docs[0].Body
```
adamperlin, line 364: Did this case come up because of a change in the yaml library?

cpuguy83: Yes, it caused a panic in tests.

adamperlin (approved):

PR #692, opened 2025-07-18
Skip installing docs into containers unless spec has docs

cpuguy83:

Fixes #

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -185,3 +185,16 @@ func (a *Artifacts) IsEmpty() bool {
 	}
 	return true
 }
+
+func (a Artifacts) HasDocs() bool {
+	if len(a.Docs) > 0 {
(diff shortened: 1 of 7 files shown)


artifacts.go
```diff
@@ -185,3 +185,16 @@ func (a *Artifacts) IsEmpty() bool {
 	}
 	return true
 }
+
+func (a Artifacts) HasDocs() bool {
+	if len(a.Docs) > 0 {
+		return true
+	}
+	if len(a.Licenses) > 0 {
```
jeremyrickard, line 193: won't this always be true for our uses with build-deft? We require license I believe?

cpuguy83: Oh yes. Let me fix that. I think I just auto-completed this and didn't think about it.

PR #723, opened 2025-08-05
bazel: Remove unnecessary llb.IgnoreCache

cpuguy83:
This was never really neccessary and only added to make sure tests can run effectively.
It did not help tests at all and also makes it so we can end up always busting the buildkit cache for builds with a bazel cache.


adamperlin (approved):
This lgtm pending integration tests!

PR #734, opened 2025-08-13
feat: Add support for building systemd system extensions

chewi:
Sysexts are not containers and therefore do not necessarily need to include all their dependencies. Exactly how much to include ultimately depends on how the sysext is intended to be used. To give the spec author full control over this, only the built package and the dependencies explicitly listed in the spec are installed to the sysext. These are extracted rather than installed by the package manager.

For now, the generated sysext is a bare erofs filesystem rather than a partitioned disk image.

This adds `/sysext` to the azlinux3 and noble targets. These are the only targets with a new enough erofs-utils to include tar support.

/etc can only be included in confexts rather than sysexts. For now, move anything in /etc (except systemd) to /usr/share/NAME/etc and copy that data back again at runtime with systemd-tmpfiles. This is what Flatcar does.

Any systemd services are automatically started when the sysext is attached thanks to a drop-in against multi-user.target.

### DIFF
--- a/test/target_ubuntu_test.go
+++ b/test/target_ubuntu_test.go
@@ -26,7 +26,7 @@ func withPackageOverride(oldPkg, newPkg string) func(cfg *testLinuxConfig) {
 func debLinuxTestConfigFor(targetKey string, cfg *distro.Config, opts ...func(*testLinuxConfig)) testLinuxConfig {
 	var sysextTarget string
 	if cfg.SysextSupported {
-		sysextTarget = targetKey + "/sysext"
+		sysextTarget = targetKey + "/testing/sysext"
 	}
 
 	tlc := testLinuxConfig{
--- a/website/docs/examples/targets.md
+++ b/website/docs/examples/targets.md
@@ -5,7 +5,7 @@ azlinux3/rpm                     Builds an rpm and src.rpm.
 azlinux3/rpm/debug/buildroot     Outputs an rpm buildroot suitable for passing to rpmbuild.
 azlinux3/rpm/debug/sources       Outputs all the sources specified in the spec file in the format given to rpmbuild.
 azlinux3/rpm/debug/spec          Outputs the generated RPM spec file
-azlinux3/sysext                  Builds a systemd system extension.
+azlinux3/testing/sysext          Builds a systemd system extension for testing purposes only.
 azlinux3/worker                  Builds the base worker image responsible for building the rpm
 bionic/deb (default)             Builds a deb package.
 bionic/dsc                       Builds a Debian source package.
@@ -42,7 +42,7 @@ mariner2/worker                  Builds the base worker image responsible for bu
 noble/deb (default)              Builds a deb package.
 noble/dsc                        Builds a Debian source package.
 noble/testing/container          Builds a container image for testing purposes only.
-noble/sysext                     Builds a systemd system extension.
+noble/testing/sysext             Builds a systemd system extension for testing purposes only.
 noble/worker                     Builds the worker image.
 windowscross/container (default) Builds binaries and installs them into a Windows base image
 windowscross/worker              Builds the base worker image responsible for building the package
--- a/website/docs/system-extensions.md
+++ b/website/docs/system-extensions.md
@@ -1,5 +1,5 @@
 ---
-title: System Extensions
+title: System Extensions (EXPERIMENTAL)
 ---
 
 Dalec can create system extensions, known as sysexts, for use with systemd. Such extensions are filesystem images that get overlaid onto an existing running system. These are often designed to be standalone, making them very portable, but they don't necessarily have to be.
@@ -31,7 +31,7 @@ dependencies:
 Build the container:
 
 ```shell
-docker build -f my-sysext.yml --target=azlinux3/sysext --output=. .
+docker build -f my-sysext.yml --target=azlinux3/testing/sysext --output=. .
 ```
 
 This creates `my-sysext-v0.1.0-1-azlinux3-x86-64.raw` in the current directory. You can merge this with `systemd-sysext` or just mount it to some directory with `mount -o loop`. It contains:
--- a/targets/linux/deb/distro/distro.go
+++ b/targets/linux/deb/distro/distro.go
@@ -121,8 +121,8 @@ func (cfg *Config) Handle(ctx context.Context, client gwclient.Client) (*gwclien
 	})
 
 	if cfg.SysextSupported {
-		mux.Add("sysext", linux.HandleSysext(cfg), &targets.Target{
-			Name:        "sysext",
+		mux.Add("testing/sysext", linux.HandleSysext(cfg), &targets.Target{
+			Name:        "testing/sysext",
 			Description: "Builds a systemd system extension image.",
 		})
 	}
--- a/targets/linux/rpm/distro/distro.go
+++ b/targets/linux/rpm/distro/distro.go
@@ -83,8 +83,8 @@ func (cfg *Config) Handle(ctx context.Context, client gwclient.Client) (*gwclien
 	})
 
 	if cfg.SysextSupported {
-		mux.Add("sysext", linux.HandleSysext(cfg), &targets.Target{
-			Name:        "sysext",
+		mux.Add("testing/sysext", linux.HandleSysext(cfg), &targets.Target{
+			Name:        "testing/sysext",
 			Description: "Builds a systemd system extension image.",
 		})
 	}
--- a/test/target_azlinux_test.go
+++ b/test/target_azlinux_test.go
@@ -85,7 +85,7 @@ func TestAzlinux3(t *testing.T) {
 			Package:               "azlinux3/rpm",
 			Container:             "azlinux3/container",
 			Worker:                "azlinux3/worker",
-			Sysext:                "azlinux3/sysext",
+			Sysext:                "azlinux3/testing/sysext",
 			ListExpectedSignFiles: azlinuxListSignFiles("azl3"),
 		},
 		LicenseDir: "/usr/share/licenses",


targets/linux/build_sysext.sh
```diff
@@ -0,0 +1,65 @@
...
+ARCHITECTURE=${ARCH}
+EXTENSION_RELOAD_MANAGER=1
+EOF
+
+cd /input
+shopt -s extglob nullglob
+
+# Sysexts cannot include /etc, so move that data to /usr/share/${NAME}/etc and
+# copy it to /etc at runtime.
+for ITEM in etc/!(systemd); do
```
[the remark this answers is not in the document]
chewi: Dalec already assumes bash in at least one other place. The rest of the script would be quite awkward if I didn't assume bash.

targets/linux/deb/distro/install.go
```diff
@@ -197,3 +197,57 @@ func (d *Config) InstallTestDeps(sOpt dalec.SourceOpts, targetKey string, spec *
...
+	}
+
+	constraints := deps.Runtime
+	if constraints == nil {
+		return llb.Scratch()
+	}
+
+	scriptPath := "/tmp/dalec/internal/deb/download.sh"
+	const scriptSrc = `#!/usr/bin/env sh
+set -ex
```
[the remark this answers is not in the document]
chewi: `sh` might not be bash, so `pipefail` might not be supported. On reflection though, I probably should use bash here.

targets/linux/build_sysext.sh
```diff
@@ -0,0 +1,65 @@
...
+for ITEM in usr/lib/systemd/system/!(*@*).service; do
+	ITEM=${ITEM##*/}
+	mkdir -p "${TMPDIR}"/usr/lib/systemd/system/multi-user.target.d
+	cat > "${TMPDIR}/usr/lib/systemd/system/multi-user.target.d/10-${NAME}-${ITEM%.service}.conf" <<-EOF
+	[Unit]
+	Upholds=${ITEM}
+	EOF
+done
+
+tar \
```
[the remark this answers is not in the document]
chewi: This script already calls `set -euo pipefail`, so that is covered.

targets/linux/build_sysext.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
cpuguy83:
```suggestion
#!/usr/bin/env bash
```

targets/linux/deb/distro/install.go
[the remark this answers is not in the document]
cpuguy83: Yeah I think using bash here is fine and pipefail a good option given the script is using pipes.

targets/linux/deb/distro/install.go
```diff
@@ -197,3 +197,57 @@ func (d *Config) InstallTestDeps(sOpt dalec.SourceOpts, targetKey string, spec *
...
+cd /output
+
+# Make sure any cached data from local repos is purged since this should not
+# be shared between builds.
+rm -f /var/lib/apt/lists/_*
```
cpuguy83, line 215: Why shouldn't it be shared between builds?
Also, if this is going to remove stuff then consider not setting `dalec.WithMountedAptCache`

targets/linux/distro_handler.go
```diff
@@ -126,6 +131,83 @@ func HandleContainer(c DistroConfig) gwclient.BuildFunc {
...
+			}
+
+			script := llb.Scratch().File(
+				llb.Mkfile("build_sysext.sh", 0o755, []byte(buildSysextSh)),
+				dalec.WithConstraints(opts...),
+			)
+
+			erofs := worker.Run(
+				llb.Args([]string{"/build_sysext.sh", spec.Name, spec.Version, platform.Architecture}),
+				llb.AddMount("/build_sysext.sh", script, llb.SourcePath("build_sysext.sh"), llb.Readonly),
```
cpuguy83: Can you mount into some path like `/tmp/dalec/internal/systext/build.sh`?

targets/linux/rpm/azlinux/azlinux3.go
```diff
@@ -23,7 +23,7 @@ var Azlinux3Config = &distro.Config{
 	CacheDir:  "/var/cache/tdnf",
 
 	ReleaseVer:         "3.0",
-	BuilderPackages:    builderPackages,
+	BuilderPackages:    append(builderPackages, "dnf-utils", "erofs-utils"),
```
cpuguy83: Can these added packages be installed only for sysext target instead?

test/target_ubuntu_test.go
```diff
@@ -24,11 +25,17 @@ func withPackageOverride(oldPkg, newPkg string) func(cfg *testLinuxConfig) {
 }
 
 func debLinuxTestConfigFor(targetKey string, cfg *distro.Config, opts ...func(*testLinuxConfig)) testLinuxConfig {
+	var sysextTarget string
+	if slices.Contains(cfg.BuilderPackages, "erofs-utils") {
```
cpuguy83: Maybe noble can pass an opt here that sets sysextTarget instead and only run the sysext test when the target is set?

test/linux_target_test.go
```diff
@@ -666,6 +668,311 @@ echo "$BAR" > bar.txt
...
+
+			var cfg dalec.DockerImageSpec
+			assert.Assert(t, json.Unmarshal(dt, &cfg))
+			assert.Check(t, cfg.Created.After(beforeBuild))
+			assert.Check(t, cfg.Created.Before(time.Now()))
+
+			// Map Docker to systemd architecture.
+			var systemdArch string
+			switch cfg.Platform.Architecture {
+			case "amd64":
```
cpuguy83, line 1114: Worthwhile to do arm64 here as well, even though our CI only has amd64 right now.... assuming this is aarch64 for systemd.

cpuguy83 (changes requested):

targets/linux/deb/distro/install.go
chewi: This is copied from `AptInstall` at the top of that file. I didn't know this before, but I guess local repo cache files start with a `_`. I think it makes sense that you would not want to share these.

test/linux_target_test.go
chewi: That's covered by the `default` case. It's arm64 for systemd as well. See `systemd-analyze architectures`.

test/target_ubuntu_test.go
chewi: I've done this in a slightly different way now, but I think it's close enough to what you wanted.

chewi:
I've now added a new `sysext` dependency type to use for this rather than `runtime`. This allows existing specs to be extended for sysext use rather than having to create new ones just to avoid unwanted runtime dependencies.

targets/linux/deb/distro/install.go
cpuguy83: Oh! Right I remember this now.
This is fine.

deps.go
```diff
@@ -31,6 +31,8 @@ type PackageDependencies struct {
 	// Recommends is the list of packages recommended to install with the generated package.
 	// Note: Not all package managers support this (e.g. rpm)
 	Recommends map[string]PackageConstraints `yaml:"recommends,omitempty" json:"recommends,omitempty"`
+	// Sysext is the list of packages to include in the generated system extension.
+	Sysext map[string]PackageConstraints `yaml:"sysext,omitempty" json:"sysext,omitempty"`
```
cpuguy83, line 37: This is a pretty clever way to get around including dependencies that we don't want in the sysext.
I've given a couple of other ideas some thought here and I think this is still solves the issue for what we actually need, so :+1:

cpuguy83: Perhaps this can include some more for the doc string to state that all deps that need to be included must be explicity defined and we don't do any resolution on the dependency graph beyond exactly what is specified.

cpuguy83 (changes requested):
Some extra docs on the website about the new target type and how dependency resolution is different would be really helpful here.

deps.go
chewi: Thanks! I've expanded the doc string as requested.

targets/linux/deb/distro/pkg.go
```diff
@@ -245,3 +245,21 @@ func (cfg *Config) HandleSourcePkg(ctx context.Context, client gwclient.Client)
...
+
+	opts = append(opts, dalec.ProgressGroup("Extracting DEBs"))
+
+	return worker.Run(
+		llb.Args([]string{"find", "/input", "-name", "*.deb", "-exec", "dpkg-deb", "--verbose", "--extract", "{}", "/output", ";"}),
+		llb.AddMount("/input/build", debSt),
+		llb.AddMount("/input/deps", depDebs),
+		llb.AddMount("/output", llb.Scratch()),
+		dalec.WithConstraints(opts...),
+	).GetMount("/output")
```
adamperlin: I believe the `AddMount("/output")` and  `GetMount("/output"/)` can be replaced by a single `AddMount("/output", llb.Scratch())` on the return value of `Run` which will both mount the scratch fs for the Run operation and also return the contents of `"/output"`

adamperlin:
```suggestion
	return worker.Run(
		llb.Args([]string{"find", "/input", "-name", "*.deb", "-exec", "dpkg-deb", "--verbose", "--extract", "{}", "/output", ";"}),
		llb.AddMount("/input/build", debSt),
		llb.AddMount("/input/deps", depDebs),
		dalec.WithConstraints(opts...),
	).AddMount("/output", llb.Scratch())
```

targets/linux/distro_handler.go
```diff
@@ -33,6 +35,10 @@ type DistroConfig interface {
 		sOpt dalec.SourceOpts,
 		spec *dalec.Spec, targetKey string, opts ...llb.ConstraintsOpt) (llb.State, error)
 
+	ExtractPkg(ctx context.Context, client gwclient.Client, worker llb.State, sOpt dalec.SourceOpts,
```
adamperlin, line 42: Can you add a comment here explaining what this method should do if implemented?

test/linux_target_test.go
adamperlin: Can you add this in a comment then to explain why we don't need a separate case here?

adamperlin (changes requested):
I'm wondering if it is possible to test the system extension after building. I noticed that the system extension target doesn't execute the Dalec spec's tests in any way, and I'm curious if there would be a way to do this.

targets/linux/deb/distro/pkg.go
chewi: Nice catch, thanks! I've done this with the other instances too.

chewi:
While putting the CI changes together for dalec-build-defs, I decided to include the revision in the sysext filename, e.g. `kubelet-1.33.3-1-x86-64.raw`. Package filenames include it, so I think it makes sense to include it here too, and it also makes the CI stuff easier.

pmengelbert (changes requested):
> > Some extra docs on the website about the new target type and how dependency resolution is different would be really helpful here.
> 
> I would certainly like to add that, but I thought it might be best to get some soak time internally before making it more widely visible. Ideally, I would add support for signing and make it generate full disk images first. I would have tackled the disk images already, but I think systemd-repart requires a signature.

It's too easy for documentation of a feature to be deferred and forgotten about. If you really want to keep it invisible for now, commit the docs in this PR and prevent them from being processed when the website generation job runs. You'll also want to open an issue to add it back to the website generation rules when you're ready to go public with it.

That said, I don't think there will be any harm done by documenting the feature on the website now, even if you don't consider the feature finished. You can mark it as experimental in the docs.

deps.go
```diff
@@ -31,6 +31,10 @@ type PackageDependencies struct {
 	// Recommends is the list of packages recommended to install with the generated package.
 	// Note: Not all package managers support this (e.g. rpm)
 	Recommends map[string]PackageConstraints `yaml:"recommends,omitempty" json:"recommends,omitempty"`
+	// Sysext is the list of packages to include in the generated system
+	// extension. No dependency resolution is performed when generating system
+	// extensions, so all required dependencies must be explicitly listed here.
+	Sysext map[string]PackageConstraints `yaml:"sysext,omitempty" json:"sysext,omitempty"`
```
pmengelbert, line 37: Can you add a test that defines and proves the precise behavior expected when these dependencies are added to the spec?

targets/linux/deb/distro/install.go
```diff
@@ -197,3 +197,50 @@ func (d *Config) InstallTestDeps(sOpt dalec.SourceOpts, targetKey string, spec *
 		).Root()
 	}
 }
+
+func (d *Config) DownloadDeps(worker llb.State, sOpt dalec.SourceOpts, spec *dalec.Spec, targetKey string, constraints map[string]dalec.PackageConstraints, opts ...llb.ConstraintsOpt) llb.State {
```
pmengelbert, line 201: As mentioned above, this code path should have test coverage

pmengelbert (changes requested):

adamperlin:
> While putting the CI changes together for dalec-build-defs, I decided to include the revision in the sysext filename, e.g. `kubelet-1.33.3-1-x86-64.raw`. Package filenames include it, so I think it makes sense to include it here too, and it also makes the CI stuff easier.

I think this is a good idea, since we make use of revision basically everywhere Dalec artifacts are used (including in image tags).

adamperlin:
> > I'm wondering if it is possible to test the system extension after building. I noticed that the system extension target doesn't execute the Dalec spec's tests in any way, and I'm curious if there would be a way to do this.
> 
> This is done for both the container and package targets. For the latter, they build the container to run the tests in and then throw it away. I guess we could do the same here. It wouldn't test whether the sysext itself works, but we should still run some kind of test in case the other targets are not used. I expect the test run would be cached, so it shouldn't hurt the build time? I don't think testing the sysext itself is feasible.

I do think we should run some kind of test in case the other targets aren't used, that was the case I was thinking of in my original comment. I guess I'm curious what makes testing the sysext itself infeasible? Could it theoretically be tested in a container with systemd running? That being said, since the sysext is just extracting from the built package, running the tests on a container with the corresponding package installed should hopefully be sufficient for ensuring that files are included, etc.

website/docs/system-extensions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Static linking
+
+The above issues can be largely avoided by including statically-linked rather than dynamically-linked binaries. This will make your system extension much more portable. However, binaries shipped by distributions will practically always be dynamic, so this will require you to build the binaries and their dependencies yourself as part of the spec.
+
+### Prefixing
+
+Conflicts may still occur even when static linking. The extension should ideally place its files under a unique prefix, e.g. `/usr/local/my-sysext`. The standard layout may exist below this, e.g. `/usr/local/sysext/bin`. For this to work, the binaries should be built specifically for this prefix. Dynamically-linked binaries should explicitly include somewhere like `/usr/local/sysext/lib` in the `RUNPATH` to avoid the need for wrappers that set `LD_LIBRARY_PATH`.
```
jepio: Are you sure about giving this recommendation as "ideally"?

This approach seems like it's reserved for exotic cases and there's so many things to get right/wrong here.

chewi: It can be as simple as `LDFLAGS="-Wl,-rpath=/usr/local/sysext/lib" ./configure --prefix=/usr/local/sysext`. I was considering cases where you need additional executable tools that could be provided by the distro (e.g. jq), but you don't want to assume that they are installed. On reflection, I suppose the `PATH` would need adjusting too. I could qualify it, saying something like "if your extension includes files that might conflict".

chewi:
> No need for --privileged. Should just need `--security-opt writable-cgroups`

Heh, added by my own third line manager, no less. Both Copilot and my own searches hadn't caught up with that yet. Unfortunately, there is apparently no way to set security options via BuildKit. More importantly, as I said above, the lack of dependencies means you will probably be limited to basic file checks, which you can already do using the container build.

website/docs/system-extensions.md
```diff
@@ -0,0 +1,73 @@
...
+
+In the above example, jq also requires `libonig.so.5` at the time of writing. This library is unlikely to be present on most systems, so it should either be installed before merging or included with the extension by adding `oniguruma` under `sysext`. Do whatever is most appropriate for your use case.
+
+### Conflicts
+
+Files on the running system will be overshadowed by those in the system extension if they have the same path. Care must therefore be taken to ensure you don't break the running system by including incompatible files. For example, it is vitally important that you don't include glibc unless it has been installed to an alternative path. Note that the running system is not modified by the merge, and the original files become visible again when unmerging.
+
+### Static linking
+
+The above issues can be largely avoided by including statically-linked rather than dynamically-linked binaries. This will make your system extension much more portable. However, binaries shipped by distributions will practically always be dynamic, so this will require you to build the binaries and their dependencies yourself as part of the spec.
```
pmengelbert: Not a request for changes so much as thinking out loud: it would be interesting to see how well Dalec could automate building binaries and all dependencies statically. IIRC the [wolfi](https://github.com/wolfi-dev) distro produced by chainguard is made for running static binaries, could possibly be used to build them as well.

website/docs/system-extensions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Prefixing
+
+Conflicts may still occur even when static linking. The extension should ideally place its files under a unique prefix, e.g. `/usr/local/my-sysext`. The standard layout may exist below this, e.g. `/usr/local/sysext/bin`. For this to work, the binaries should be built specifically for this prefix. Dynamically-linked binaries should explicitly include somewhere like `/usr/local/sysext/lib` in the `RUNPATH` to avoid the need for wrappers that set `LD_LIBRARY_PATH`.
+
+### Permitted paths
+
+System extensions will only merge files under `/usr` and `/opt`. Files located elsewhere will be not appear once merged. systemd supports configuration extensions, known as confexts, for handling `/etc`, but these are not yet supported by Dalec. In the meantime, anything installed under `/etc` by the spec is relocated under `/usr/share/<NAME>/etc`, and a tmpfiles.d entry is created to copy these files to `/etc` when the extension is merged. They are not removed when the extension is unmerged.
```
pmengelbert: nit: change "will be not appear" to "will not appear"

website/docs/system-extensions.md
chewi: Yeah, I vaguely know of wolfi. I'm personally more interested in Gentoo Prefix, which doesn't do static builds, but does build entire environments into a prefix. I gather Guix (and maybe NixOS?) can do this too.

website/docs/system-extensions.md
jepio: How about dropping it altogether. The only usecase we currently have for sysexts in DALEC relies on packaging existing binary packages.

deps.go
chewi: The sysext test now installs jq and zsh, checks that those binaries are in the sysext, and checks that bash and ls (runtime deps) are not.

website/docs/system-extensions.md
chewi: This documentation is for the public, who may have other use cases that we should consider. I've now adjusted the wording.

We're not really using existing binary packages either, that only affectively happens because of the BuildKit cache. Our case is easy because we're only dealing with static Go binaries that we know won't be present on the target.

pmengelbert:
I'd like @cpuguy83 to weigh in here. @chewi and I were talking offiline about the possibility of merging this without testing that the `sysext` EROFS archives work with `systemd-sysext`.

My POV is that we cannot merge it without such tests, **unless** the feature is marked sufficiently experimental to the user without them having to reference the docs to know it's experimental. I proposed 2 solutions:

a. add a feature gate, and disallow use of the feature unless an environment variable is provided, *or*
b. nest the target under `azlinux3/experimental/sysext` until the additional tests have been added

Either option makes it clear to the user that this feature is use-at-your-own-risk and is not as fully tested as other artifacts that dalec produces.

test/linux_target_test.go
```diff
@@ -666,6 +669,517 @@ echo "$BAR" > bar.txt
...
+				t.Fatalf("error extracting sysext: %v", evalErr)
+			}
+
+			for _, file := range []string{"/usr/bin/jq", "/usr/bin/zsh"} {
+				_, statErr = ref.StatFile(ctx, gwclient.StatRequest{Path: file})
+				if statErr != nil {
+					t.Fatalf("expected file in sysext not found: %v", statErr)
+				}
+			}
+
```
pmengelbert: Can you add a check showing that dependencies were not included? `jq` should have a predictable dependency for which you can do an absence check.

pmengelbert (changes requested):
The tests new LGTM except I'd like to see verification that the dependencies are not installed.

Will await @cpuguy83's thoughts on the additional testing.

Docs look excellent.

test/linux_target_test.go
chewi: I have replaced jq with zstd, which depends on libz.so.1. This is less likely to change than libonig.so.5. I still have to check for libz.so.1 is different locations though.

chewi:
I have added a commit to move the */sysext targets to */testing/sysext while this remains experimental. That commit can be easily reverted later.

adamperlin (commented):
Just following up on what @pmengelbert said here, I think I agree on pretty much all counts.
I think the current tests are reasonable at a baseline level for ensuring that the generated erofs includes/do not include what it is supposed to.

I think it is best practice and probably sufficient to mark this as under `testing` until we are able to add code in Dalec itself to ensure the generated extensions work with `systemd-sysext`, unless that truly proves infeasible. If we merge this as-is, I'd want that investigation/work to be prioritized in the near future.

website/docs/examples/targets.md
```diff
@@ -5,6 +5,7 @@ azlinux3/rpm                     Builds an rpm and src.rpm.
 azlinux3/rpm/debug/buildroot     Outputs an rpm buildroot suitable for passing to rpmbuild.
 azlinux3/rpm/debug/sources       Outputs all the sources specified in the spec file in the format given to rpmbuild.
 azlinux3/rpm/debug/spec          Outputs the generated RPM spec file
+azlinux3/testing/sysext          Builds a systemd system extension for testing purposes only.
```
cpuguy83, line 8: FWIW, I think this can go into something like `<worker>/sysext/erofs`
Later I think we could likely come up with an artifact type that can easily pushed to an OCI registry directly and output that at `<worker>/sysext`

cpuguy83: Given this is all up in the air, having this in `testing/sysext` I guess makes sense for now.

cpuguy83 (approved):

cpuguy83:
(we'll see if that works ^)

chewi:
If not, I'll file it tomorrow. Need to go now.

PR #738, opened 2025-08-22
Update SECURITY.md to point at aka.ms/SECURITY.md

bridgetkromhout:

Microsoft is standardizing the SECURITY.md file; this copies in the text from https://github.com/microsoft/repo-templates/blob/main/shared/SECURITY.md?plain=1 which is what new repos are generated with as of July 16th, 2025.

Note that the content on https://aka.ms/SECURITY.md is largely the same as what this file previously contained.


pmengelbert (approved):
Thanks

Issue #742, opened 2025-08-28
[REQ] Exponential backoff retry logic


MorrisLaw:
### What kind of request is this?

None

### What is your request or suggestion?

Implement retries with exponential backoff for pulling packages.

### Are you willing to submit PRs to contribute to this feature request?

- [x] Yes, I am willing to implement it.

MorrisLaw:
Closing this issue because there is another way to address the problem this is looking to solve, in addition to what Brian pointed out above.

cpuguy83:
I'll also note that clients can customize the workers being used (something we do internally already) where retry logic can be customized.

cpuguy83:
Let's re-open as it was pointed out that neither apt nor yum is doing any sort of sane backoff strategy here and we can wrap the execution in a retry loop.

MorrisLaw:
> Let's re-open as it was pointed out that neither apt nor yum is doing any sort of sane backoff strategy here and we can wrap the execution in a retry loop.

Could this look something like wrapping `WithRunOptions()` in a `WithRetry()` passing in the `opts ...llb.RunOption` and `maxRetries int` as args? That way where `WithRunOptions(runOpts...)` is called we can update to calling `WithRetry(runOpts..., 3)` instead in the `TdnfInstall()` and `DnfInstall()` functions in dnf_install.go.

PR #745, opened 2025-09-02
dnf: remove manual metadata refresh

cpuguy83:
dnf automatically refreshes repo metadata after a configured expiry period.
This change removes our manual metadata refreshes (which we end up doing twice per package install command).

This had some extra implications around custom repos in CI.
To work around this I have namespaced all custom repo ID's with a random string in order to prevent caches from mixing.

This also spotted an issue with the cache for almalinux.
Alma does not include the $basearch in the mirrorlist url, which means that when dnf computes a cache key that gets used in /var/cache/dnf it is not unique per arch which broke the cross_platform test for alma.
To fix this I added an option to the rpm distro config to namespace the buildkit cache key for the dnf cache with the target platform.

Addresses part of #742

### DIFF
--- a/test/createrepo.go
+++ b/test/createrepo.go
@@ -22,6 +22,7 @@ baseurl=file://` + repoPath + `
 gpgcheck=0
 priority=0
 enabled=1
+metadata_expire=0
 `)
 
 			pg := dalec.ProgressGroup("Install local repo for test")
--- a/test/createrepo.go
+++ b/test/createrepo.go
@@ -1,18 +1,24 @@
 package test
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
+	"path/filepath"
+
 	"github.com/Azure/dalec"
 	"github.com/Azure/dalec/targets/linux/rpm/distro"
 	"github.com/moby/buildkit/client/llb"
 )
 
-func createYumRepo(installer *distro.Config) func(rpms llb.State, opts ...llb.StateOption) llb.StateOption {
-	return func(rpms llb.State, opts ...llb.StateOption) llb.StateOption {
+func createYumRepo(installer *distro.Config) func(rpms llb.State, repoPath string, opts ...llb.StateOption) llb.StateOption {
+	return func(rpms llb.State, repoPath string, opts ...llb.StateOption) llb.StateOption {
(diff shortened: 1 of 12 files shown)


MorrisLaw (approved):
yeah this makes sense, lgtm! 🚢

cpuguy83:
Yay, all green now!
Updated the PR description.
Going to make one more change to document (in the code) why alma needs to namespace the cache key for the repos.

test/createrepo.go
```diff
@@ -1,35 +1,42 @@
...
 	"github.com/Azure/dalec/targets/linux/rpm/distro"
 	"github.com/moby/buildkit/client/llb"
 )
 
-func createYumRepo(installer *distro.Config) func(rpms llb.State, opts ...llb.StateOption) llb.StateOption {
-	return func(rpms llb.State, opts ...llb.StateOption) llb.StateOption {
+func createYumRepo(installer *distro.Config) func(rpms llb.State, repoPath string, opts ...llb.StateOption) llb.StateOption {
+	return func(rpms llb.State, repoPath string, opts ...llb.StateOption) llb.StateOption {
 		return func(in llb.State) llb.State {
+			suffixBytes := sha256.Sum256([]byte(repoPath))
```
MorrisLaw, line 16: Do you think a crypto hash is necessary here? IIUC, this appears to be necessary for uniqueness but not cryptographic security.

If that's true, then can we consider using a non-cryptographic hash? Mainly for this [reason](https://eng.ms/docs/microsoft-security/digital-security-and-resilience/sr-assurance/cspa/sdl-home/sdl-home/cryptoboard/cryptoboardfaq), as defined by the crypto board:

```
Additionally, note that while using approved crypto for non-security uses is permitted, if that crypto becomes banned in the future, you will be required to change it. It's recommended for this reason that if a situation does not require security, do not use cryptographic algorithms to solve it.

Examples of non-cryptographic hashes which can be used where non-security hashing is required include:

SpookyHash
[xxHash](https://aka.ms/xxHash64)
CityHash
MurMurHash
Marvin32
```

cpuguy83: This is all fine, it is test code and we don't use msft go for this either.

PR #756, opened 2025-09-12
testrunner: Simplify by moving test execution inside llb.Run

cpuguy83:
This makes it so the test evaluator runs *inside* the container. This means we don't need to do crazy things with stdio routing nor do we need to go through the buidlkit API for every file request.


internal/testrunner/files.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+const CheckFilesCmdName = "test-checkfiles"
+
+// CheckFilesCmd is the entrypoint for the file checking subcommand.
+// It reads the file checks from the provided file path (first argument)
+// and executes them, writing output to os.Stdout and os.Stderr.
+//
+// This should only be called from inside a container where the test is meant to run.
+func CheckFilesCmd(args []string) {
```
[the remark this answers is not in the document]
cpuguy83: In this case I'd rather have the command panic because this would truly be an bug in the core dalec code.

internal/testrunner/step.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+const StepRunnerCmdName = "test-steprunner"
+
+// StepCmd is the entrypoint for the test step runner subcommand.
+// It reads the test step from the provided file path (first argument)
+// and executes it, writing output to os.Stdout and os.Stderr.
+//
+// This should only be called from inside a container where the test is meant to run.
+func StepCmd(args []string) {
```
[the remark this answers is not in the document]
cpuguy83: In this case I'd rather have the command panic because this would truly be an bug in the core dalec code.

frontend/test_runner.go
```diff
@@ -60,16 +62,20 @@ func RunTests(ctx context.Context, client gwclient.Client, spec *dalec.Spec, ref
...
 	}
 
-	type testPair struct {
-		st     llb.State
-		t      *dalec.TestSpec
-		stdios map[int]llb.State
-		opts   []llb.ConstraintsOpt
+	ctrWithDeps := ctr.With(withTestDeps)
+
+	fSt, err := client.(frontendClient).CurrentFrontend()
```
MorrisLaw: Should we use [GetCurrentFrontend(client)](https://github.com/Azure/dalec/blob/f604f38ff1bd447624e1e04aa39337f6f9ab64b7/frontend/build.go#L219) instead?

frontend/test_runner.go
```diff
@@ -42,11 +44,11 @@ func RunTests(ctx context.Context, client gwclient.Client, spec *dalec.Spec, ref
 		return nil
 	}
 
+	// Force evaluation here so that any errors for the build itself can surface
```
MorrisLaw, line 47: Can't see it here, but since you're in the file can you please update the comment to `RunTests` instead of `Run tests`: https://github.com/Azure/dalec/blob/8a9d65bde9a122ed5c9c04190b7e980c3037b092/frontend/test_runner.go#L23

cpuguy83: Fixed.

frontend/test_runner.go
cpuguy83: Done.

PR #769, opened 2025-09-30
feat: go mod replace and require

sozercan:

adds go mod dependency substitution via replace and require functionality

Fixes #


spec.go
```diff
@@ -240,6 +243,145 @@ type GeneratorGomod struct {
...
+	// Require applies go.mod require directives before downloading module dependencies.
+	// Each entry must be of the form "module:target@version" to match the syntax accepted by
+	// `go mod edit -require`.
+	Require []GomodRequire `yaml:"require,omitempty" json:"require,omitempty"`
+}
+
+// GomodReplace represents a go.mod replace directive in shorthand "old:new" form.
+// This allows users to substitute module dependencies before go mod download runs,
+// useful for pointing to local forks or alternate versions.
+type GomodReplace string
```
cpuguy83: Since we have to parse this anyway, seems like this should be a struct?

spec.go
```diff
@@ -240,6 +243,145 @@ type GeneratorGomod struct {
 	Paths []string `yaml:"paths,omitempty" json:"paths,omitempty"`
 	// Auth is the git authorization to use for gomods. The keys are the hosts, and the values are the auth to use for that host.
 	Auth map[string]GomodGitAuth `yaml:"auth,omitempty" json:"auth,omitempty"`
+	// Replace applies go.mod replace directives before downloading module dependencies.
+	// Each entry must be of the form "old:new" where "old" and "new" match the syntax accepted by
+	// `go mod edit -replace`.
+	Replace []GomodReplace `yaml:"replace,omitempty" json:"replace,omitempty"`
```
cpuguy83, line 256: The gomod generator supports supplying multiple subpaths for different modules.
It doesn't seem like the implementation here can deal with that.

cpuguy83: Maybe good to move these under a new field like `edits`

spec.go
cpuguy83: Totally fine to if we want to be able to supply a string to have the unmarshaller look if the value is a string or mapping and handle accordingly.

DannyBrito:
> @sozercan @cpuguy83 Is this ready for another round of review? Just wondering as I'm working on a follow-up to this

I have #825 with some rebase but not much progress. There some failure in the tests. But unfortunately haven't had many cycles to continue this.

PR #773, opened 2025-10-01
Source maps

cpuguy83:
Add source maps to errors and LLB definitions

This makes it so when an error is returned the error carries with it
details about where in the spec file the error occurred.

To accomplish, each object that wants to provide detailed location info
needs to implement a custom UnmarshalYAML method which takes the
location from the yaml AST and stores it so it can be retrieved later as
needed.

Future work here would involve providing even more detailed source
mappings in some cases (e.g. a source type that has a "digest" field,
which is a type we don't own so can't just store the mapping to it
easily like we do for owned types).

Closes #756 (includes the changes from there already, and some additional ones).

### DIFF
--- a/cmd/gen-source-variants/main.go
+++ b/cmd/gen-source-variants/main.go
@@ -70,8 +70,9 @@ func extractSourceFields() ([]SourceField, error) {
 				break
 			}
 
+			const sourceTypeRef = "Source"
 			ast.Inspect(node, func(n ast.Node) bool {
-				if ts, ok := n.(*ast.TypeSpec); ok && ts.Name.Name == "Source" {
+				if ts, ok := n.(*ast.TypeSpec); ok && ts.Name.Name == sourceTypeRef {
 					found = true
 					if st, ok := ts.Type.(*ast.StructType); ok {
 						for _, field := range st.Fields.List {
@@ -85,7 +86,7 @@ func extractSourceFields() ([]SourceField, error) {
 							if ptr, ok := field.Type.(*ast.StarExpr); ok {
 								if ident, ok := ptr.X.(*ast.Ident); ok {
 									typeName := ident.Name
-									if strings.HasPrefix(typeName, "Source") {
+									if strings.HasPrefix(typeName, sourceTypeRef) {
 										sourceFields = append(sourceFields, SourceField{
 											Name:     fieldName,
 											TypeName: typeName,


internal/testrunner/files.go
```diff
@@ -0,0 +1,183 @@
...
+	for path, check := range files {
+		if err := checkFile(path, check); err != nil {
+			results = append(results, FileCheckErrResult{Filename: path, Checks: getFileCheckErrs(err)})
+		}
+	}
+	return results
+}
+
+func checkFile(p string, check dalec.FileCheckOutput) error {
+	stat, err := os.Lstat(p)
```
MorrisLaw, line 97: Out of curiosity (I'm sure there's a good reason) why don't we want to get the file info with `os.Stat`, even if it ends up being the target of a symlink?

packaging/linux/deb/template_control.go
```diff
@@ -130,6 +122,9 @@ func (w *controlWrapper) depends(buf *strings.Builder, depsSpec *dalec.PackageDe
 				needsClone = false
 			}
```
MorrisLaw, line 124: **super nit:** extra whitespace that makes it not match the similar block below
```suggestion
```

targets/linux/rpm/distro/debug.go
```diff
@@ -38,8 +39,8 @@ func (c *Config) DebugWorker(ctx context.Context, client gwclient.Client, spec *
 		hasRust := func(s string) bool {
 			return s == "rust"
 		}
-		if !slices.ContainsFunc(deps, hasRust) {
-			return llb.Scratch(), errors.New("spec contains go modules but does not have golang in build deps")
+		if !slices.ContainsFunc(pkgNames, hasRust) {
+			return llb.Scratch(), errors.New("spec contains cargo homes but does not have rust in build deps")
```
MorrisLaw, line 43: oh wow, nice catch lol

cmd/gen-source-variants/main.go
```diff
@@ -51,41 +51,59 @@ type SourceField struct {
 
 func extractSourceFields() ([]SourceField, error) {
 	fset := token.NewFileSet()
-	node, err := parser.ParseFile(fset, "spec.go", nil, parser.ParseComments)
+	packages, err := parser.ParseDir(fset, ".", nil, parser.ParseComments)
```
MorrisLaw, line 54: could it be helpful to make the directory passed in here configurable?

cmd/gen-source-variants/main.go
```diff
@@ -51,41 +51,59 @@ type SourceField struct {
...
+			break
+		}
+		for _, node := range pkg.Files {
+			// Find the Source struct
+			if found {
+				break
+			}
+
+			ast.Inspect(node, func(n ast.Node) bool {
+				if ts, ok := n.(*ast.TypeSpec); ok && ts.Name.Name == "Source" {
```
MorrisLaw: nit: could make "Source" a constant so we don't inadvertently add a typo in one of the spots at some point.

`const sourceTypeName = "Source"` or something like that maybe.

internal/testrunner/files.go
cpuguy83: We actually want to be able to validate that something is a symlink or not. We aren't doing it right now, but that is the desire.

cmd/gen-source-variants/main.go
cpuguy83: I don't think so, because this is always run from `go generate` which would be executed from the context of the file that has the `//go:generate` pragma.

cmd/gen-source-variants/main.go
cpuguy83: Updated to use the const.

PR #775, opened 2025-10-01
rpm: Fix issue where ignore-cache is not respected

cpuguy83:
Makes sure `--no-cache` and `--no-cache-filter` is respected for rpm-based targets.


sozercan (approved):
lgtm

PR #784, opened 2025-10-08
Add user support for artifacts

Stevenjin8:

Adds user support for artifacts.

I was also planning on changing `chmod` to `chmod -R`, but this more nuanced than it seems since directories usually  set the `x` bit, but files almost never want it.

### DIFF
--- a/test/main_test.go
+++ b/test/main_test.go
@@ -30,6 +30,7 @@ func TestMain(m *testing.M) {
 		externalTestHost = "https://github.com"
 	}
 	flag.StringVar(&externalTestHost, "external-test-host", externalTestHost, "http server to use for validating network access")
+	flag.BoolVar(&testenv.InlineBuildOutput, "inline-build-output", testenv.InlineBuildOutput, "Stream the test's build output as it happens instead of waiting until the end. Note: parallel tests will interleave output.")
 
 	flag.Parse()
 
--- a/test/testenv/build.go
+++ b/test/testenv/build.go
@@ -194,8 +194,6 @@ func getBuildOutputStream(ctx context.Context, t *testing.T, done <-chan struct{
 		return tw
 	}
 
-
-
 	dir := t.TempDir()
 	f, err := os.OpenFile(filepath.Join(dir, "build.log"), os.O_RDWR|os.O_CREATE, 0o600)
 	if err != nil {


packaging/linux/rpm/template.go
```diff
@@ -526,6 +535,119 @@ func (w *specWrapper) postGroups() string {
...
+		targetDir := filepath.Join(root, cfg.SubPath)
+		var targetPath string
+		file := cfg.ResolveName(p)
+		if !strings.Contains(file, "*") {
+			targetPath = filepath.Join(targetDir, file)
+		} else {
+			targetPath = targetDir + "/"
+		}
+		if user != "" {
+			fmt.Fprintf(b, "chown -R %s %s\n", user, targetPath)
```
cpuguy83, line 589: For rpms we can use the `%attr` macro to set ownership.

```
%attr(mode, user, group) filename
```

You can use the default values by setting the value to `-`, e.g. `%attr(-, www-data, www-data)`
This goes into the entry under `%files`

Stevenjin8: I don't think `attr` is recursive. So we can't do the `-R` in `chmod -R`.

Stevenjin8: @cpuguy also, it seems tests run before the postinstall script is run... not sure how to test properly

cpuguy83: > also, it seems tests run before the postinstall script is run... not sure how to test properly

I'm not sure I understand this.
What are you trying to test and why does it matter if its in postinstall or not matter?

> I don't think attr is recursive. So we can't do the -R in chmod -R

It should work with globs, though?

The main reason to prefer `%attr` is I don't think this requires a shell at runtime which makes it much nicer for (minimal) container builds.

cpuguy83: One option for the recursive case would be to move all directory-type entries to the end and use `%defattr`.

Stevenjin8: ok nice

Stevenjin8: @cpuguy83 i tried to implement this and I don't think it really works.

First, to create a user, we need to run a pre/post install script anyways, so the only time where `attr` would be any different over `chown` would be when giving ownership to a preexisting user.

Second, because we create users postinstall, symlink ownership doesn't work ATM.

So I don't see any functional difference between doing it with `attr` vs `chmod`. They would both require shells in almost every that has to do with minimal container builds in dalec

cpuguy83: Fair point re: user/group creation.

cpuguy83:
> I can't figure out why the tests are failing though... they pass locally

Rockylinux repos being flakey (probably in this case repos being updated).
Works locally because the packages are probably cached on your machine.

PR #793, opened 2025-10-15
Added service start on install, disable/stop on uninstall.

mipresmsft:
Adds ability to start service on install.

Correctly stops/disables service prior to uninstall.

### DIFF
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -2203,8 +2203,7 @@
 		},
 		"SystemdUnitConfig": {
 			"required": [
-				"name",
-				"enable"
+				"name"
 			],
 			"properties": {
 				"enable": {
--- a/systemd.go
+++ b/systemd.go
@@ -21,7 +21,7 @@ type SystemdUnitConfig struct {
 
 	// Enable is used to enable the systemd unit on install
 	// This determines what will be written to a systemd preset file
-	Enable bool `yaml:"enable" json:"enable"`
+	Enable bool `yaml:"enable,omitempty" json:"enable,omitempty"`
 
 	// Start can be optionally used to start the systemd unit on install
 	// This determines what will be written to a systemd preset file


docs/spec.schema.json
```diff
@@ -2204,7 +2204,8 @@
 		"SystemdUnitConfig": {
 			"required": [
 				"name",
-				"enable"
+				"enable",
+				"start"
 			],
```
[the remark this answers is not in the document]
mipresmsft: Nice catch.

packaging/linux/rpm/template.go
```diff
@@ -435,21 +435,40 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
 	return b
 }
 
+func systemdPreUnScript(unitName string, cfg dalec.SystemdUnitConfig) string {
+	// if service isn't explicitly specified as enabled in the spec,
+	// then we don't need to do anything in the preun script
+	if !cfg.Enable {
```
cpuguy83, line 441: Do we need to check if it should be stopped before uninstall even if its not enabled?

packaging/linux/rpm/template.go
```diff
@@ -462,12 +481,24 @@ func systemdPostScript(unitName string, cfg dalec.SystemdUnitConfig) string {
...
 	// but without the use of a .preset file
-	return fmt.Sprintf(`
+	s := `
 if [ $1 -eq 1 ]; then
-    # initial installation
-    systemctl enable %s
+    # initial installation`
+
+	// Enable/start service when package is installed
+	if cfg.Start {
```
cpuguy83, line 489: I try to reduce if/else as much as possible for readability.
Maybe this could be reduced to something like `fmt.Sprintf("systemctl enable %s %s", enableFlags, unitName)`

cpuguy83: Actually thinking about this more using `--now` would require a connection to systemd which could cause this to fail when installing in, e.g., a container, where as just a normal `systemctl enable` creates symlinks only, and probably even the existing code is not right (Even if it works without systemd available)

Instead should be something like this:

```
if systemctl >/dev/null 2>&1 && systemctl is-system-running >/dev/null 2>&1; then
    systemctl enable --now myservice.service || true
fi
```

It would also be good to note on the fields (for both enable and start) that use of these fields is discouraged as it is typically expected that higher level tooling handles when a unit should be started and/or enabled.

packaging/linux/rpm/template.go
```diff
@@ -435,21 +435,40 @@ func (w *specWrapper) BuildSteps() fmt.Stringer {
...
+	if !cfg.Enable {
+		return ""
+	}
+
+	// should be equivalent to the systemd_preun scriptlet in the rpm spec,
+	// but without the use of a .preset file
+	return fmt.Sprintf(`
+if [ $1 -eq 0 ]; then
+    # complete uninstallation
+    systemctl disable --now %s
```
cpuguy83, line 450: Same comment as https://github.com/Azure/dalec/pull/793/files#r2433340247

packaging/linux/rpm/template.go
mipresmsft: I tested both scenarios and using "--now" is a no-op if the service isn't currently running.

packaging/linux/rpm/template.go
mipresmsft: @cpuguy83 Well, if the user doesn't want their service started, then they can just not use this feature, right?  They can use "systemctl start" as a separate step.  The bigger issue is on uninstall.  If the user doesn't stop/disable the service before uninstalling, it leaves the service in a limbo state.

packaging/linux/rpm/template.go
mipresmsft: Same answer.  🙂

packaging/linux/rpm/template.go
mipresmsft: @cpuguy83 Also, what do you mean by "higher-level tooling"?  My use case is for when the service enable/start needs to be done during AKS VM first-time installation.

cpuguy83: A preset file (provided by the admin) or other system configuration tool.
Basically, give the control to the admin to deal with it rather than the package itself.

mipresmsft: @cpuguy83 Understood.  Like I said, in that case, the package doesn't need to use the start field or set the enable field to true.  This is just for the case where we want the package to do it automatically.  In my case, adding "start" to the package isn't a big issue because I could just call "systemctl start" immediately after installing the package.  However, I still think the stop/disable should be considered for the cases where we want to uninstall the package.

mipresmsft: Enable has always been a required parameter, but I made start optional.

cpuguy83: Yeah I agree its fine to have the functionality, just to note on the spec field that depending on distribution there may be different expectations on different distributions as to if the package should be doing this.

mipresmsft: Sounds good.  Added a detailed explanation in the spec comments.

systemd.go
```diff
@@ -22,6 +22,12 @@ type SystemdUnitConfig struct {
 	// Enable is used to enable the systemd unit on install
 	// This determines what will be written to a systemd preset file
 	Enable bool `yaml:"enable,omitempty" json:"enable"`
+
+	// Start can be optionally used to start the systemd unit on install
+	// This determines what will be written to a systemd preset file
+	// Note that depending on distribution there may be different
+	// expectations as to if the package should be responsible for this
+	Start bool `yaml:"start,omitempty" json:"start"`
 }
```
mipresmsft, line 31: @cpuguy83 Should the yaml and json "omitempty" values be consistent?  I'm having trouble with the lint validation of the required parameters.  Running "go generate" shows that the fields in spec.schema.json are coming from the json values here, not the yaml values.  I can change the json value for Start to "start,omitempty", but I'm wondering if the yaml "omitempty" values are correct.

cpuguy83: Yes, it should be the same, looks like the json value for other unit config fields was overlooked.
However, the lint error is because you need to re-run `go generate ./...`
There should be a decent error message with that but for some reason it seems like the script running like it has `set -e`.

cpuguy83: Addressing error message here: https://github.com/Azure/dalec/pull/795

mipresmsft: Right.  Running "go generate" was how I noticed the discrepancy.  Let me know if it looks right now.  I removed "omitempty" from the yaml values for Name and Enable.

systemd.go
```diff
@@ -17,11 +17,17 @@ type SystemdUnitConfig struct {
 	// Name is the name systemd unit should be copied under.
 	// Nested paths are not supported. It is the user's responsibility
 	// to name the service with the appropriate extension, i.e. .service, .timer, etc.
-	Name string `yaml:"name,omitempty" json:"name"`
+	Name string `yaml:"name" json:"name"`
 
 	// Enable is used to enable the systemd unit on install
 	// This determines what will be written to a systemd preset file
-	Enable bool `yaml:"enable,omitempty" json:"enable"`
+	Enable bool `yaml:"enable" json:"enable"`
```
cpuguy83: This should be omitempty.

mipresmsft: @cpuguy83 The current spec has "enable" as required.  Do you want to remove that?  See the related lines in spec.schema.json.

cpuguy83: Yeah, it shouldn't be required.

mipresmsft: @cpuguy83 OK.  Done.

cpuguy83 (approved):

PR #794, opened 2025-10-15
tests: Add option to inline the build output to the test log

cpuguy83:
When integration running tests (go test ./test), normally the test runner will only output build logs (ie, logs from the buildkit solve(s) executed by a given test) at the end of a test.
With this change you can now provide a flag, `--inline-build-output`, which will make build output output directly to the test log (by line) instead of defering the output until the end of the test.

As noted in the flag description, when running multiple tests (which are generally run with t.Parallel) the build outputs for multiple tests may be interleaved, hence why this is disabled by default.

### DIFF
--- a/test/testenv/build.go
+++ b/test/testenv/build.go
@@ -2,11 +2,13 @@ package testenv
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
 	"io"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 
 	"github.com/moby/buildkit/client"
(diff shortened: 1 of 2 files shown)


test/testenv/build.go
```diff
@@ -170,6 +223,19 @@ func displaySolveStatus(ctx context.Context, t *testing.T) chan *client.SolveSta
 		}
 	})
 
+	return f
+}
+
+func displaySolveStatus(ctx context.Context, t *testing.T) chan *client.SolveStatus {
```
MorrisLaw, line 229: IIUC, the caller would only need to read from this channel. So maybe we can enforce some type safety here?

```suggestion
func displaySolveStatus(ctx context.Context, t *testing.T) <-chan *client.SolveStatus {
	ch := make(chan *client.SolveStatus)
	done := make(chan struct{})

	output := getBuildOutputStream(ctx, t, done)
	display, err := progressui.NewDisplay(output, progressui.AutoMode, progressui.WithPhase(t.Name()))
	if err != nil {
		t.Fatal(err)
	}

```

cpuguy83: The caller needs `chan *client.SolveStatus`.
It gets passed to the buildkit client which is what writes to the channel.

MorrisLaw: Ahh ok

PR #815, opened 2025-10-24
Add capabilities to artifacts

Stevenjin8:

This allows for setting capabilities for binaries in artifacts. This is helpful in contexts that do not allow for privilege escalation.

Fixes #

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -136,7 +136,7 @@ type ArtifactConfig struct {
 	// Group is the group name that should own the artifact
 	Group string `yaml:"group,omitempty" json:"group,omitempty"`
 	// LinuxCapabilities is the list of Linux capabilities to set on the artifact
-	LinuxCapabilities []ArtifactCapability `yaml:"capabilities,omitempty" json:"capabilities,omitempty"`
+	LinuxCapabilities []ArtifactCapability `yaml:"linux_capabilities,omitempty" json:"linux_capabilities,omitempty"`
 }
 
 // ArtifactCapability represents a Linux Capability to set on an artifact
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -143,7 +143,14 @@
 		},
 		"ArtifactConfig": {
 			"properties": {
-				"capabilities": {
+				"group": {
+					"type": [
+						"string",
+						"null"
+					],
+					"description": "Group is the group name that should own the artifact"
+				},
+				"linux_capabilities": {
 					"items": {
 						"$ref": "#/$defs/ArtifactCapability"
 					},
@@ -151,13 +158,6 @@
 						"array"
 					],
 					"description": "LinuxCapabilities is the list of Linux capabilities to set on the artifact"
-				},
-				"group": {
-					"type": [
-						"string",
-						"null"
-					],
-					"description": "Group is the group name that should own the artifact"
 				},
 				"name": {
 					"type": [


packaging/linux/deb/debroot.go
```diff
@@ -728,3 +729,33 @@ func setArtifactOwnershipPostInst(w *bytes.Buffer, spec *dalec.Spec, target stri
...
+			p := filepath.Join(root, cfg.SubPath, resolved)
+			fmt.Fprintf(w, "setcap %s \"$DESTDIR%s\"\n", cfg.Capabilities, p)
+		}
+	}
+
+	apply(artifacts.Binaries, BinariesPath)
+	apply(artifacts.ConfigFiles, ConfigFilesPath)
+	apply(artifacts.Manpages, filepath.Join(ManpagesPath, spec.Name))
+	apply(artifacts.Headers, HeadersPath)
+	apply(artifacts.Licenses, filepath.Join(LicensesPath, spec.Name))
```
MorrisLaw: I think copilot is right about not needing to set something like "capabilities" on non-binaries.

docs/spec.schema.json
```diff
@@ -104,6 +104,13 @@
 		},
 		"ArtifactConfig": {
 			"properties": {
+				"capabilities": {
+					"type": [
+						"string",
+						"null"
+					],
+					"description": "Capabilities is the set of capabilities to set on the artifact (e.g., \"cap_net_bind_service=+ep\")"
```
MorrisLaw: TBC, for my understanding... these are specifically [Linux capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html)?

Stevenjin8: yes

packaging/linux/deb/debroot.go
Stevenjin8: its a little weird, but you could say the same thing about making files here executable, and that's perfectly allowed. So it seems inconsistent to be able to make a license or manpage executable, but not give it any capabilities

MorrisLaw: that's a fair point

Stevenjin8: You could make the argument the other way around though, caps are only relevant to actual binaries while permissions are relevant to everything. Up to you, I don't feel strongly

artifacts.go
```diff
@@ -132,6 +132,8 @@ type ArtifactConfig struct {
 	User string `yaml:"user,omitempty" json:"user,omitempty"`
 	// Group is the group name that should own the artifact
 	Group string `yaml:"group,omitempty" json:"group,omitempty"`
+	// Capabilities is the set of POSIX capabilities to set on the artifact (e.g., "cap_net_bind_service=+ep")
+	Capabilities string `yaml:"capabilities,omitempty" json:"capabilities,omitempty"`
```
Stevenjin8: there might be a better way to express this.

jeremyrickard: Seems like there is an undocumented feature for RPM like

> %caps(cap_net_admin=pe) %{_sbindir}/foobar

This is probably expressive. would we want to make this a slice of capabilities? Would there be cases where we would want to specify multiple and not have a single string?

packaging/linux/deb/debroot.go
jeremyrickard: The undocumented macro for RPM just lands in the files section. I think this is ok, document that it isn;t supported?

packaging/linux/deb/debroot.go
```diff
@@ -728,3 +729,33 @@ func setArtifactOwnershipPostInst(w *bytes.Buffer, spec *dalec.Spec, target stri
...
+		}
+		sorted := dalec.SortMapKeys(artifacts)
+		for _, key := range sorted {
+			cfg := artifacts[key]
+			if cfg.Capabilities == "" {
+				continue
+			}
+			resolved := cfg.ResolveName(key)
+			p := filepath.Join(root, cfg.SubPath, resolved)
+			fmt.Fprintf(w, "setcap %s \"$DESTDIR%s\"\n", cfg.Capabilities, p)
```
jeremyrickard: can we do this with the RPM macro?

artifacts.go
Stevenjin8: @jeremyrickard 

User/group ownership is done postinstall. `chown` also clears any capabilities:

```
12:06:50 (demo-aks) /home/sj/ingress2gateway $ touch delete
12:06:55 (demo-aks) /home/sj/ingress2gateway $ chmod +x delete
12:07:40 (demo-aks) /home/sj/ingress2gateway $ sudo setcap 'cap_net_bind_service=+ep' delete
12:07:43 (demo-aks) /home/sj/ingress2gateway $ sudo getcap delete
delete cap_net_bind_service=ep
12:07:56 (demo-aks) /home/sj/ingress2gateway $ chown  root delete
chown: changing ownership of 'delete': Operation not permitted
12:08:01 (demo-aks) /home/sj/ingress2gateway $ sudo chown  root delete
12:08:06 (demo-aks) /home/sj/ingress2gateway $ sudo getcap delete
```

So we have to do setcap after chown/chgrp, i.e. the postinstall script.

TBH, I also don't love how much we're relying on postinstall scripts as they end up injecting bash

Stevenjin8: > would we want to make this a slice of capabilities?

this is fair... Right now, you can add commas, which is underwhelming.

packaging/linux/deb/debroot.go
Stevenjin8: see other comment

artifacts.go
Stevenjin8: @jeremyrickard updated the api to be a more intuitive

jeremyrickard: Thanks for clarifying the behavior here.

cpuguy83: How about we go ahead and use the macro, then if there is a chown we do the setcap as well.
This way we only add the postinstall script when absolutely necessary.

artifacts.go
```diff
@@ -132,6 +133,56 @@ type ArtifactConfig struct {
 	User string `yaml:"user,omitempty" json:"user,omitempty"`
 	// Group is the group name that should own the artifact
 	Group string `yaml:"group,omitempty" json:"group,omitempty"`
+	// Capabilities is the list of POSIX capabilities to set on the artifact
+	Capabilities []ArtifactCapability `yaml:"capabilities,omitempty" json:"capabilities,omitempty"`
```
cpuguy83: Nit: LinuxCapabilities

packaging/linux/deb/debroot.go
```diff
@@ -728,3 +729,33 @@ func setArtifactOwnershipPostInst(w *bytes.Buffer, spec *dalec.Spec, target stri
...
+
+	apply(artifacts.Binaries, BinariesPath)
+	apply(artifacts.ConfigFiles, ConfigFilesPath)
+	apply(artifacts.Manpages, filepath.Join(ManpagesPath, spec.Name))
+	apply(artifacts.Headers, HeadersPath)
+	apply(artifacts.Licenses, filepath.Join(LicensesPath, spec.Name))
+	apply(artifacts.Docs, filepath.Join(DocsPath, spec.Name))
+	apply(artifacts.Libs, LibsPath)
+	apply(artifacts.Libexec, LibexecPath)
+	apply(artifacts.DataDirs, DataDirsPath)
```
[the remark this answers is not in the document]
cpuguy83: I tend to agree with this, I think we should add validation before the build that only allows these caps on binaries and libexecs.
If we need to open that up later that's fine.

artifacts.go
cpuguy83: Ping @Stevenjin8

Stevenjin8: I swear I fixed this

Stevenjin8: @cpuguy83 fixed. Sorry about that.

cpuguy83:
Needs a `go generate ./...`

Stevenjin8:
@cpuguy83 I rebased, but [`2373892`](https://github.com/project-dalec/dalec/pull/815/commits/2373892e512fc80024a00210171d0c76d666155a) is the first new commit.

cpuguy83 (changes requested):
Seems to be missing the last change to swtich to linux_capabilities

cpuguy83 (approved):

PR #819, opened 2025-10-24
refresh metadata!

Stevenjin8:

When installing from local repos, dalec will rebuild the repo every time, but keep the metadata cache across runs. This means that on subsequent builds, digests don't match.


cpuguy83:
This is intentionally not manually refreshing dnf metadata so that we don't hammer repos too often since dnf manages its own metadata cache.

For repositories that need more frequent metadata updates, this should be set in the repo config.
We do call this out for local repos in https://project-dalec.github.io/dalec/repositories

PR #849, opened 2025-11-08
Fix test runner wrapping the wrong error

cpuguy83:
In this case the error was always nil, which made tests pass since `errors.Wrap(<nil err>, ...)` will always return nil.


DannyBrito (approved):
lgtm

PR #867, opened 2025-11-21
feat: add Debian Trixie (13) target

bketelsen:
Adds Debian Trixie as a build target, updates tests, docs, and CI.

Fixes #

### DIFF
--- a/.github/workflows/retag/images.yml
+++ b/.github/workflows/retag/images.yml
@@ -13,9 +13,6 @@
 - source: "debian:bookworm"
   dest: mirror/dockerhub/library/debian:bookworm
 
-- source: "debian:trixie"
-  dest: mirror/dockerhub/library/debian:trixie
-
 - source: "ubuntu:bionic"
   dest: mirror/dockerhub/library/ubuntu:bionic
 


bketelsen:
this is chicken/egg isn't it? I added trixie/worker to the ci worker-images matrix. but it doesn't exist.

cpuguy83:
@bketelsen CI is expecting debian:trixie to be in the mirror and its not. Will look at adding this.

bketelsen:
> Trixie mirror is merged and now available. You should be able to remove the mirror changes now..
> 
> Also looks like you need to do a `go generate ./...`

Generate is done and pushed.

Can you be more explicit about what I should remove?

.github/workflows/retag/images.yml
```diff
@@ -13,6 +13,9 @@
 - source: "debian:bookworm"
   dest: mirror/dockerhub/library/debian:bookworm
 
+- source: "debian:trixie"
```
cpuguy83: Can you remove this change now that it is already in main.

bketelsen:
@cpuguy83 should be removed now, PTAL

cpuguy83 (approved):

PR #847, opened 2025-11-06
wip: add go mod require & replace support

DannyBrito:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

PR #868, opened 2025-11-21
add go mod replace support

willie-yao:
Continuation of #847

Fixes #

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -208,12 +208,13 @@ type SymlinkTarget struct {
 // before downloading module dependencies.
 type GomodEdits struct {
 	// Replace applies go.mod replace directives before downloading module dependencies.
-	// Each entry can be either a string "old:new" or a struct with Old and New fields.
+	// Each entry can be either a string "old => new" or a struct with Old and New fields.
 	Replace []GomodReplace `yaml:"replace,omitempty" json:"replace,omitempty"`
 }
 
 // GomodReplace represents a go.mod replace directive.
-// It can be specified as a string "old:new" or as a struct with Original and Update fields.
+// It can be specified as a string "old => new" or as a struct with Original and Update fields.
+// The string format matches Go's native go.mod syntax.
 // This allows users to substitute module dependencies before go mod download runs,
 // useful for pointing to local forks or alternate versions.
 type GomodReplace struct {
@@ -226,7 +227,7 @@ type GomodReplace struct {
 }
 
 func (r GomodReplace) String() string {
-	return r.Original + ":" + r.Update
+	return r.Original + " => " + r.Update
 }
 
 func (r GomodReplace) goModEditArg() (string, error) {
@@ -243,9 +244,9 @@ func (r *GomodReplace) UnmarshalYAML(ctx context.Context, node ast.Node) error {
 		if err := yaml.NodeToValue(node, &raw, decodeOpts(ctx)...); err != nil {
 			return err
 		}
-		parts := strings.SplitN(raw, ":", 2)
+		parts := strings.SplitN(raw, " => ", 2)
 		if len(parts) != 2 {
-			return errors.Errorf("invalid gomod replace %q, expected format old:new", raw)
+			return errors.Errorf("invalid gomod replace %q, expected format 'old => new'", raw)
 		}
 		r.Original = strings.TrimSpace(parts[0])
 		r.Update = strings.TrimSpace(parts[1])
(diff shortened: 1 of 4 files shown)


targets/windows/handle_zip.go
```diff
@@ -164,7 +164,7 @@ func buildBinaries(ctx context.Context, spec *dalec.Spec, worker llb.State, clie
 	// Preprocess the spec to generate patches for gomod edits and other generators
 	// This must happen after build deps are installed so Go is available
 	if err := spec.Preprocess(client, sOpt, worker, opts...); err != nil {
-		return llb.Scratch()
+		return dalec.ErrorState(worker, err)
```
willie-yao, line 167: This change exposes the error of the state to the user as this is in an error check

spec.go
```diff
@@ -204,16 +204,237 @@ type SymlinkTarget struct {
...
+	Require []GomodRequire `yaml:"require,omitempty" json:"require,omitempty"`
+}
+
+// GomodReplace represents a go.mod replace directive.
+// It can be specified as a string "old:new" or as a struct with Old and New fields.
+// This allows users to substitute module dependencies before go mod download runs,
+// useful for pointing to local forks or alternate versions.
+type GomodReplace struct {
+	// Old is the module path to replace (can include @version)
+	Old string `yaml:"old" json:"old"`
```
cpuguy83: Nit: `Original`

spec.go
```diff
@@ -204,16 +204,237 @@ type SymlinkTarget struct {
...
+	// New is the replacement module path or local directory
+	New string `yaml:"new" json:"new"`
```
cpuguy83: Nit: `Update`

spec.go
```diff
@@ -204,16 +204,237 @@ type SymlinkTarget struct {
...
+}
+
+func (r GomodReplace) goModEditArg() (string, error) {
+	if r.Old == "" || r.New == "" {
+		return "", errors.Errorf("invalid gomod replace, old and new must be non-empty")
+	}
+	return r.Old + "=" + r.New, nil
+}
+
+func (r *GomodReplace) UnmarshalYAML(b []byte) error {
```
cpuguy83: In these we'll need to also set source maps, which would also require using the ast version of UnmarshalYAML (see other UnmarshalYAML's in the repo).

spec.go
```diff
@@ -204,16 +204,237 @@ type SymlinkTarget struct {
...
 
+// GomodEdits groups the go.mod manipulation directives that can be applied
+// before downloading module dependencies.
+type GomodEdits struct {
+	// Replace applies go.mod replace directives before downloading module dependencies.
+	// Each entry can be either a string "old:new" or a struct with Old and New fields.
+	Replace []GomodReplace `yaml:"replace,omitempty" json:"replace,omitempty"`
+	// Require applies go.mod require directives before downloading module dependencies.
+	// Each entry can be either a string "module:version" or a struct with Module and Version fields.
+	Require []GomodRequire `yaml:"require,omitempty" json:"require,omitempty"`
```
cpuguy83: So this injects totally new requires?

scripts/gomod-patch.sh
```diff
@@ -0,0 +1,59 @@
...
+  for godir in /usr/lib/go-*/bin; do
+    if [ -d "$godir" ] && [ -x "$godir/go" ]; then
+      export PATH="$godir:$PATH"
+      break
+    fi
+  done
+fi
+
+export GOMODCACHE="${TMP_GOMODCACHE}"
+: > "{{.PatchPath}}"
```
cpuguy83: Instead of using templates, can we pass args/envs in?

test/fixtures/simple-go-require.yml
```diff
@@ -0,0 +1,100 @@
...
+  msft-repo-envs: &msft-repo-envs
+    - build
+  jammy-repo: &jammy-repo
+    - keys: *msft-key
+      config:
+        microsoft-prod.list:
+          inline:
+            file:
+              contents: deb [arch=amd64,arm64,armhf signed-by=/usr/share/keyrings/msft.asc] https://packages.microsoft.com/ubuntu/22.04/prod jammy main
+      envs: *msft-repo-envs
```
cpuguy83: Let's make sure to remove this stuff before merge, I don't think we need to pull from msft repos.

cpuguy83 (changes requested):

spec.go
willie-yao: I think so. Is that a concern or should we be doing this a different way? @DannyBrito can maybe also clarify since he wrote the original struct.

cpuguy83: I guess the main thing is, do we have a use-case for this right now?

DannyBrito: sorry I missed this but tbh I don't really understand the question here; I reused initial implementation of the GomodEdits https://github.com/project-dalec/dalec/pull/769/changes#diff-f23a639898f3d395e15d7331eeab9d8d6217fe161c2851610171da61d9f92dd5R255.

cpuguy83: Basically, do we need to support adding new requires in this feature.

sozercan: we don't need to inject new requires since they'll get cleaned up by `go mod tidy`

we should only use `require` to update existing requires. `require: old:new`, not `require: new`

PR #881, opened 2025-12-01
Fix missing distro names in rpm targets

### LINKED ISSUE #875: [BUG] Printing targets for containers is not formatted properly.
### Expected Behavior

When running
```sh
docker buildx build --call targets --build-arg BUILDKIT_SYNTAX=ghcr.io/project-dalec/dalec/frontend:latest - <<< "null" | grep 'container ('
```

output is correctly formatted with `.`

### Actual Behavior

```console
$ docker buildx build --call targets --build-arg BUILDKIT_SYNTAX=ghcr.io/project-dalec/dalec/frontend:latest - <<< "null" | grep 'container ('
[+] Building 0.5s (3/3) FINISHED                                                                        docker-container:retina
 => resolve image config for docker-image://ghcr.io/project-dalec/dalec/frontend:latest                                    0.3s
 => CACHED docker-image://ghcr.io/project-dalec/dalec/frontend:latest@sha256:28aa75eb10152fb297f1c585ea2de75a2c62ceb4c904  0.0s
 => => resolve ghcr.io/project-dalec/dalec/frontend:latest@sha256:28aa75eb10152fb297f1c585ea2de75a2c62ceb4c904fe1483c4f8f  0.0s
 => [internal] load build definition from Dockerfile                                                                       0.0s
 => => transferring dockerfile: 42B                                                                                        0.0s
almalinux8/container (default)   Builds a container image for
almalinux9/container (default)   Builds a container image for
azlinux3/container (default)     Builds a container image for
mariner2/container (default)     Builds a container image for
rockylinux8/container (default)  Builds a container image for
rockylinux9/container (default)  Builds a container image for
windowscross/container (default) Builds binaries and installs them into a Windows base image
```

```go
	mux.Add("container", linux.HandleContainer(cfg), &targets.Target{
		Name:        "container",
		Description: "Builds a container image for " + cfg.FullName,
		Default:     true,
	})
```

It seems FullName appears empty.

### Steps To Reproduce

1. Run `docker buildx build --call targets --build-arg BUILDKIT_SYNTAX=ghcr.io/project-dalec/dalec/frontend:latest - <<< "null" | grep 'container ('`.

cpuguy83:
Fixes #875 

Also adds a command to generate target docs.
Ideally this would be underneath adjacent to the doc file itself, however because the website currently lives in another go module it complicates the `go generate ./...` case.

Meanwhile the go module for the website is purely to make it run in a dagger container so it can be executed without extra tooling on the machine.
We could probably remove this separate module and do something different in a separate change.

### DIFF
--- a/targets/linux/rpm/almalinux/v8.go
+++ b/targets/linux/rpm/almalinux/v8.go
@@ -32,4 +32,5 @@ var ConfigV8 = &distro.Config{
 	BasePackages:       basePackages(V8TargetKey),
 	RepoPlatformConfig: &defaultPlatformConfig,
 	InstallFunc:        distro.DnfInstall,
+	FullName:           v8FullName,
 }
--- a/targets/linux/rpm/almalinux/v9.go
+++ b/targets/linux/rpm/almalinux/v9.go
@@ -32,4 +32,5 @@ var ConfigV9 = &distro.Config{
 	BasePackages:       basePackages(V9TargetKey),
 	RepoPlatformConfig: &defaultPlatformConfig,
 	InstallFunc:        distro.DnfInstall,
+	FullName:           v9FullName,
 }
--- a/targets/linux/rpm/azlinux/azlinux3.go
+++ b/targets/linux/rpm/azlinux/azlinux3.go
@@ -29,4 +29,5 @@ var Azlinux3Config = &distro.Config{
 	InstallFunc:        distro.TdnfInstall,
 
 	SysextSupported: true,
+	FullName:        AzLinux3FullName,
 }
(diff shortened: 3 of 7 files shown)


cmd/gen-doc-targets/main.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		f, err := os.OpenFile(flag.Arg(0), os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to open output file: %v\n", err)
+			os.Exit(1)
+		}
+		defer f.Close()
+		out = f
+	}
```
DannyBrito, line 57: not a must but wouldn't you want to validate the flags right at the beginning before processing everything else?

cpuguy83: Yeah, I suppose so.
Artifact of how I iterated on this.
I think this is fine, especially given the use case.
It would be weird for someone to do something wrong here and even then its a relatively cheap oops since this doesn't need to roundtrip through buildkit or anything.

PR #904, opened 2025-12-16
frontend: Add json schema subrequest for dalec specs

### LINKED ISSUE #883: [REQ] Make json schema downloadable from frontend

In developing #850 it has become apparent that it would be nice if the json for the dalec frontend could be fetched from the frontend itself.

This would be a subrequest, similar to listing targets and other things.
This would make it so the vscode extension doesn't need to embed the json schema, which could even be the incorrect version of the schema depending on the syntax directive of a given spec file.

ashu8912:
Fixes #883 

Adds a new subrequest frontend.dalec.schema that returns the JSON schema for Dalec spec files. This enables tools like VS Code extensions to dynamically fetch the correct schema version based on the # syntax= directive in spec files.

To test
```
go run ./cmd/gen-jsonschema/ /tmp/test-schema.json
```

### DIFF
--- a/schema_embed.go
+++ b/schema_embed.go
@@ -1,0 +1,14 @@
+package dalec
+
+import _ "embed"
+
+// schemaJSON contains the embedded JSON schema for Dalec specs.
+// This is generated at build time by running `go generate`.
+//
+//go:embed docs/spec.schema.json
+var schemaJSON []byte
+
+// GetJSONSchema returns the embedded JSON schema for Dalec specs.
+func GetJSONSchema() []byte {
+	return schemaJSON
+}
--- a/frontend/mux.go
+++ b/frontend/mux.go
@@ -205,14 +205,10 @@ func handleDefaultPlatform() (*gwclient.Result, error) {
 }
 
 func handleJSONSchema() (*gwclient.Result, error) {
-	schemaJSON, err := dalec.GenerateJSONSchema()
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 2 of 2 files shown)


cpuguy83 (changes requested):
Instead of generating on the fly, I think we should embed the pre-generated file.

schema.go
```diff
@@ -0,0 +1,110 @@
...
+	"slices"
+	"strings"
+
+	"github.com/atombender/go-jsonschema/pkg/schemas"
+	"github.com/invopop/jsonschema"
+)
+
+// GenerateJSONSchema generates and returns the JSON schema for Dalec specs.
+// This schema can be used by editors and tools for validation and autocomplete.
+func GenerateJSONSchema() ([]byte, error) {
```
cpuguy83: I don't really see a need to export this, can we move this back into the cmd main package?

schema_embed.go
```diff
@@ -0,0 +1,9 @@
+package dalec
+
+import _ "embed"
+
+// SchemaJSON contains the embedded JSON schema for Dalec specs.
+// This is generated at build time by running `go generate`.
+//
+//go:embed docs/spec.schema.json
+var SchemaJSON []byte
```
cpuguy83: Let's keep this unexported and add a function to get it.

ashu8912: updated

schema.go
ashu8912: updated

cpuguy83 (approved):

PR #898, opened 2025-12-12
Move spec tests into the LLB graph

cpuguy83:
This moves Dalec spec/package tests into the normal LLB graph instead of running test containers manually through the BuildKit API after the main solve.

The main implementation details and rationale are in `tests: run package tests in LLB graph`.

This also includes small supporting fixes to protect shared test spec mutation and to build docs examples through the existing bake target.

PR #909, opened 2025-12-22
tests: always make sure we return an LLB state

cpuguy83:
Before this we are returning errors to buildkit when there are test errors.
Or for that matter, if there is an error during evaluation of the stuff running up to the tests.
This causes issues with the debug workflow ("docker buildx dap") since there is no LLB state to attach to, so it can't set breakpoints properly.

With this change even when we get a solve error when evaluating state, we return the bad state so that the debugger can attach to it. When a test fails, we add a new LLB exec op which prints the joined test errors and exits non-zero so that the build will fail.

There's also some minor changes to make it simpler to discard any changes to the filesystem from running tests, or in the case of non-container outputs return the original state, while keeping the dependency chain in tact.

This is a simpler alternative to #898.
#898 is hitting some race condition in buildkit (or at least that's what it looks like).
Additionally this can actually be backported to older release branches as the changes are minor.

### DIFF
--- a/frontend/test_runner.go
+++ b/frontend/test_runner.go
@@ -2,7 +2,6 @@ package frontend
 
 import (
 	"context"
-	"encoding/json"
 	stderrors "errors"
 	"fmt"
 	"io"
@@ -24,6 +23,13 @@ import (
 	"github.com/project-dalec/dalec/frontend/pkg/bkfs"
 	"github.com/project-dalec/dalec/internal/testrunner"
 )
+
+const (
+	testErrorsOutputFile = ".errors.txt"
+	testOutputPath       = "/tmp/dalec/internal/test/step/output"
+)
+
+var errorsOutputFullPath = filepath.Join(testOutputPath, testErrorsOutputFile)
 
 // RunTests runs the tests defined in the spec against the given the input state.
 // The result of this is either the provided `finalState` or a state that errors when executed with the errors produced by the tests.
@@ -75,39 +81,10 @@ func RunTests(ctx context.Context, client gwclient.Client, spec *dalec.Spec, fin
 
 		var group errGroup
 
-		const errorsOutputFile = ".errors.txt"
-		const outputPath = "/tmp/dalec/internal/test/step/output"
-		errorsOutputFullPath := filepath.Join(outputPath, errorsOutputFile)
-
 		for _, test := range tests {
-			base := in
-			for k, v := range test.Env {
-				base = base.AddEnv(k, v)
-			}
-
-			var opts []llb.RunOption
 			pg := llb.ProgressGroup(identity.NewID(), "Test: "+path.Join(target, test.Name), false)
-			opts = append(opts, pg)
 
-			for _, sm := range test.Mounts {
-				opts = append(opts, sm.ToRunOption(sOpt, pg))
-			}
-
-			result := base
-			result = result.File(llb.Mkdir(outputPath, 0o755, llb.WithParents(true)), pg)
-
-			for i, step := range test.Steps {
-				opts := append(opts, testrunner.WithTestStep(frontendSt, &step, i, errorsOutputFullPath))
-				opts = append(opts, step.GetSourceLocation(result))
-				result = result.Run(opts...).Root()
-			}
-
-			if len(test.Files) > 0 {
-				opts := append(opts, testrunner.WithFileChecks(frontendSt, test, errorsOutputFullPath))
-
-				opts = append(opts, llb.WithCustomNamef("Execute file checks for test: %s", test.Name))
-				result = result.Run(opts...).Root()
-			}
+			result := in.With(runTest(frontendSt, sOpt, test, pg))
 
 			group.Go(func() (retErr error) {
 				defer func() {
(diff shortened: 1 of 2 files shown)


frontend/test_runner.go
```diff
@@ -23,180 +25,218 @@ import (
...
-		}
-		if skip {
-			Warn(ctx, client, llb.Scratch(), "Tests skipped due to build-arg DALEC_SKIP_TESTS="+skipVar)
-			return nil
+// RunTests runs the tests defined in the spec against the given the input state.
+// The result of this is either the provided `finalState` or a state that errors when executed with the errors produced by the tests.
+// The input state should be a runnable container with all dependencies already installed.
+func RunTests(ctx context.Context, client gwclient.Client, spec *dalec.Spec, finalState llb.State, target string, platform *ocispecs.Platform) llb.StateOption {
+	return func(in llb.State) llb.State {
+		if skipVar := client.BuildOpts().Opts["build-arg:"+"DALEC_SKIP_TESTS"]; skipVar != "" {
```
MorrisLaw: nit: we could just combing the string literals together

frontend/test_runner.go
```diff
@@ -23,180 +25,218 @@ import (
...
+			skip, err := strconv.ParseBool(skipVar)
+			if err != nil {
+				err = errors.Wrapf(err, "could not parse build-arg %s", "DALEC_SKIP_TESTS")
```
MorrisLaw: nit: no need to use a format specifier since it's just a hardcoded string; we could combine it into 1 string

cpuguy83: Changed this up a bit.

frontend/test_runner.go
cpuguy83: Changed this up a bit.

targets/linux/rpm/distro/pkg.go
```diff
@@ -68,29 +67,14 @@ func (c *Config) BuildPkg(ctx context.Context, client gwclient.Client, sOpt dale
 	return frontend.MaybeSign(ctx, client, st, spec, targetKey, sOpt, opts...)
 }
 
-// runTests runs the package tests
+// RunTests runs the package tests
 // The returned reference is the solved container state
-func (cfg *Config) RunTests(ctx context.Context, client gwclient.Client, spec *dalec.Spec, sOpt dalec.SourceOpts, ctr llb.State, targetKey string, opts ...llb.ConstraintsOpt) (gwclient.Reference, error) {
-	def, err := ctr.Marshal(ctx, opts...)
```
invidian, line 74: Nice seeing that duplicated bits refactored!

test/linux_target_test.go
```diff
@@ -3089,44 +3093,69 @@ func testLinuxPackageTestsFail(ctx context.Context, t *testing.T, cfg testLinuxC
...
-				assert.ErrorContains(t, err, v)
+		doTest := func(target string) func(t *testing.T) {
+			return func(t *testing.T) {
+				t.Parallel()
 
-				v = (&dalec.CheckOutputError{Path: "/", Kind: dalec.CheckFilePermissionsKind, Expected: "-rw-r--r--", Actual: "-rwxr-xr-x"}).Error()
-				assert.ErrorContains(t, err, v)
+				dir := t.TempDir()
+				f, err := os.OpenFile(filepath.Join(dir, "out.txt"), os.O_CREATE|os.O_RDWR, 0o600)
+				assert.NilError(t, err)
```
invidian, line 3115: I'm not sure that's the good place for discussing this, but generally speaking, `assert` does not interrupt the execution of the function which often leads to panics and false-negatives (e.g. when f is nil), so maybe it would be better to use `require` instead, which gives clearer failure handling?

test/linux_target_test.go
```diff
@@ -3089,44 +3093,69 @@ func testLinuxPackageTestsFail(ctx context.Context, t *testing.T, cfg testLinuxC
...
+					}
+				}
 
-				v = (&dalec.CheckOutputError{Path: "/some_symlink2", Kind: dalec.CheckFileLinkTargetPathKind, Expected: "/not-a-real-path2", Actual: "/usr/bin/a-thing-for-symlinking"}).Error()
-				assert.ErrorContains(t, err, v)
+				ctx := startTestSpan(ctx, t)
+				testEnv.RunTest(ctx, t, func(ctx context.Context, client gwclient.Client) {
+					sr := newSolveRequest(withSpec(ctx, t, spec), withBuildTarget(target))
+					_, err := client.Solve(ctx, sr)
+					assert.Assert(t, err != nil)
```
invidian, line 3133: `require.NilError`?

test/linux_target_test.go
```diff
@@ -3089,44 +3093,69 @@ func testLinuxPackageTestsFail(ctx context.Context, t *testing.T, cfg testLinuxC
...
+			(&dalec.CheckOutputError{Path: "/some_symlink3", Kind: dalec.CheckFileNotExistsKind, Expected: "exists=true", Actual: "exists=false"}).Error(),
+			(&dalec.CheckOutputError{Path: "/some_symlink4", Kind: dalec.CheckFileLinkTargetPathKind, Expected: "/incorrect-target", Actual: "/not-a-real-path4"}).Error(),
+			"step did not complete successfully: exit code: 42",
+			"stdout not hello",
+			"stderr not hello",
+		}
 
-				v := (&dalec.CheckOutputError{Path: "/non-existing-file", Kind: dalec.CheckFileNotExistsKind, Expected: "exists=true", Actual: "exists=false"}).Error()
-				assert.ErrorContains(t, err, v)
+		doTest := func(target string) func(t *testing.T) {
```
invidian: Maybe?
```suggestion
		testsForTarget := func(target string) func(t *testing.T) {
```

frontend/test_runner.go
```diff
@@ -23,180 +25,222 @@ import (
...
-		if err != nil {
-			return errors.Wrapf(err, "could not parse build-arg %s", "DALEC_SKIP_TESTS")
-		}
-		if skip {
-			Warn(ctx, client, llb.Scratch(), "Tests skipped due to build-arg DALEC_SKIP_TESTS="+skipVar)
-			return nil
+// RunTests runs the tests defined in the spec against the given the input state.
+// The result of this is either the provided `finalState` or a state that errors when executed with the errors produced by the tests.
+// The input state should be a runnable container with all dependencies already installed.
+func RunTests(ctx context.Context, client gwclient.Client, spec *dalec.Spec, finalState llb.State, target string, platform *ocispecs.Platform) llb.StateOption {
```
invidian, line 37: If implementing `llb.StateOption`, do we actually need to take `finalState llb.State` as argument? Wouldn't `in llb.State` always be the final state?

cmd/frontend/main.go
```diff
@@ -55,6 +55,9 @@ func main() {
 	case testrunner.CheckFilesCmdName:
 		args := flag.Args()[2:]
 		testrunner.CheckFilesCmd(args)
+	case frontend.TestErrorCmdName:
```
invidian, line 58: Why do we need a separate frontend command to handle printing errors? Couldn't this just run after we are finished running tests?

test/linux_target_test.go
cpuguy83: `assert` here is not testify, its `gotest.tools/v3/asssert`.
`assert.NilError` does halt the test

test/linux_target_test.go
cpuguy83: This is asserting that we have an error.

frontend/test_runner.go
cpuguy83: `finalState` here is because the input state is not always going to be the final state, e.g. for packaging targets such as jammy/deb and sysext targets.

cmd/frontend/main.go
cpuguy83: This isn't just printing the errors, its exiting non-zero.
The reason we have this is so we can return an LLB state to buildkit that will error out and fail the build rather than returning an error to buildkit which messes up the dap debugger.

I actually have an alternative to this where each test assertion is actually run as an exec (`state.Run(...)`) but that seems to be triggering a bug in buildkit.
So I've settled for this case where we keep running the tests like we used to which requires un-lazying the build but still give buildkit a valid LLB state that produces an error.

test/linux_target_test.go
invidian: Ah, that's confusing, TIL. Maybe we could remove testify assert completely to avoid this confusion. I'll have a look into that.

cpuguy83: Indeed, I opened an issue yesterday to track this, except I suggested switching to testify since that is a more well known library (of course either way seems fine to me).

frontend/test_runner.go
invidian: I think I still don't understand. Could you elaborate to help me graps how this works? So the input state could be worker with test dependencies, while final state would be where deb/sysext are? Do I get this right?

frontend/test_runner.go
```diff
@@ -23,180 +25,222 @@ import (
...
 				}
-			}
 
-			return stderrors.Join(errs...)
-		})
+				return stderrors.Join(errs...)
+			})
+		}
+
+		err = group.Wait()
```
invidian: I don't think it's the task for this PR, but I feel code in RunTests would benefit from the extensive test suite for all different behaviours, since there is really a lot of code and as a newcomer it's hard to figure out what it all does.

Or maybe alternatively some refactoring to better organize levels of abstraction in this function could help make it easier to understand and drive down the complexity would be sufficient.

frontend/test_runner.go
cpuguy83: The input state must always be a state that is ready to have test assertions executed.
deb/rpm/sysext are not that but we still want to run tests when building those targets (actually even the container state may have extra stuff due to test dependencies being installed).

In those cases we make the input state the container (ie package installed) + test dependencies.
The final state, which the test runner outputs, needs to be what we plan to output OR we need to handle that at the call site, which could also be fair (maybe a separate state option called Reset or something).

frontend/test_runner.go
cpuguy83: I think we have a large amount of coverage for the test runner within the integration test suite as it is.
Its possible an audit may come up with some case that's not tested but we do cover both positive and negative case in there to validate the test runner is correct and we also rely on the test runner for some of the rest of the test suite.

frontend/test_runner.go
cpuguy83: That said, I agree the code itself is complex.
I factored out a bunch of this in #898 but ran into some issues that need to be dug into further and may require fixes in buildkit (which would also delay actually being able to refactor since the test runner wouldn't be usable on older buildkits).

cpuguy83: I factored out a few things to split this loop up in to smaller chunks, hope this helps (all in the 2nd commit).

invidian: Thanks for the explanation. It makes sense now :)

PR #916, opened 2026-01-06
Remove use of dagger

cpuguy83:
Dagger was convenient at the time for getting the site running without needing a bunch of things installed on the machine, but is overkill for the usage and even problematic because:

1. Introduces a whole different go module (b/c dagger dependencies)
2. Doesn't work in some cases (namely I have issues with dagger+nftables)
3. Is still an extra thing in the tooling that needs to be installed even if its handled automatically.

Now that we have better stuff for interacting with buildkit, and understand more of how to do that, we can actually serve the whole website  directly out of a buildkit build without any extra tooling than what is required to work on the rest of the project.

This uses buildkit to build the static site content and then serves it directly out of the buildkit cache.


website/docs/architecture.md
```diff
@@ -0,0 +1,141 @@
+# Dalec Architecture
```
cpuguy83, line 1: This is just moved from the repo root (and root has a symlink) so that the builder has access to it.

website/main.go
```diff
@@ -5,104 +5,268 @@ import (
...
+			).AddMount("/website", in)
+
+		out := llb.Scratch().File(llb.Copy(generated, "build", "/", dalec.WithDirContentsOnly()), llb.WithCustomName("Get static content"))
+
+		return out
+	}
+}
+
+func website(ctx context.Context, bkc *client.Client, addr string) error {
+	defer bkc.Close()
```
MorrisLaw: Just need to address these (and the other 2 missed err checks) to make the linter happy

cpuguy83: Should be good now.

PR #918, opened 2026-01-07
test/linux_target_test.go: refactor container tests into multiple suites

invidian:
This way we get more targeted tests and assertions, test scenarios
complexity goes down and it makes it easier to extend for additional
tests.

It also modifies some fields in Spec tags to omit them while serializing when they are empty, so we can use serialization for merging multiple specs, which I think is handy and I use it in tests to build up the spec from the basic one.

### DIFF
--- a/test/linux_target_test.go
+++ b/test/linux_target_test.go
@@ -197,12 +197,74 @@ func testLinuxDistro(ctx context.Context, t *testing.T, testConfig testLinuxConf
 		testTargetArtifactsTakePrecedence(ctx, t, testConfig.Target)
 	})
 
-	t.Run("container", func(t *testing.T) {
+	t.Run("build_steps", func(t *testing.T) {
 		t.Parallel()
-		ctx := startTestSpan(baseCtx, t)
 
-		const src2Patch3File = "patch3"
-		src2Patch3Content := []byte(`
+		t.Run("multiline_command_works_with_env_vars", func(t *testing.T) {
+			t.Parallel()
+
+			ctx := startTestSpan(baseCtx, t)
+
+			spec := testLinuxSpec(t, dalec.Spec{
+				Build: dalec.ArtifactBuild{
+					Steps: []dalec.BuildStep{
+						{
+							// Test that a multiline command works with env vars
+							Env: map[string]string{
+								"FOO": "foo",
+								"BAR": "bar",
+							},
+							Command: `
+echo "${FOO}_0" > foo0.txt
+echo "${FOO}_1" > foo1.txt
+echo "$BAR" > bar.txt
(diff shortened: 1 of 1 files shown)


test/linux_target_test.go
```diff
@@ -4230,3 +4454,31 @@ echo "This is a third test binary"
...
+		Version:     "0.0.1",
+		Revision:    "1",
+		License:     "MIT",
+		Website:     "https://github.com/project-dalec/dalec",
+		Vendor:      "Dalec",
+		Packager:    "Dalec",
+		Description: "Testing container target",
+
+		Dependencies: &dalec.PackageDependencies{
+			Runtime: map[string]dalec.PackageConstraints{
```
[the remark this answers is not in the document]
invidian: It seems all tests are passing only with `coreutils`, so explicitly adding bash seems unnecessary, no-op.

test/linux_target_test.go
```diff
@@ -306,155 +352,214 @@ index 0000000..5260cb1
...
+							{
+								Command: "/usr/bin/file2",
+								Stdout:  dalec.CheckOutput{Equals: "Added a new file\n"},
+								Stderr:  dalec.CheckOutput{Empty: true},
+							},
+						},
 					},
 					{
-						Command: "grep 'Added another new file' ./src2/file3",
+						Name: "Post-install symlinks should be created and have correct ownership",
```
invidian, line 465: Those tests are probably unnecessary here since they don't test source patches.

spec.go
```diff
@@ -17,17 +17,17 @@ import (
 // Spec is the specification for a package build.
 type Spec struct {
 	// Name is the name of the package.
-	Name string `yaml:"name" json:"name" jsonschema:"required"`
+	Name string `yaml:"name" json:"name,omitempty" jsonschema:"required"`
```
cpuguy83, line 20: I don't think these changes should be omitempty.
What was the reasoning for the change?

cpuguy83 (changes requested):

spec.go
invidian: So that those fields do not need explicit copy when you serialize one over another. Without that change, those empty fields override existing values from the original struct.

This is not a strictly necessary change, just a convenience. If it's undesired, I'll just do an explicit copy.

cpuguy83: The main thing here is the jsonschema generator is using this to determine if the field is nullable, hence the schema changes.

If there's another way to annotate the field to keep the jsonschema correct, then that's fine.

invidian: I added `jsonschema:"required"` to the changed fields, this way the schema does not change.

PR #922, opened 2026-01-07
ci: Prebuild frontend and worker

cpuguy83:
Since tests run in parallel we can end up firing off multiple builds where we have not yet cached the frontend or worker image and end up doing more work than we really should.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,6 +241,25 @@ jobs:
           echo
 
           echo "================ CLEANUP COMPLETE ================"
+
+      - name: Pre-build base images
+        run: |
+          set -eu
+
+          docker buildx bake frontend
+          if [ "${TEST_SUITE}" = "other" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -241,6 +241,25 @@ jobs:
...
+          if [ "${TEST_SUITE}" = "other" ]; then
+            exit 0
+          fi
+
+          # downcase the test suite to get the worker target
+          worker="${TEST_SUITE,,}"
+          if [ "${worker}" = "windows" ]; then
+            worker="windowscross"
+          fi
+          export WORKER_TARGET=${worker}/worker
```
DannyBrito, line 259: `WORKER_TARGET` how or where is this used? does this get picked by bake worker?

cpuguy83: Yes, it's declared in docker-bake.hcl

Issue #841, opened 2025-11-04
[REQ] Speed up package installation for non-native builds


cpuguy83:
### What kind of request is this?

Improvement of existing experience

### Summary
When dalec builds for a non-native architecture (e.g., arm64 on amd64), package installation runs under qemu emulation. This is slow and accounts for most of the total build time.

### Proposal
Build worker images natively using the host’s package manager, targeting the foreign architecture:

For Debian/Ubuntu: use --arch or --foreign-architecture flags to resolve and install build dependencies for the target arch natively.

For RPM-based systems: evaluate whether --forcearch or similar can be safely used, or just fall back to emulated install.

The result is a worker image for the target arch that avoids qemu during dependency installation.

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

PR #923, opened 2026-01-08
deb: speed up cross-arch worker builds by installing deps into mounted target rootfs

kartikjoshi21:
Cross builds now separate build-platform execution from target-rootfs mutation.
We run apt on the BuildKit executor platform, download deps into per-arch caches,
and install/configure inside a chroot of the mounted target rootfs.

Includes:
- BuildKit native platform detection + platform normalization helpers.
- Explicit build vs target base resolution.
- New AptInstallIntoRoot helper with dpkg wrappers and arch-scoped cache dirs.
- Cross-safe cache scoping and SysextWorker parity.
- Native builds are unchanged; cross builds become faster and more reliable.

Cross builds now run APT resolution + downloads natively on the BuildKit executor platform (using target-rootfs dpkg state and per-arch caches), and only use QEMU/binfmt for the chrooted install/configure step where maintainer scripts must execute in the target rootfs.
Fixes: #841 
Signed-off-by: Kartik Joshi <karikjoshi21@gmail.com>

### DIFF
--- a/targets/linux/deb/distro/install.go
+++ b/targets/linux/deb/distro/install.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 
 	"github.com/moby/buildkit/client/llb"
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/project-dalec/dalec"
 	"github.com/project-dalec/dalec/packaging/linux/deb"
 )
@@ -21,10 +22,11 @@ func AptInstall(packages []string, opts ...llb.ConstraintsOpt) llb.RunOption {
 		const installScript = `#!/usr/bin/env sh
 set -ex
 
+
 # Make sure any cached data from local repos is purged since this should not
 # be shared between builds.
 rm -f /var/lib/apt/lists/_*
-apt autoclean -y
+apt autoclean -y 
 
 # Remove any previously failed attempts to get repo data
 rm -rf /var/lib/apt/lists/partial/*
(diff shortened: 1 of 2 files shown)


.github/workflows/worker-images.yml
```diff
@@ -108,8 +108,7 @@ jobs:
           echo "EXTRA_FLAGS=${EXTRA_FLAGS}" >> $GITHUB_ENV
 
           # Format repository name (lowercase) for registry mirror
-          FORMATTED_REPO=$(echo "${{ github.repository }}" | tr '[:upper:]' '[:lower:]')
-          docker run -d --net=host ghcr.io/${FORMATTED_REPO}/mirror/dockerhub/library/registry:latest
+          docker run -d --net=host --restart=always -p 5000:5000 registry:2
```
cpuguy83: I assume this was to get it working in your fork?

.github/workflows/worker-images.yml
```diff
@@ -124,7 +123,7 @@ jobs:
           # Inline the cache with the image so others can use it as a cache-from for their own builds.
           EXTRA_FLAGS="--set worker.cache-to=type=inline,mode=max --set worker.tags=${ref} --push"
 
-          latest="$(gh release view --json tagName -q .tagName)"
+          latest="$(gh release view --json tagName -q .tagName 2>/dev/null || echo '')"
```
cpuguy83: I think `|| true` would work here.

targets/linux/deb/distro/install.go
```diff
@@ -18,9 +19,10 @@ import (
 // The constraints are applied after any constraint set on the [llb.ExecInfo]
 func AptInstall(packages []string, opts ...llb.ConstraintsOpt) llb.RunOption {
 	return dalec.RunOptFunc(func(ei *llb.ExecInfo) {
-		const installScript = `#!/usr/bin/env sh
+		const installScript = `#!/bin/sh
 set -ex
 
+
```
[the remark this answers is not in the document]
cpuguy83: Any particular reason that this is changed?

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +47,272 @@ apt install -y "$@"
...
+ -o Dir::Etc::sourcelist=${ROOTFS}/etc/apt/sources.list
+"
+        fi
+fi
+
+apt-get ${APT_OPTS} update
+
+DEBIAN_FRONTEND=noninteractive apt-get ${APT_OPTS} --download-only install -y "$@"
+
+ls -1 "${ARCHIVE_DIR}"/*_"${ARCH}".deb "${ARCHIVE_DIR}"/*_all.deb >/dev/null 2>&1 || {
```
cpuguy83: Why not `*.deb`?
If we keep this as is, can you add a comment explaining why?

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +47,272 @@ apt install -y "$@"
...
+#!/bin/sh
+exit 101
+EOF
+chmod +x "${POLICY}"
+cleanup_policy() { rm -f "${POLICY}" 2>/dev/null || true; }
+
+# Make chroot environment usable for maintainer scripts.
+mkdir -p "${ROOTFS}/proc" "${ROOTFS}/sys" "${ROOTFS}/dev" "${ROOTFS}/dev/pts"
+mount_proc_sys_dev() {
+        mount -t proc proc "${ROOTFS}/proc" 2>/dev/null || true
```
cpuguy83: I don't think any of these mount calls will actually work since these mounts require CAP_SYS_ADMIN.

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +47,272 @@ apt install -y "$@"
...
+fi
+
+# Phase 2: install/configure inside the target rootfs.
+ROOT_ARCHIVE_DIR="${ROOTFS}/var/cache/apt/archives"
+mkdir -p "${ROOT_ARCHIVE_DIR}/partial"
+# Prefer bind-mounting archives for speed and to avoid inflating the target rootfs.
+if mount --bind "${ARCHIVE_DIR}" "${ROOT_ARCHIVE_DIR}" 2>/dev/null; then
+        BOUND_ARCHIVES=1
+else
+        cp -a "${ARCHIVE_DIR}/." "${ROOT_ARCHIVE_DIR}/" || true
```
cpuguy83: This would ideally be handled by `llb.AddMount` instead of trying to do this in a script (which the mount call will always fail anyway).

.github/workflows/worker-images.yml
kartikjoshi21: Yes, removed it , wanted to run things on my fork

targets/linux/deb/distro/install.go
kartikjoshi21: added comments , Thanks

targets/linux/deb/distro/install.go
kartikjoshi21: These mounts are not required for correctness. dpkg --configure and maintainer scripts are designed to run in minimal chroot/container environments, and the only strict requirement for cross-arch operation here is binfmt/qemu availability. and yes it actually doesnt work without CAP SYS. Removed

targets/linux/deb/distro/install.go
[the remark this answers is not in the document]
kartikjoshi21: it was to make it consistent , not intended.

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +47,272 @@ apt install -y "$@"
...
+  --root="${ROOTFS}" \
+  --admindir="${ADMINDIR}" \
+  --force-architecture \
+  "$@"
+
+EOF
+chmod +x /tmp/dalec/dpkg
+
+cat > /tmp/dalec/dpkg-query <<'EOF'
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
kartikjoshi21: Keeping `#!/usr/bin/env sh` everywhere

.github/workflows/worker-images.yml
kartikjoshi21: this was for local testing.

targets/linux/deb/distro/install.go
kartikjoshi21: Yes it was Updated

.github/workflows/worker-images.yml
```diff
@@ -108,8 +108,8 @@ jobs:
           echo "EXTRA_FLAGS=${EXTRA_FLAGS}" >> $GITHUB_ENV
 
           # Format repository name (lowercase) for registry mirror
-          FORMATTED_REPO=$(echo "${{ github.repository }}" | tr '[:upper:]' '[:lower:]')
-          docker run -d --net=host ghcr.io/${FORMATTED_REPO}/mirror/dockerhub/library/registry:latest
+          # Temporary change
+          docker run -d --net=host --restart=always -p 5000:5000 registry:2
```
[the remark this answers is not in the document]
kartikjoshi21: will be removed later on

targets/linux/deb/distro/install.go
```diff
@@ -18,19 +19,40 @@ import (
...
+	BOOT_OPTS="-o Acquire::https::Verify-Peer=false -o Acquire::https::Verify-Host=false -o Acquire::Retries=5 -o Acquire::http::Timeout=30 -o Acquire::https::Timeout=30"
+	set +e
+	apt-get ${BOOT_OPTS} update || apt-get ${BOOT_OPTS} -o Acquire::ForceIPv4=true update
+	DEBIAN_FRONTEND=noninteractive apt-get ${BOOT_OPTS} install -y --no-install-recommends ca-certificates
+	set -e
+	command -v update-ca-certificates >/dev/null 2>&1 && update-ca-certificates || true
+}
+
+bootstrap_ca_certs_if_needed || true
+apt-get update -o Acquire::Retries=5 -o Acquire::http::Timeout=30 -o Acquire::https::Timeout=30 \
```
cpuguy83: These settings I don't think we should be here.
I get you are trying to work around archive.ubuntu.com timing out, though.
I opened https://github.com/project-dalec/dalec/pull/946 to solve this in CI at least.

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +67,207 @@ apt install -y "$@"
...
+ -o Dir::State::status=${ROOTFS}/var/lib/dpkg/status
+ -o Dir::Cache=/var/cache/apt
+ -o Dir::Cache::archives=${ARCHIVE_DIR}
+ -o APT::Architecture=${ARCH}
+ -o APT::Architectures::=${ARCH}
+ -o APT::Architectures::=all
+ -o Dir::Bin::dpkg=/tmp/dalec/dpkg
+ -o Dir::Bin::dpkg-query=/tmp/dalec/dpkg-query
+ -o Acquire::Languages=none
+ -o Acquire::Retries=5
```
cpuguy83: Same comment re: retries/timeout/etc.

targets/linux/deb/distro/install.go
kartikjoshi21: Yes, i just wanted everything to pass , didnt see already open issue, Thanks

cpuguy83:
Building runc from the docker-bake.hcl, good improvement:

## PR
amd64->arm64 4m25s
amd64->amd64 1m27s

## main
amd64->arm64 5m5s
amd64->amd64 1m27s

targets/linux/deb/distro/install.go
```diff
@@ -21,10 +22,11 @@ func AptInstall(packages []string, opts ...llb.ConstraintsOpt) llb.RunOption {
 		const installScript = `#!/usr/bin/env sh
 set -ex
 
+
 # Make sure any cached data from local repos is purged since this should not
 # be shared between builds.
 rm -f /var/lib/apt/lists/_*
-apt autoclean -y
+apt autoclean -y
```
cpuguy83, line 29: Nit, extra whitespace here (not worth a ci run,  though).

cpuguy83 (approved):

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +47,187 @@ apt install -y "$@"
 	})
 }
 
+// AptInstallIntoRoot installs packages into a mounted root filesystem (rootfsPath)
+// while running apt/dpkg from the *current* container environment (i.e. build platform).
+// Notes:
+//   - This relies on the target rootfs containing valid /etc/apt and /var/lib/dpkg state.
+//   - Maintainer scripts execute inside the target rootfs. If they invoke target-arch binaries,
```
invidian, line 54: What are maintainer scripts?

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +47,187 @@ apt install -y "$@"
...
+    log "target rootfs dpkg arch (${native_arch}) != requested (${ARCH})"
+    exit 4
+  fi
+fi
+
+SOURCELIST=""
+SOURCEPARTS=""
+[ -f "${ROOTFS}/etc/apt/sources.list" ] && SOURCELIST="${ROOTFS}/etc/apt/sources.list"
+[ -d "${ROOTFS}/etc/apt/sources.list.d" ] && SOURCEPARTS="${ROOTFS}/etc/apt/sources.list.d"
+if [ -z "${SOURCELIST}" ] && [ -z "${SOURCEPARTS}" ]; then
```
invidian, line 95: nit: That's not really comprehensive. `sources.list.d` could still be empty or has empty files.

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +47,187 @@ apt install -y "$@"
...
+  "$@"
+EOF
+chmod +x /tmp/dalec/dpkg-query
+
+# Host-side arch-scoped APT caches (speed + avoid collisions)
+ARCHIVE_DIR="/var/cache/apt/archives-${ARCH}"
+LISTS_DIR="/var/cache/apt/lists-${ARCH}"
+mkdir -p "${ARCHIVE_DIR}/partial" "${LISTS_DIR}/partial"
+rm -rf "${LISTS_DIR}/"* 2>/dev/null || true
+mkdir -p "${LISTS_DIR}/partial"
```
invidian, line 155: nit: `mkdir -p "${LISTS_DIR}/partial"` is called twice here? Probably `rm -rf` first, then `mkdir` would yield less code.

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +47,187 @@ apt install -y "$@"
...
+
+APT_OPTS="
```
invidian, line 157: nit: custom apt config file could be used for better readability, since the commands will be really long with this.

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +47,187 @@ apt install -y "$@"
...
+trap 'rm -f "${POLICY}" 2>/dev/null || true' EXIT
+
+chroot_configure() { chroot "${ROOTFS}" /usr/bin/dpkg --configure -a; }
+
+export DEBIAN_FRONTEND=noninteractive
+export DEBCONF_NONINTERACTIVE_SEEN=true
+export DEBIAN_PRIORITY=critical
+export NEEDRESTART_MODE=a
+
+# Phase 2a: UNPACK (native, outside chroot) into mounted ROOTFS using dpkg --root.
```
invidian, line 204: What are other phases?

targets/linux/deb/distro/install.go
```diff
@@ -45,6 +47,187 @@ apt install -y "$@"
...
+# This is the fast path: avoids running unpack under QEMU.
+for deb in "${ARCHIVE_DIR}"/*_"${ARCH}".deb "${ARCHIVE_DIR}"/*_all.deb; do
+	[ -f "${deb}" ] || continue
+	/tmp/dalec/dpkg --unpack --force-depends --no-triggers "$deb"
+done
+
+# Phase 2b: configure inside target rootfs.
+chroot_configure
```
invidian, line 212: chroot_configure is called only once, does it need a separate function? This seems to be making flow more confusing.

PR #925, opened 2026-01-14
Add missing constraints propagation and verify propagation in tests

invidian:
To make sure we are consistent in propagating constraints to operations
    which supports it, for example for annotating with progress groups or
    invalidating cache, this commit adds tests which will verify that all
    LLB operations from states which dalec generates includes a progress
    group at least, which is a good proxy for other constraints as well.

    To make tests pass, this commit also adds missing constraints
    propagation to various places where it was missing.

### DIFF
--- a/test/handlers_test.go
+++ b/test/handlers_test.go
@@ -55,7 +55,8 @@ func TestHandlerTargetForwarding(t *testing.T) {
 							Image: phonyRef,
 						},
 					},
-				}}
+				},
+			}
 
 			// Make sure phony is in the list of targets since it should be registered in the forwarded frontend.
 			ls = listTargets(ctx, t, gwc, spec)
@@ -75,7 +76,8 @@ func TestHandlerTargetForwarding(t *testing.T) {
 							Image: phonyRef,
 						},
 					},
-				}}
+				},
+			}
 
 			sr := newSolveRequest(withSpec(ctx, t, spec), withBuildTarget("phony/check"))
 			res := solveT(ctx, t, gwc, sr)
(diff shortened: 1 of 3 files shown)


sourcemap.go
```diff
@@ -48,7 +48,8 @@ func (sm *sourceMap) getLocation(st *llb.State) llb.ConstraintsOpt {
 	if sm == nil {
 		return ConstraintsOptFunc(func(*llb.Constraints) {})
 	}
-	sourceMap := llb.NewSourceMap(st, sm.filename, sm.language, sm.data)
+	// state passed here is wrong.
+	sourceMap := llb.NewSourceMap(nil, sm.filename, sm.language, sm.data)
```
invidian: This is here so we verify the constraints propagation also for scenarios which generate bad input op, but this needs to be reverted with an exception for those scenarios or addressed.

frontend/debug/handle_cargohome.go
```diff
@@ -25,15 +25,17 @@ func Cargohome(ctx context.Context, client gwclient.Client) (*gwclient.Result, e
 			return nil, nil, err
 		}
 
+		pg := dalec.ProgressGroup("cargohome-deps")
```
invidian, line 28: Most of new progress group names are provisional.

sourcemap.go
cpuguy83: Not sure I'm following this change.

invidian: See #930 for context.

cpuguy83: Ok, the issue is caused by this:

https://github.com/project-dalec/dalec/blob/ea157ea735b763548cbe45fd05aa9d8cdff42ccf/generator_cargohome.go#L79

We can't use `llb.Scratch()` as an input to the source map.
We can work around this by setting `st` to nil if `st.Output() == nil`

invidian: Thanks :) I opened #939 with a fix.

PR #931, opened 2026-01-20
rpm: speed up cross-arch worker builds by installing deps into mounted target rootfs

kartikjoshi21:
Enable faster and more reliable cross-arch RPM worker builds by installing build dependencies into a mounted target rootfs while running the package manager on the native BuildKit executor platform (avoids slow/full emulation and “exec format error” during install).

**Changes**
- Detect native executor platform from BuildKit worker opts and normalize platform tuples.
- For build != target:
      1. resolve target base image as target platform
      2. run install on build platform and install into mounted target rootfs via --installroot
      3. return the mutated rootfs as the target-platform worker
- Add optional --forcearch support (applied when supported by dnf/tdnf).
- Enable CacheAddPlatform for Rocky 8/9, Azure Linux 3, and Mariner 2 to avoid cross-arch cache/repo metadata collisions.

### DIFF
--- a/targets/linux/rpm/distro/distro.go
+++ b/targets/linux/rpm/distro/distro.go
@@ -37,11 +37,10 @@ type Config struct {
 	CacheName string
 	// Whether to namespace the cache key by platform
 	// Not all distros need this, hence why it is configurable.
-	// The cache key is only added when the build platform and target platform differ.
 	CacheAddPlatform bool
 
-	// e.g. /var/cache/tdnf or /var/cache/dnf
-	CacheDir string
+	// Cache directories to mount (e.g. ["/var/cache/dnf"] or ["/var/cache/tdnf", "/var/cache/dnf"])
+	CacheDir []string
 
 	// erofs-utils 1.7+ is required for tar support.
 	SysextSupported bool
@@ -61,7 +60,27 @@ func (cfg *Config) PackageCacheMount(root string) llb.RunOption {
 			}
 			cacheKey += "-" + platforms.Format(*p)
 		}
-		llb.AddMount(filepath.Join(root, cfg.CacheDir), llb.Scratch(), llb.AsPersistentCacheDir(cacheKey, llb.CacheMountLocked)).SetRunOption(ei)
+
+		if len(cfg.CacheDir) == 0 {
+			return
+		}
+
+		// Mount each cache dir. If there are multiple, suffix key with the dir base.
+		for _, d := range cfg.CacheDir {
+			if d == "" {
+				continue
+			}
+			k := cacheKey
+			if len(cfg.CacheDir) > 1 {
+				k = cacheKey + "-" + filepath.Base(d)
+			}
+			llb.AddMount(
+				joinUnderRoot(root, d),
+				llb.Scratch(),
+				llb.AsPersistentCacheDir(k, llb.CacheMountLocked),
+			).SetRunOption(ei)
+		}
+
 	})
 }
 
--- a/targets/linux/rpm/distro/worker.go
+++ b/targets/linux/rpm/distro/worker.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 
+	"github.com/containerd/platforms"
 	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/client/llb/sourceresolver"
 	gwclient "github.com/moby/buildkit/frontend/gateway/client"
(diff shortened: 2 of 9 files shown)


targets/linux/rpm/distro/worker.go
```diff
@@ -57,7 +65,54 @@ func (cfg *Config) HandleWorker(ctx context.Context, client gwclient.Client) (*g
 	})
 }
 
+// nativeExecutorPlatform returns the BuildKit worker's native (executor) platform.
+// Do NOT use platforms.DefaultSpec() for this: the frontend image may be running as the
+// TARGET platform (multi-arch frontend), which would make DefaultSpec() lie.
```
cpuguy83: I don't think this should ever be the case?
Did you run into this?

targets/linux/rpm/distro/worker.go
```diff
@@ -17,13 +18,20 @@ import (
 // TODO: can this be refactored to share code?
 func (cfg *Config) HandleWorker(ctx context.Context, client gwclient.Client) (*gwclient.Result, error) {
 	return frontend.BuildWithPlatform(ctx, client, func(ctx context.Context, client gwclient.Client, platform *ocispecs.Platform, spec *dalec.Spec, targetKey string) (gwclient.Reference, *dalec.DockerImageSpec, error) {
-		sOpt, err := frontend.SourceOptFromClient(ctx, client, platform)
+		var normPlatform *ocispecs.Platform
+                if platform != nil {
+                        p := platforms.Normalize(*platform)
```
cpuguy83: Did you run into a case where the platform needed to be normalized?

targets/linux/rpm/distro/worker.go
```diff
@@ -70,17 +125,94 @@ func (cfg *Config) Worker(sOpt dalec.SourceOpts, opts ...llb.ConstraintsOpt) llb
...
+                return targetBase.Run(
+                        dalec.WithConstraints(append(opts, llb.Platform(targetPlat))...),
+                        cfg.Install(cfg.BuilderPackages),
+                ).Root()
+        }
+
+	// tdnf-based distros (Azure Linux / CBL-Mariner) do not have a reliable
+        // cross-installroot flow (no --forcearch), so fall back to running the
+        // package manager on the target platform under qemu.
+        if cfg.CacheDir == "/var/cache/tdnf" {
```
cpuguy83: It would be great if we can abstract this a bit.
For the case of azlinux/mariner we can `tdnf install dnf`.
Its just unfortunate that right now we pass in an image ref and something else pulls it.

I'm thinking instead if we just have some helpers that each distro calls.
I *think* this would require passing a function around  (because we need details not available at compile time) or do some interface like:

```
type Worker interface {
  Worker(...) llb.State
}
```

But in any case... the idea here being shifting some logic into each distro's setup rather than a centralized config that makes it hard to do such customization..

cpuguy83 (changes requested):

targets/linux/rpm/distro/worker.go
cpuguy83: Also, having azlinux/mariner install dnf up front also means we can delete a bunch of code that handles tdnf.

targets/linux/rpm/distro/worker.go
kartikjoshi21: I didn’t hit a concrete repro of the frontend “running as TARGET and lying”. The goal here was simply “don’t guess” the executor platform — BuildOpts already tells us the worker platforms, so we use that as best-effort and keep DefaultSpec as a fallback. I have reworded the comment, we can remove this completely if you think thats better, it is kept just as safeguard

targets/linux/rpm/distro/worker.go
kartikjoshi21: No, I didn’t see a concrete failure in SourceOptFromClient due to lack of normalization. I’ll remove normalization there and rely on Normalize only for comparisons / arch mapping.

targets/linux/rpm/distro/distro.go
```diff
@@ -37,12 +37,14 @@ type Config struct {
...
 	// Whether to namespace the cache key by platform
 	// Not all distros need this, hence why it is configurable.
-	// The cache key is only added when the build platform and target platform differ.
 	CacheAddPlatform bool
 
 	// e.g. /var/cache/tdnf or /var/cache/dnf
 	CacheDir string
 
+	// Additional cache directories to mount (e.g. tdnf-based distros may also want /var/cache/dnf).
+	ExtraCacheDirs []string
```
cpuguy83: Can we make `CacheDir` a []string?

kartikjoshi21: Makes sense, i have tried to make it looks simpler using above suggestions please check, Thanks

cpuguy83 (approved):
Much cleaner here, thanks for the update!

Some basic test results like I did for the deb changes:

```
amd64->arm64
pr - 5m9s
main - 6m20s

amd64->amd64
pr - 2m14s
main - 2m20s
```

PR #933, opened 2026-01-22
Add go mod vendor to gomod-patch

willie-yao:
This PR adds `go mod vendor` after `go mod tidy` for patching to support projects using vendors

Fixes #

### DIFF
--- a/scripts/gomod-patch.sh
+++ b/scripts/gomod-patch.sh
@@ -107,9 +107,16 @@ EOF
     if [ "$vendor_existed" = "true" ]; then
       # If go.work exists, use 'go work vendor' instead of 'go mod vendor'
       # because 'go mod vendor' cannot be run in workspace mode
+      # Note: 'go work vendor' requires Go 1.22+
       if [ -f "$gowork_path" ]; then
-        echo "  Running go work vendor to sync workspace vendor directory"
-        go work vendor
+        # Check if 'go work vendor' is available (Go 1.22+)
+        if go work vendor -h >/dev/null 2>&1; then
+          echo "  Running go work vendor to sync workspace vendor directory"
+          go work vendor
+        else
+          echo "  Warning: go work vendor not available (requires Go 1.22+), using GOWORK=off go mod vendor"
+          GOWORK=off go mod vendor
+        fi
       else
         echo "  Running go mod vendor to sync existing vendor directory"
         go mod vendor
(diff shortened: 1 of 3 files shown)


scripts/gomod-patch.sh
```diff
@@ -59,27 +59,64 @@ EOF
...
       cp "$gosum_path" "$tmpdir/$GOSUM_FILENAME"
     else
       : > "$tmpdir/$GOSUM_FILENAME"
     fi
 
+    # Backup vendor directory if it exists
+    vendor_existed=false
+    if [ -d "$vendor_path" ]; then
+      vendor_existed=true
+      cp -r "$vendor_path" "$tmpdir/vendor"
```
cpuguy83: Why do we need to backup?

cpuguy83: I guess this is used for diffing?
If we really need this, instead of copying here we should be able to mount the original state as read-only at some predefined path.

This avoids the need to do a copy.

cpuguy83 (changes requested):

scripts/gomod-patch.sh
willie-yao: Yup it's just for diffing. I'll give that a try

test/linux_target_test.go
```diff
@@ -2053,6 +1982,84 @@ func main() {
 		})
 	})
 
+	t.Run("gomod replace with vendor directory", func(t *testing.T) {
```
cpuguy83, line 1987: I think we have several configurations that need to be tested here:

* go.mod only
* go.mod with go.work
* go.mod with vendor
* go.mod with go.work and vendor

willie-yao: Added in https://github.com/project-dalec/dalec/pull/933/commits/a662ab83cd5c84e14a765d160a28c3cea80cc718

Also added a test for when go.mod replaces update the go.mod version, it should also update the go version in go.work.

willie-yao: Still working on these tests. go work vendor didn't exist before go 1.22 so there's some issues with the older versions on the latter two tests you listed

willie-yao: I decided to just skip the vendor tests for the versions that don't have go work vendor, and set the go versions to 1.18 in the test to support some of the older versions as well https://github.com/project-dalec/dalec/pull/933/commits/a4e7b3855611960f0ec0bc71ce78a1349389b151

cpuguy83 (approved):

PR #938, opened 2026-01-23
test: gomod git auth: remove need for custom buildx instance

cpuguy83:
This should speed up the test since we don't have to re-cache everything in the new buildx instance.
It also means we don't need to setup any listeners in the host network namespace.

It accomplishes this by connecting directly to the git server's container IP address which we detect when we start the server.

There are likely some other improvements we can do here but the goal was to remove the custom buildx instance for this change.

### DIFF
--- a/test/gomod_git_auth_test.go
+++ b/test/gomod_git_auth_test.go
@@ -104,6 +104,7 @@ go {{ .ModFileGoVersion }}
 		ctx := startTestSpan(baseCtx, t)
 
 		testEnv.RunTest(ctx, t, func(ctx context.Context, client gwclient.Client) {
+			attr := attr
 			testState := gitservices.NewTestState(t, client, &attr)
 
 			worker, _, gitHost := initStates(&testState)
@@ -161,6 +162,7 @@ go {{ .ModFileGoVersion }}
 		agentErrChan := startSSHAgent(t, privkey, sockaddr)
 
 		testEnv.RunTest(ctx, t, func(ctx context.Context, client gwclient.Client) {
+			attr := attr
 			testState := gitservices.NewTestState(t, client, &attr)
 
 			_, repo, gitHost := initStates(&testState)


test/git_services/teststate.go
```diff
@@ -301,94 +279,126 @@ func (b *bufCloser) Close() error {
...
-func (ts *TestState) runContainer(cont gwclient.Container, env []string, s Script) <-chan error {
-	ctx := ts.Ctx
+// runContainerWithEventsDirect runs a container with the given args and returns a containerProcess.
+func (ts *TestState) runContainerWithEventsDirect(ctx context.Context, cont gwclient.Container, args []string) *containerProcess {
+	var (
+		t = ts.T
+	)
 
-	stdout := bufCloser{bytes.NewBuffer(nil)}
+	pr, pw := io.Pipe()
```
invidian: nit: maybe `stdoutRead` and `stdoutWrite` would make this code a bit easier to read. `pw`, `pr`, `ec`, `cp`, that's a lot of cryptic acronyms to keep track of for a new reader (like me) it seems.

test/git_services/teststate.go
```diff
@@ -301,94 +279,126 @@ func (b *bufCloser) Close() error {
...
-	if err != nil {
-		ts.T.Logf("unable to copy env: %s", err)
+// waitForReady waits for a Ready event from the container process.
+func (ts *TestState) waitForReady(ctx context.Context, proc *containerProcess, timeout time.Duration) *ReadyEvent {
+	t := ts.T
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	// Close the pipe on timeout to unblock the decoder
+	go func() {
```
invidian, line 375: Would it make sense for `Events()` to return a channel, so we can here for select on both context and the channel itself, to avoid a separate goroutine to run and maintain?

cpuguy83: My initial thought here is that we still need the extra goroutine otherwise we'll end up blocking while reading the stream.
However I see this is mixing ownership of the stream which isn't great.
One thing I like about the current approach is it keeps the control flow at the caller--no hidden behavior.

Let me see if I can clean this up a bit.

cpuguy83: Spent a bit of time on this and I'm not sure making changes here is worth it.
I wouldn't use this generically, however for the use-case this satisfies everything we need.
Pushing the timeout handling further in the stack is even less desirable than managing the stream here.

test/git_services/teststate.go
cpuguy83: Fixed.

Issue #448, opened 2024-11-26
[REQ] "Distroless" Ubuntu/Debian images


cpuguy83:
### What kind of request is this?

None

### What is your request or suggestion?

When we were working on Ubuntu Jammy support we originally had support for building a "distroless" image.
We did this by relying on mmdebstrap.
Unfortunately we found the only way to use mmdebstrap to produce a correct image was to enable privileged mode so mmdebstrap can setup chroots.
mmdebstrap does support an unprivileged mode, however a lot of things didn't work properly due to debhelper/maintainer scripts in deb packages not being able to handle this properly. An example of this was dh_systemd failing to put things in the right place.

We should definitely support this, though.
One option can be to use mmdebstrap (where supported) and require privileged mode.
This is at least a relatively low-effort way to achieve this, but it does require people to reconfigure their buildkit daemons and build invocations to allow the privileged entitlement.

We *should* be able to do what mmdebstrap does ourselves since we control all the things that mmdebstrap wants to control. This will take a bit more work to tease apart and implement all the requirements.

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

PR #945, opened 2026-01-26
Add more tests to building (deb) containers

invidian:

As part of #448, to minimize a chance for regression once we re-implement deb container building to use minimal images, this PR adds additional tests to deb container building to cover untested parts of code. New tests can easily be extended to cover RPM containers as well.

Some tests cannot be written as integration tests, hence they are written as unit tests for container building.

Part of #448

### DIFF
--- a/targets/linux/deb/distro/container_test.go
+++ b/targets/linux/deb/distro/container_test.go
@@ -1,0 +1,479 @@
+package distro
+
+import (
+	"context"
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/moby/buildkit/client/llb"
+	"github.com/moby/buildkit/client/llb/sourceresolver"
+	"github.com/moby/buildkit/frontend/gateway/client"
+	"github.com/moby/buildkit/solver/pb"
+	"github.com/opencontainers/go-digest"
+	"github.com/project-dalec/dalec"
+	"github.com/project-dalec/dalec/internal/test"
+)
+
+func Test_Building_container(t *testing.T) {
+	t.Parallel()
+
+	// Base image is test suite specific, since RPM packages can not always be installed on base
+	// DEB image, hence local test instead of integration one.
+	t.Run("uses_base_image_from_spec", func(t *testing.T) {
+		t.Parallel()
+
+		c := &Config{
+			DefaultOutputImage: "foo",
+		}
+
+		client := &testClient{
+			buildOpts: client.BuildOpts{},
+		}
+
+		expectedRef := "bar"
+
+		spec := &dalec.Spec{
+			Image: &dalec.ImageConfig{
+				Bases: []dalec.BaseImage{
+					{
+						Rootfs: dalec.Source{
+							DockerImage: &dalec.SourceDockerImage{
+								Ref: expectedRef,
+							},
+						},
+					},
+				},
+			},
+		}
+
+		ctx := t.Context()
+
+		state := c.BuildContainer(ctx, client, dalec.SourceOpts{}, spec, "target", llb.State{})
+
+		ops, err := test.LLBOpsFromState(ctx, state)
+		if err != nil {
+			t.Fatalf("failed to get llb ops from state: %v", err)
+		}
+
+		if len(ops) == 0 {
+			t.Fatalf("expected at least one llb op, got none")
+		}
+
+		s := ops[0].Op.GetSource()
+		if s == nil {
+			t.Fatalf("expected source op, got nil")
+		}
+
+		if !strings.Contains(s.Identifier, expectedRef) {
+			t.Fatalf("expected source identifier to contain %q, got %q", expectedRef, s.Identifier)
+		}
+	})
+
+	// Most tests here must assert values in Config which are not user-configurable, hence
+	// they cannot be placed in integration tests.
+	t.Run("uses_default_output_image_if_spec_does_not_set_one", func(t *testing.T) {
+		t.Parallel()
+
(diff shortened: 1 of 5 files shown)


test/linux_target_test.go
```diff
@@ -901,6 +902,342 @@ index 0000000..5260cb1
...
+				)
+
+				res := solveT(ctx, t, gwc, sr)
+
+				ops, err := dalec.LLBOpsFromState(ctx, resultToState(t, res))
+				if err != nil {
+					t.Fatalf("Unexpected error extracting LLB OPs from state: %v", err)
+				}
+
+				cacheIgnored := 0
```
invidian, line 928: To make this test more robust, we could modify the build step to include the timestamp when building a package and to generate a timestamp when package is installed. This way, we could assert that the package building timestamp does not change between runs, but installation timestamp does.

test/linux_target_test.go
```diff
@@ -901,6 +902,342 @@ index 0000000..5260cb1
...
+								},
+								Envs: []string{"build"},
+							},
+						},
+					},
+					Build: dalec.ArtifactBuild{
+						Steps: []dalec.BuildStep{
+							{
+								Command: `
+# This is not a debian build, skip this.
```
invidian, line 1053: We should be able to add a separate step for RPM-based builds to also cover RPM implementation with this test.

helpers.go
```diff
@@ -699,3 +701,52 @@ func ErrorStateOption(err error) llb.StateOption {
 		return asyncState(in, err)
 	}
 }
+
+// LLBOpsFromState extracts the list of LLB operations from the provided LLB state.
+//
+// This function and LLBOp type has been inspired by
+// https://github.com/moby/buildkit/blob/c70e8e666f8f6ee3c0d83b20c338be5aedeaa97a/cmd/buildctl/debug/dumpllb.go#L59.
+func LLBOpsFromState(ctx context.Context, state llb.State) ([]LLBOp, error) {
```
cpuguy83: These should be in internal/ since they aren't intended for public use.

targets/linux/deb/distro/container_test.go
```diff
@@ -0,0 +1,479 @@
+package distro_test
```
cpuguy83: I'm not a fan of this pattern.
Any particular reason to make a separate package here?

helpers.go
invidian: Right, done.

targets/linux/deb/distro/container_test.go
invidian: From [testing](https://pkg.go.dev/testing) docs:
>If the file is in a separate "_test" package, the package being tested must be imported explicitly and only its exported identifiers may be used. This is known as "black box" testing.

This style enforces no internal code being used or tested in tests, which should be yielding more robust tests and easier to refactor implementation, since public API is a contract and all observable or modifiable behaviour should be accessible via public API.

I believe this is a good pattern to promote as a default and treat other cases like possible red flags and exceptions.

targets/linux/deb/distro/container_test.go
```diff
@@ -0,0 +1,480 @@
...
+					}
+				}
+			}
+
+			if !found {
+				t.Errorf("Expected to find mount for extra install repo %q, but did not", extraInstallRepo)
+			}
+		})
+
+		t.Run("with_mounted_apt_cache", func(t *testing.T) {
```
invidian, line 187: I _guess_ that could have an integration tests which does not rely on LLB inspection, but we would have to run build for all targets and verify that cache is empty for each target and that it is persistent across different runs.

And also verify that dpkg/apt actually writes to those paths something we expect to cache. That seem like a lot of work though.

targets/linux/deb/distro/container_test.go
cpuguy83: I understand the semantics, my point here is using a separate package doesn't really buy us anything.
One can just not use unexported things in the test code.

invidian: I mean, I think it buys us as much as any other automatic enforcement vs convention to follow in terms of the promoted ways of testing.

I'm happy to elaborate more on that if you like, but I don't insist on that style. Let me know if you want me to revert it to package without `_test`.

cpuguy83: Sorry yeah I think I would prefer to use just `distro` here.

invidian: Sure, done.

targets/linux/deb/distro/container_test.go
```diff
@@ -0,0 +1,479 @@
...
+			ops, err := test.LLBOpsFromState(ctx, state)
+			if err != nil {
+				t.Fatalf("failed to get llb ops from state: %v", err)
+			}
+
+			aptCacheFound := false
+
+			for _, op := range ops {
+				e := op.Op.GetExec()
+				if e == nil || op.OpMetadata.ProgressGroup.Name != "Install base image packages" {
```
cpuguy83, line 416: Checking these look good, but perhaps it should be referencing a constant (both here and where we create the group)?

invidian: I believe a general rule of thumb would be to not use constants in tests, so that tests can act as a test doubles, so if the constant change, you know what gets affected by it, which makes it more intentional. I know this specific case is low impact since it's only about the progress group, but it still sets an example which may be followed for more important constants in the future.

Let me know if you still want me to convert that to constant.

PR #946, opened 2026-01-27
Use Azure's ubuntu archive in CI

cpuguy83:
archive.ubuntu.com has frequent outages causing CI to fail. Given CI is in Azure anyway, it is best to use Azure's mirror.

Fixes #


invidian (approved):
Bummer that it seems it's not possible to serve CNAME records from dnsmasq, so we wouldn't have to hardcode the IP address, but I'm guessing for the time CI runs, that shouldn't make a noticeable difference.

PR #952, opened 2026-01-30
Fix version matching test

willie-yao:
This PR fixes an issue with the version matching test where the go version gets bumped too high and we need to set whether each distro supports go mod version updates

Fixes #

### DIFF
--- a/test/linux_target_test.go
+++ b/test/linux_target_test.go
@@ -8,7 +8,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"go/version"
 	"io"
 	"io/fs"
 	"os"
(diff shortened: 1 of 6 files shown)


test/linux_target_test.go
```diff
@@ -97,9 +96,10 @@ type testLinuxConfig struct {
 		Units   string
 		Targets string
 	}
-	Libdir  string
-	Worker  workerConfig
-	Release OSRelease
+	Libdir    string
+	Worker    workerConfig
+	Release   OSRelease
+	GoVersion string
```
cpuguy83: Can t his be a bool instead for checking the specific feature?

willie-yao: Do you mean a bool whether it supports go mod version update? That would be simpler but I'm thinking there may be other cases in the future where go version is needed. I'm happy to change it though just need a bit of clarification

cpuguy83: Yeah as you said, a bool if go mod version update is supported.

willie-yao: Done in https://github.com/project-dalec/dalec/pull/952/commits/9ef8d81232cea34b06fed98e736c6dc6c7b98eb1

PR #957, opened 2026-02-03
targets/linux/deb/distro: add minimal images target for DEB distros

### LINKED ISSUE #448: [REQ] "Distroless" Ubuntu/Debian images

invidian:

    This PR adds a new target for building minimal images for DEB-based
    distros.

    In new implementation, base image for DEB-based distros is not required,
    although still supported. New implementation in a nutshell downloads all
    required DEB packages into a volume using work image, extracts them into
    a target image then runs dpkg --install to run post-install scripts in
    the proper environment.

    It also moves container-related tests to a separate function so they can
    be executed to test both implementations while we keep the old,
    experimental implementation around.

Fixes #448 

### DIFF
--- a/targets/linux/deb/distro/container.go
+++ b/targets/linux/deb/distro/container.go
@@ -193,25 +193,27 @@ essential_packages=$(dpkg-query -Wf '${Package} ${Essential}\n' | awk '$2 == "ye
 # the container, we define them here.
 bootstrap_extra_packages="passwd"
 
+# Local spec-built .deb files. Passing these by path to apt-get (apt 1.1+
+# syntax — available on every distro dalec supports) lets apt parse the
+# control files itself and resolve dependencies natively, including:
+#   - Pre-Depends (in addition to Depends)
+#   - alternatives ("pkg-a | pkg-b") — apt picks an installable option
+#   - virtual packages (Provides) — apt picks a real provider
+#   - version constraints — unsatisfiable ones cause the build to fail
+#   - architecture restrictions
 local_package_files=$(ls /spec-packages/*.deb)
-
-# Get names of local packages so we can exclude them from apt-get install.
-local_package_names=$(for f in ${local_package_files}; do dpkg-deb -f "${f}" Package 2>/dev/null; done | sort -u)
-
-# Extract dependencies of local packages, since we need to download those as well.
-#
-# Spec packages may depend on base packages, so we need to filter to only download remaining packages, since downloading local packages
-# would fail.
-dependencies_to_download=$(for f in ${local_package_files}; do dpkg-deb -f "${f}" Depends 2>/dev/null; done | tr ',' '\n' | sed 's/([^)]*)//g; s/|.*//; s/ //g' | grep -v '^$' | sort -u | grep -vxF "${local_package_names}" || true)
 
 # Get the exact filenames apt needs by using --print-uris with an empty cache dir.
 # This forces apt to report ALL needed packages (not just uncached ones), giving
 # us exact filenames including correct version and architecture suffixes.
 # --print-uris output format: 'URL' filename size hash
 # We extract the second field (the filename).
+#
+# Local .deb paths are recognized as already-available and don't appear in
+# --print-uris output, so the filenames here are only remote deps.
 needed_filenames=$(apt-get -o Dir::State::status="${rootfs}/var/lib/dpkg/status" \
     -o Dir::Cache::Archives=/tmp \
-    --yes --print-uris install ${essential_packages} ${bootstrap_extra_packages} ${dependencies_to_download} \
+    --yes --print-uris install ${essential_packages} ${bootstrap_extra_packages} ${local_package_files} \
     | grep '\.deb ' | awk '{print $2}')
 
 mkdir -p "${rootfs}${apt_archives}"/partial
@@ -228,9 +230,11 @@ done
 
 # Download remaining needed packages directly into the rootfs cache.
 # apt skips packages already present, so only missing ones are fetched.
+# Passing the local .deb paths anchors the install plan to the spec packages
+# without re-fetching them.
 apt-get -o Dir::State::status="${rootfs}/var/lib/dpkg/status" \
     -o Dir::Cache::Archives="${rootfs}${apt_archives}" \
-    --yes --download-only install ${essential_packages} ${bootstrap_extra_packages} ${dependencies_to_download}
+    --yes --download-only install ${essential_packages} ${bootstrap_extra_packages} ${local_package_files}
 
 deb_files=$(ls "${rootfs}${apt_archives}"/*.deb)
 
@@ -359,6 +363,15 @@ rm -f /var/lib/dpkg/info/libpam-runtime.prerm 2>/dev/null || true
 
 # Recursive dependency resolver: prints the transitive closure of installed
 # Depends/Pre-Depends starting from the given space-separated package list.
+#
+# Handles three tricky cases that a naive dpkg-query+sed pipeline gets wrong:
+#   - Alternatives ("pkg-a | pkg-b"): walk every option, keep whichever is
+#     installed (rather than blindly picking the first).
+#   - Virtual packages (Provides): if a dep name is not an installed package
+#     itself, look up installed packages whose Provides field lists it and
+#     keep those instead. Without this, virtual deps like "awk" silently
+#     drop their real provider (mawk/gawk) from the keep set.
+#   - Multi-arch qualifiers (":amd64"): stripped so the name matches.
 resolve_deps() {
     queue="$1"
     resolved=""
@@ -370,17 +383,47 @@ resolve_deps() {
 
         resolved="${resolved} ${pkg}"
 
+        # Strip version constraints "(>= 1.0)", arch restrictions "[amd64]",
+        # whitespace, and multi-arch qualifiers ":amd64". Keep alternatives
+        # ("|") as-is; they are split below.
         deps=$(dpkg-query -W -f='${Depends}\n${Pre-Depends}\n' "${pkg}" 2>/dev/null \
-            | tr ',' '\n' | sed 's/([^)]*)//g; s/|.*//; s/ //g; s/:.*//g' | grep -v '^$' | sort -u)
+            | tr ',' '\n' | sed 's/([^)]*)//g; s/\[[^]]*\]//g; s/[[:space:]]//g; s/:[a-z0-9-]*//g' | grep -v '^$' | sort -u)
 
-        for dep in ${deps}; do
-            if ! dpkg -s "${dep}" 2>/dev/null | grep -q '^Status: install ok installed'; then
-                continue
-            fi
-            if echo "${resolved}" | grep -qw "${dep}"; then
-                continue
-            fi
-            queue=$(printf '%s\n%s' "${queue}" "${dep}")
+        for dep_alt in ${deps}; do
+            # Walk all alternatives ("pkg-a|pkg-b") so we don't lose track
+            # of whichever option is actually installed.
+            for dep in $(echo "${dep_alt}" | tr '|' ' '); do
+                if [ -z "${dep}" ] || echo "${resolved}" | grep -qw "${dep}"; then
+                    continue
+                fi
+
+                # Real, directly-installed package.
+                if dpkg -s "${dep}" 2>/dev/null | grep -q '^Status: install ok installed'; then
+                    queue=$(printf '%s\n%s' "${queue}" "${dep}")
+                    continue
+                fi
+
+                # Virtual package: find installed packages whose Provides
+                # field lists this name (with optional version constraint
+                # stripped) and keep them.
+                providers=$(dpkg-query -W -f='${Package}|${Provides}\n' 2>/dev/null \
+                    | awk -F'|' -v want="${dep}" '
+                        $2 == "" { next }
+                        {
+                            n = split($2, list, ",")
+                            for (i = 1; i <= n; i++) {
+                                name = list[i]
+                                sub(/\(.*$/, "", name)
+                                gsub(/^[ \t]+|[ \t]+$/, "", name)
+                                if (name == want) { print $1; next }
+                            }
+                        }
+                    ')
+                for prov in ${providers}; do
+                    if echo "${resolved}" | grep -qw "${prov}"; then continue; fi
+                    queue=$(printf '%s\n%s' "${queue}" "${prov}")
+                done
+            done
         done
     done
     echo "${resolved}"
--- a/targets/linux/deb/distro/container.go
+++ b/targets/linux/deb/distro/container.go
@@ -487,6 +487,39 @@ keep_set="${keep_set} base-files"
 # transitively), in which case we keep them and their deps.
 keep_set=$(resolve_deps "$(echo ${keep_set} | tr ' ' '\n')")
 
+# Surface the resolved keep set in build logs for diagnostic purposes.
+# A small keep_set (just the spec package itself, no transitive deps)
+# is a legitimate outcome for specs that declare no runtime deps and
+# whose generated .deb's Depends field is empty — those builds intend
+# the cleanup to purge everything except the spec package. We log the
+# resolved set so unexpected smallness is at least visible to anyone
+# triaging "the binary I expected is missing" issues, but we do NOT
+# fail the build here.
+echo "DALEC keep_set (resolved): ${keep_set}"
+
+# Persist the keep set so the worker dpkg-remove step (which runs against
+# /target using --root=) can validate that every keep-set package is still
+# installed after both purge passes. We can't reliably validate this from
+# inside the container: purge_last may remove dpkg's own shared libraries
+# (libmd, libc, libpam...) before we get to run dpkg-query, producing
+# false "missing" verdicts. The worker has its own working dpkg + libs and
+# can safely query the target rootfs via --root=/target.
+#
+# Only the names that correspond to actually-installed packages are
+# persisted. The seed includes the raw Depends names (e.g. 'awk') which
+# resolve_deps then follows via Provides to a real installed package
+# (e.g. 'mawk'); both end up in ${keep_set}, but the worker can only
+# audit the latter — dpkg-query against the virtual name 'awk' would
+# return 'not-installed' and produce a false-positive validation
+# failure even though the build is correct.
+mkdir -p /var/lib/dpkg
+: > /var/lib/dpkg/.dalec-keep-set
+for pkg in ${keep_set}; do
+    if dpkg-query -W -f='${db:Status-Status}\n' "${pkg}" 2>/dev/null | grep -qx installed; then
+        printf '%s\n' "${pkg}" >> /var/lib/dpkg/.dalec-keep-set
+    fi
+done
+
 # purge_last: cleanup tools (+ their deps) not in the keep set. These
 # survive the main purge so they remain available for it, then get purged
 # at the very end.
@@ -570,7 +603,6 @@ fi
 # Remove leftover directories (after dpkg purge so maintainer scripts still work).
 cleanup_dirs="
 /etc/apt
-/etc/systemd
 /usr/lib/apt
 /usr/share/bash-completion
 /usr/share/bug
@@ -581,9 +613,57 @@ cleanup_dirs="
 /var/cache/debconf
 /var/lib/apt
 /var/lib/pam
-/var/lib/systemd
 /var/log
 "
+
+# Preserve /etc/systemd and /var/lib/systemd if the final image actually
+# uses systemd. Otherwise, these are dead config / state directories that
+# can be pruned.
+#
+# We check the dpkg status here (after purge_first has run) rather than
+# only looking at keep_set, because keep_set tracks names from the spec
+# .deb but systemd may also have arrived via a base-image dependency,
+# Provides, or by being pulled in as a Recommends/Suggests.
+#
+# Detection covers three signals, in order:
+#   1. The systemd PID-1 binary itself is on disk. This is the most
+#      reliable indicator because dpkg-deb --extract unpacks files
+#      regardless of whether postinst later succeeds. In stripped /
+#      minimal containers systemd's postinst frequently fails (no init,
+#      no D-Bus, can't enable units), leaving the package in
+#      'half-configured' state — but the binary is present and
+#      /etc/systemd still matters.
+#
+#      We deliberately do NOT check for /usr/lib/systemd/system as a
+#      proxy: many non-systemd packages (e2fsprogs, init-system-helpers,
+#      dbus, etc.) ship unit files there, so its presence does not imply
+#      that systemd itself is installed.
+#   2. dpkg-query reports any installed-ish state for the systemd
+#      package (covers the half-configured case explicitly, in case the
+#      binary lives in a non-standard location).
+#   3. systemctl on PATH — pragmatic fallback for custom base images
+#      where systemd is set up by other means.
+keep_systemd=0
+if [ -x /usr/lib/systemd/systemd ] || [ -x /lib/systemd/systemd ]; then
+    keep_systemd=1
+fi
+if [ "${keep_systemd}" != "1" ]; then
+    case "$(dpkg-query -W -f='${db:Status-Status}\n' systemd 2>/dev/null)" in
+        installed|half-configured|triggers-awaited|triggers-pending)
+            keep_systemd=1
+            ;;
+    esac
+fi
+if [ "${keep_systemd}" != "1" ] && command -v systemctl >/dev/null 2>&1; then
+    keep_systemd=1
+fi
+
+if [ "${keep_systemd}" != "1" ]; then
+    cleanup_dirs="${cleanup_dirs}
+/etc/systemd
+/var/lib/systemd
+"
+fi
 
 if [ "${DALEC_HAS_DOCS}" != "true" ]; then
     cleanup_dirs="${cleanup_dirs}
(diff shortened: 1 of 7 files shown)


targets/linux/deb/distro/container.go
```diff
@@ -136,3 +140,159 @@ func installPackagesInContainer(input buildContainerInput, ro []llb.RunOption) l
...
+# us exact filenames including correct version and architecture suffixes.
+# --print-uris output format: 'URL' filename size hash
+# We extract the second field (the filename).
+needed_filenames=$(apt-get -o Dir::State::status="${rootfs}/var/lib/dpkg/status" \
+    -o Dir::Cache::Archives=/tmp \
+    --yes --print-uris install ${essential_packages} ${dependencies_to_download} \
+    | grep '\.deb ' | awk '{print $2}')
+
+mkdir -p "${rootfs}${apt_archives}"/partial
+cp ${local_package_files} "${rootfs}${apt_archives}"/
```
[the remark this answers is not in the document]
invidian: This is not relevant, at least one local package must exist, otherwise whole script does not make any sense.

invidian:
Thanks for feedback, let me see if more can be removed and I'll document why not. Will also graph the dependencies, so we know why we have these packages. Perhaps we can drop ca-certificates from deb base images then.

invidian:
>Can we clean up dash after everything?

I think we could, but we might need to do some magic and use worker image with dpkg to remove it, since dpkg itself depends on dash. I'm not sure it's worth the effort since it has just one dependency. This approach could simplify the cleanup process though, so I might have a look.

On the other hand, we could try to remove `passwd`, since it pulls so much. Either by writing directly to etc files or add users/groups using busybox `adduser/addgroup`.

>I built the go-md2man example image and checked the installed packages:

Which image you built? For trixie, this is the list of packages we end up with, if we remove `ca-certificates` as well, although openssl does not pull a lot of deps.
```
base-files
base-passwd
dalec-deb-base-packages
dash
debconf
debianutils
dpkg
gcc-14-base
go-md2man
go-md2man-dbgsym
libacl1
libattr1
libaudit-common
libaudit1
libbsd0
libbz2-1.0
libc-bin
libc6
libcap-ng0
libcap2
libcrypt1
libdb5.3t64
libdebconfclient0
libgcc-s1
libmd0
libpam-modules
libpam-modules-bin
libpam0g
libpcre2-8-0
libselinux1
libsemanage-common
libsemanage2
libsepol2
libstdc++6
libsystemd0
login.defs
passwd
```

```mermaid
graph LR
    base-files
    base-passwd
    base-passwd --> libc6
    base-passwd --> libdebconfclient0
    base-passwd --> libselinux1
    dalec-deb-base-packages
    dalec-deb-base-packages --> passwd
    dash
    dash --> debianutils
    dash --> libc6
    debconf
    debianutils
    debianutils --> libc6
    gcc-14-base
    go-md2man
    go-md2man-dbgsym
    go-md2man-dbgsym --> go-md2man
    libacl1
    libacl1 --> libc6
    libattr1
    libattr1 --> libc6
    libaudit-common
    libaudit1
    libaudit1 --> libaudit-common
    libaudit1 --> libc6
    libaudit1 --> libcap-ng0
    libbsd0
    libbsd0 --> libc6
    libbsd0 --> libmd0
    libbz2-1.0
    libbz2-1.0 --> libc6
    libc-bin
    libc-bin --> libc6
    libc6
    libc6 --> libgcc-s1
    libcap-ng0
    libcap-ng0 --> libc6
    libcap2
    libcap2 --> libc6
    libcrypt1
    libcrypt1 --> libc6
    libdb5.3t64
    libdb5.3t64 --> libc6
    libdebconfclient0
    libdebconfclient0 --> libc6
    libgcc-s1
    libgcc-s1 --> gcc-14-base
    libgcc-s1 --> libc6
    libmd0
    libmd0 --> libc6
    libpam-modules
    libpam-modules --> debconf
    libpam-modules --> libaudit1
    libpam-modules --> libc6
    libpam-modules --> libcrypt1
    libpam-modules --> libdb5.3t64
    libpam-modules --> libpam-modules-bin
    libpam-modules --> libpam0g
    libpam-modules --> libselinux1
    libpam-modules --> libsystemd0
    libpam-modules-bin
    libpam-modules-bin --> libaudit1
    libpam-modules-bin --> libc6
    libpam-modules-bin --> libcrypt1
    libpam-modules-bin --> libpam0g
    libpam-modules-bin --> libselinux1
    libpam-modules-bin --> libsystemd0
    libpam0g
    libpam0g --> debconf
    libpam0g --> libaudit1
    libpam0g --> libc6
    libpcre2-8-0
    libpcre2-8-0 --> libc6
    libselinux1
    libselinux1 --> libc6
    libselinux1 --> libpcre2-8-0
    libsemanage-common
    libsemanage2
    libsemanage2 --> libaudit1
    libsemanage2 --> libbz2-1.0
    libsemanage2 --> libc6
    libsemanage2 --> libselinux1
    libsemanage2 --> libsemanage-common
    libsemanage2 --> libsepol2
    libsepol2
    libsepol2 --> libc6
    libsystemd0
    libsystemd0 --> libc6
    libsystemd0 --> libcap2
    login.defs
    passwd
    passwd --> base-passwd
    passwd --> libacl1
    passwd --> libattr1
    passwd --> libaudit1
    passwd --> libbsd0
    passwd --> libc6
    passwd --> libcrypt1
    passwd --> libpam-modules
    passwd --> libpam0g
    passwd --> libselinux1
    passwd --> libsemanage2
    passwd --> login.defs
```

invidian:
```
Package: dash
...
Depends: debianutils (>= 5.6-0.1), dpkg (>= 1.19.1)
```

wat, that's Ubuntu noble...

invidian:
@cpuguy83 please have a look again. I've updated the PR so that minimal containers do not pull BasePackages anymore, so we do not include `ca-certificates`, which results in really minimal dependency trees. If you have no runtime dependency and you statically build Go binary like with `go-md2man` for example, container ends up with just `go-md2man` package, nothing else.

Then, spec package can depend on things like `dpkg` or `curl`:
```mermaid
graph LR
    curl[curl]
    curl --> libc6
    curl --> libcurl4t64
    curl --> zlib1g
    gcc-14-base[gcc-14-base]
    go-md2man[go-md2man]
    go-md2man --> curl
    libbrotli1[libbrotli1]
    libbrotli1 --> libc6
    libc6[libc6]
    libc6 --> libgcc-s1
    libcom-err2[libcom-err2]
    libcom-err2 --> libc6
    libcurl4t64[libcurl4t64]
    libcurl4t64 --> libbrotli1
    libcurl4t64 --> libc6
    libcurl4t64 --> libgssapi-krb5-2
    libcurl4t64 --> libidn2-0
    libcurl4t64 --> libldap2
    libcurl4t64 --> libnghttp2-14
    libcurl4t64 --> libpsl5t64
    libcurl4t64 --> librtmp1
    libcurl4t64 --> libssh-4
    libcurl4t64 --> libssl3t64
    libcurl4t64 --> libzstd1
    libcurl4t64 --> zlib1g
    libdb5_3t64[libdb5.3t64]
    libdb5_3t64 --> libc6
    libffi8[libffi8]
    libffi8 --> libc6
    libgcc-s1[libgcc-s1]
    libgcc-s1 --> gcc-14-base
    libgcc-s1 --> libc6
    libgmp10[libgmp10]
    libgmp10 --> libc6
    libgnutls30t64[libgnutls30t64]
    libgnutls30t64 --> libc6
    libgnutls30t64 --> libgmp10
    libgnutls30t64 --> libhogweed6t64
    libgnutls30t64 --> libidn2-0
    libgnutls30t64 --> libnettle8t64
    libgnutls30t64 --> libp11-kit0
    libgnutls30t64 --> libtasn1-6
    libgnutls30t64 --> libunistring5
    libgssapi-krb5-2[libgssapi-krb5-2]
    libgssapi-krb5-2 --> libc6
    libgssapi-krb5-2 --> libcom-err2
    libgssapi-krb5-2 --> libk5crypto3
    libgssapi-krb5-2 --> libkrb5-3
    libgssapi-krb5-2 --> libkrb5support0
    libhogweed6t64[libhogweed6t64]
    libhogweed6t64 --> libc6
    libhogweed6t64 --> libgmp10
    libhogweed6t64 --> libnettle8t64
    libidn2-0[libidn2-0]
    libidn2-0 --> libc6
    libidn2-0 --> libunistring5
    libk5crypto3[libk5crypto3]
    libk5crypto3 --> libc6
    libk5crypto3 --> libkrb5support0
    libkeyutils1[libkeyutils1]
    libkeyutils1 --> libc6
    libkrb5-3[libkrb5-3]
    libkrb5-3 --> libc6
    libkrb5-3 --> libcom-err2
    libkrb5-3 --> libk5crypto3
    libkrb5-3 --> libkeyutils1
    libkrb5-3 --> libkrb5support0
    libkrb5-3 --> libssl3t64
    libkrb5support0[libkrb5support0]
    libkrb5support0 --> libc6
    libldap2[libldap2]
    libldap2 --> libc6
    libldap2 --> libgnutls30t64
    libldap2 --> libsasl2-2
    libnettle8t64[libnettle8t64]
    libnettle8t64 --> libc6
    libnghttp2-14[libnghttp2-14]
    libnghttp2-14 --> libc6
    libp11-kit0[libp11-kit0]
    libp11-kit0 --> libc6
    libp11-kit0 --> libffi8
    libpsl5t64[libpsl5t64]
    libpsl5t64 --> libc6
    libpsl5t64 --> libidn2-0
    libpsl5t64 --> libunistring5
    librtmp1[librtmp1]
    librtmp1 --> libc6
    librtmp1 --> libgmp10
    librtmp1 --> libgnutls30
    librtmp1 --> libhogweed6
    librtmp1 --> libnettle8
    librtmp1 --> zlib1g
    libsasl2-2[libsasl2-2]
    libsasl2-2 --> libc6
    libsasl2-2 --> libsasl2-modules-db
    libsasl2-2 --> libssl3t64
    libsasl2-modules[libsasl2-modules]
    libsasl2-modules --> libc6
    libsasl2-modules --> libssl3t64
    libsasl2-modules-db[libsasl2-modules-db]
    libsasl2-modules-db --> libc6
    libsasl2-modules-db --> libdb5_3t64
    libssh-4[libssh-4]
    libssh-4 --> libc6
    libssh-4 --> libgssapi-krb5-2
    libssh-4 --> libssl3t64
    libssh-4 --> zlib1g
    libssl3t64[libssl3t64]
    libssl3t64 --> libc6
    libtasn1-6[libtasn1-6]
    libtasn1-6 --> libc6
    libunistring5[libunistring5]
    libunistring5 --> libc6
    libzstd1[libzstd1]
    libzstd1 --> libc6
    zlib1g[zlib1g]
    zlib1g --> libc6
```

```mermaid
graph LR
    dpkg[dpkg]
    dpkg --> libc6
    dpkg --> liblzma5
    dpkg --> libmd0
    dpkg --> libselinux1
    dpkg --> libzstd1
    dpkg --> tarlibbz2-1_0
    dpkg --> zlib1g
    gcc-14-base[gcc-14-base]
    go-md2man[go-md2man]
    go-md2man --> dpkg
    libacl1[libacl1]
    libacl1 --> libc6
    libbz2-1_0[libbz2-1.0]
    libbz2-1_0 --> libc6
    libc6[libc6]
    libc6 --> libgcc-s1
    libgcc-s1[libgcc-s1]
    libgcc-s1 --> gcc-14-base
    libgcc-s1 --> libc6
    liblzma5[liblzma5]
    liblzma5 --> libc6
    libmd0[libmd0]
    libmd0 --> libc6
    libpcre2-8-0[libpcre2-8-0]
    libpcre2-8-0 --> libc6
    libselinux1[libselinux1]
    libselinux1 --> libc6
    libselinux1 --> libpcre2-8-0
    libzstd1[libzstd1]
    libzstd1 --> libc6
    tar[tar]
    tar --> libacl1
    tar --> libc6
    tar --> libselinux1
    zlib1g[zlib1g]
    zlib1g --> libc6
```

cpuguy83 (commented):
Review by GPT-5.5.

The target split and shared container test refactor look good. Main remaining concern is making cleanup policy explicit and safe: prune bootstrap/package-manager detritus, but do not remove files that belong to the final intended image. In particular, systemd dirs should depend on whether `systemd` is installed, and docs preservation should include licenses.

targets/linux/deb/distro/container.go
```diff
@@ -136,3 +141,380 @@ func installPackagesInContainer(input buildContainerInput, ro []llb.RunOption) l
...
+local_package_files=$(ls /spec-packages/*.deb)
+
+# Get names of local packages so we can exclude them from apt-get install.
+local_package_names=$(for f in ${local_package_files}; do dpkg-deb -f "${f}" Package 2>/dev/null; done | sort -u)
+
+# Extract dependencies of local packages, since we need to download those as well.
+#
+# Spec packages may depend on base packages, so we need to filter to only download remaining packages, since downloading local packages
+# would fail.
+dependencies_to_download=$(for f in ${local_package_files}; do dpkg-deb -f "${f}" Depends 2>/dev/null; done | tr ',' '\n' | sed 's/([^)]*)//g; s/|.*//; s/ //g' | grep -v '^$' | sort -u | grep -vxF "${local_package_names}" || true)
```
cpuguy83: This dependency extraction is too lossy for Debian semantics. It only reads `Depends`, ignores `Pre-Depends`, strips version constraints, and chooses the first alternative before `|`, so the bootstrap package set can differ from what apt/dpkg would resolve. Please either move more of this to apt-native resolution or add coverage for `Pre-Depends`, alternatives/virtual packages, arch restrictions, and versioned deps.

targets/linux/deb/distro/container.go
```diff
@@ -136,3 +141,380 @@ func installPackagesInContainer(input buildContainerInput, ro []llb.RunOption) l
...
+# Strip :arch suffixes (e.g. libc6:amd64 -> libc6) so names match.
+for pkg in $(dpkg-query -W -f='${Package}\n' | sed 's/:.*//g'); do
+    if [ "${pkg}" = "dpkg" ]; then continue; fi
+    if echo "${keep_set}" | grep -qw "${pkg}"; then continue; fi
+    if echo "${purge_last}" | grep -qw "${pkg}"; then continue; fi
+    purge_first="${purge_first} ${pkg}"
+done
+
+if [ -n "${purge_first}" ]; then
+    dpkg --purge --force-depends --force-remove-essential ${purge_first} || true
```
cpuguy83, line 605: Since cleanup purges with `--force-depends`, swallowing purge failures can leave a partially cleaned image or inconsistent dpkg database while still succeeding the build. Can we fail on unexpected purge errors, or at least validate the final dpkg status after cleanup?

targets/linux/deb/distro/container.go
```diff
@@ -136,3 +141,380 @@ func installPackagesInContainer(input buildContainerInput, ro []llb.RunOption) l
...
+fi
+
+# Remove leftover directories (after dpkg purge so maintainer scripts still work).
+cleanup_dirs="
+/etc/apt
+/etc/systemd
```
cpuguy83: `/etc/systemd` and `/var/lib/systemd` should not be removed unconditionally. Suggested policy: if the final image has the `systemd` package installed, preserve these dirs; otherwise prune them. Checking the installed `systemd` package in dpkg status is more precise than `libsystemd0`; checking for `systemctl` would also be a pragmatic proxy.

targets/linux/deb/distro/container.go
```diff
@@ -136,3 +141,380 @@ func installPackagesInContainer(input buildContainerInput, ro []llb.RunOption) l
...
+/usr/share/bug
+/usr/share/debconf
+/usr/share/lintian
+/usr/share/locale
+/var/cache/apt
+/var/cache/debconf
+/var/lib/apt
+/var/lib/pam
+/var/lib/systemd
+/var/log
```
cpuguy83: Consider preserving `/var/log` as a directory while removing its contents. Removing the directory itself can surprise packages or processes that expect it to exist.

targets/linux/deb/distro/container.go
```diff
@@ -136,3 +141,380 @@ func installPackagesInContainer(input buildContainerInput, ro []llb.RunOption) l
...
+"
+
+if [ "${DALEC_HAS_DOCS}" != "true" ]; then
```
cpuguy83: This docs preservation condition should also account for license artifacts. Dalec DEB licenses install under `/usr/share/doc`, but `HasDocs()` only covers `Docs` and `Manpages`; a spec with only `Licenses` could have its license files removed during cleanup.

targets/linux/deb/distro/container.go
```diff
@@ -136,3 +141,380 @@ func installPackagesInContainer(input buildContainerInput, ro []llb.RunOption) l
...
+    cleanup_dirs="${cleanup_dirs}
+/usr/share/doc
+/usr/share/man
```
cpuguy83, line 675: Preserving spec-provided manpages is correct. Pruning dependency-owned manpages by default is fine for a minimal image, but that policy should be explicit/documented so users understand that only requested docs/manpages are preserved.

targets/linux/deb/distro/container.go
```diff
@@ -136,3 +141,380 @@ func installPackagesInContainer(input buildContainerInput, ro []llb.RunOption) l
...
+	// No-op stub mounted at /tmp/diff and /tmp/tar so dpkg's maintainer scripts
+	// find the binaries they expect (diff, tar) without writing to the rootfs.
+	stubSt := llb.Scratch().File(llb.Mkfile("stub", 0o755, []byte("#!/bin/sh\nexit 1\n")), cleanupOpts...)
+
+	// Run the main cleanup inside the container (purges everything except dpkg).
+	st = st.Run(
+		dalec.WithConstraints(cleanupOpts...),
+		llb.AddMount("/tmp/dalec-cleanup.sh", scriptSt, llb.SourcePath("cleanup.sh"), llb.Readonly),
+		llb.AddMount("/tmp/dalec-spec-packages", input.SpecPackages, llb.Readonly),
+		llb.AddMount("/tmp/diff", stubSt, llb.SourcePath("stub"), llb.Readonly),
```
cpuguy83, line 845: The `/tmp/diff` and `/tmp/tar` stubs only seem effective for the final purge where `PATH="/tmp:${PATH}"` is set. Maintainer scripts during the first purge pass will not find these unless they call absolute paths. Either prepend `/tmp` for both purge phases or adjust the comment/behavior.

test/linux_target_test.go
```diff
@@ -816,449 +754,211 @@ index 0000000..5260cb1
...
-						withBuildContext(ctx, t, "install-repo-config", installRepoConfig),
-						withBuildContext(ctx, t, "build-repo-config", buildRepoConfig),
-					)
-					solveT(ctx, t, gwc, sr)
+					sr := newSolveRequest(withSpec(ctx, t, &spec), withBuildTarget(target))
+					res := solveT(ctx, t, gwc, sr)
+					ref, err := res.SingleRef()
+					assert.NilError(t, err)
+
+					_, err = ref.StatFile(ctx, gwclient.StatRequest{Path: "/usr/share/doc"})
```
cpuguy83: This should verify the actual doc artifact path/content survives cleanup, not just that `/usr/share/doc` exists. That would catch cleanup accidentally preserving an empty directory while deleting user-requested docs.

test/linux_target_test.go
```diff
@@ -816,449 +754,211 @@ index 0000000..5260cb1
...
-[ -f debian/postinst ] || (echo '#!/bin/sh' > debian/postinst; echo 'set -e' >> debian/postinst)
-[ -x debian/postinst ] || chmod +x debian/postinst
-cat >> debian/postinst << 'EOF'
-grep debug=2 /etc/dpkg/dpkg.cfg.d/99-dalec-debug
-EOF
-`,
+					Tests: []*dalec.TestSpec{
+						{
+							Name: "runtime dep binary works after cleanup",
+							Files: map[string]dalec.FileCheckOutput{
```
cpuguy83, line 1040: This test only stats `/usr/bin/curl` and `/usr/bin/dpkg`. Since the test name says runtime dependencies are functional after cleanup, please execute them too, e.g. `curl --version` and `dpkg --version`, to catch missing interpreters/shared libraries.

targets/linux/deb/distro/container.go
invidian: Addressed I think. Please have a look again.

targets/linux/deb/distro/container.go
invidian: Addressed I think. Please have a look again.

targets/linux/deb/distro/container.go
invidian: Addressed I think. Please have a look again.

targets/linux/deb/distro/container.go
invidian: Addressed I think. Please have a look again.

targets/linux/deb/distro/container.go
invidian: Addressed I think. Please have a look again.

targets/linux/deb/distro/container.go
invidian: Addressed I think. Please have a look again.

test/linux_target_test.go
invidian: Addressed I think. Please have a look again.

test/linux_target_test.go
invidian: Addressed I think. Please have a look again.

targets/linux/deb/distro/container.go
invidian: Addressed I think. Please have a look again.

targets/linux/deb/distro/distro.go
```diff
@@ -139,6 +142,17 @@ func (cfg *Config) Routes(prefix string, spec *dalec.Spec) ([]frontend.Route, er
 				SpecDefined: specDefined,
 			},
 		},
+		{
+			FullPath: prefix + "/container",
```
invidian, line 146: nit: maybe we want to add some "experimental" preffix/suffix here if we have any concerns about this new code, since it hasn't been tested in the wild.

cpuguy83: I don't think that's necessary.
Its a non-breaking change and the intent is to provide a minimal container similar to what we do for rpm-based distros.
If we find something broken with it then we'll just fix it.

targets/linux/deb/distro/container.go
```diff
@@ -136,3 +146,715 @@ func installPackagesInContainer(input buildContainerInput, ro []llb.RunOption) l
...
+#     drop their real provider (mawk/gawk) from the keep set.
+#   - Multi-arch qualifiers (":amd64"): stripped so the name matches.
+resolve_deps() {
+    queue="$1"
+    resolved=""
+    while [ -n "${queue}" ]; do
+        pkg=$(echo "${queue}" | head -n1)
+        queue=$(echo "${queue}" | tail -n +2)
+
+        if [ -z "${pkg}" ] || echo "${resolved}" | grep -qw "${pkg}"; then continue; fi
```
cpuguy83: `grep -qw` isn't safe for matching package names. `-w` treats `-`, `+`, `.` and `:` as word boundaries, so a name matches as a sub-token of a longer one — e.g. `grep -qw python3` matches `python3-minimal`, and `grep -qw dbus` matches `dbus-daemon`.

That bites here in `resolve_deps`: once a longer package (say `python3-minimal`) is already in `${resolved}`, when `python3` is later dequeued this check sees a false match and `continue`s, so `python3` never gets added to `resolved`. It then drops out of the keep set and gets purged → broken image. The worker-side keep-set validation won't catch it either, because the package was dropped *from* the set rather than left half-installed, so it's never in `.dalec-keep-set` to be checked.

Same problem on the enqueue/provider checks just below (~401, ~428) and in the purge loops (~548, ~572-573).

You're already doing the safe thing a few lines down at 561-562 (`echo " ${keep_set} " | grep -q " ${pkg} "`). Suggest standardizing on that exact, space-delimited match everywhere — and using `-F` so a `.` or `+` in a name isn't treated as a regex:

```sh
echo " ${resolved} " | grep -qF " ${pkg} "
```

or pure shell with no subprocess/regex at all:

```sh
case " ${resolved} " in *" ${pkg} "*) continue ;; esac
```

The current fixtures (curl/dpkg/apt/awk) don't have a colliding name pair so this stays latent, but it'll surface for any spec whose dep closure contains one (`python3`/`python3-minimal`, `systemd`/`systemd-sysv`, `dbus`/`dbus-daemon`, etc.).

invidian: Thanks, addressed and added a test case.

test/linux_target_test.go
```diff
@@ -2401,11 +2983,6 @@ func Value() string {
 		ctx := startTestSpan(baseCtx, t)
 		testLinuxSymlinkArtifacts(ctx, t, testConfig)
 	})
-	t.Run("test image configs", func(t *testing.T) {
-		t.Parallel()
-		ctx := startTestSpan(baseCtx, t)
-		testImageConfig(ctx, t, testConfig.Target.Container)
```
invidian, line 3740: I need to address this test removal before merging.

test/linux_target_test.go
```diff
@@ -2401,11 +2983,6 @@ func Value() string {
 		ctx := startTestSpan(baseCtx, t)
 		testLinuxSymlinkArtifacts(ctx, t, testConfig)
 	})
```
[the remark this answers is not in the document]
invidian: Done.

cpuguy83 (approved):

PR #961, opened 2026-02-05
Add code coverage

### LINKED ISSUE #889: [REQ] Add code coverage testing

Other

openssf silver badge requires proof of code coverage.
We don't currently track any coverage.
Furthermore, major of coverage would come through integration tests rather than unit tests.

There likely is some more unit tests we could add to improve coverage, but I would still like to instrument the integration tests such that we can detect coverage there, if possible.

kartikjoshi21:
### Add Go coverage reporting to CI (end-to-end)

This PR adds automated Go coverage reporting to the CI pipeline.

* Unit and integration tests now generate coverage profiles (one per integration matrix suite).
* It also collects coverage for the Dalec BuildKit frontend by building it with coverage enabled and exporting its covdata during solves.
* A new `coverage-report` job downloads all profiles, merges them, and uploads a combined report (including an `index.html` coverage page) as a workflow artifact, plus a summary in the Actions run.

Fixes: #889

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -270,6 +270,7 @@ jobs:
           set -eu
 
           docker buildx bake frontend
+
           if [ "${TEST_SUITE}" = "other" ]; then
             exit 0
           fi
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -276,19 +276,38 @@ jobs:
           docker buildx bake worker
         env:
           TEST_SUITE: ${{ matrix.suite }}
-      - name: Run integration tests
+      - name: Run integration tests (with coverage)
```
MorrisLaw: I feel like this is too vague. Something like `with coverage tracking` or `with coverage tracking enabled` seems more clear. Wdyt?

.github/workflows/ci.yml
```diff
@@ -276,19 +276,38 @@ jobs:
...
+          skip=""
+          if [ -n "${TEST_SUITE}" ] && [ "${TEST_SUITE}" != "other" ]; then
             run="-run=${TEST_SUITE}"
           fi
           if [ -n "${TEST_SKIP}" ]; then
             skip="-skip=${TEST_SKIP}"
           fi
-          go test -timeout=59m -v -json ${run} ${skip} ./test | go run ./cmd/test2json2gha --slow 120s --logdir /tmp/testlogs
+
+          go test -timeout=59m -v -json \
```
cpuguy83, line 307: Isn't this reporting coverage of what get called on the client side but not the actual frontend?
We would need to build the frontend with coverage instrumentation, and then extract the data out of it (maybe if coverage is enabled we can inject the metadata into the result object and extract it in the test client invocation?).

https://go.dev/doc/build-cover

cpuguy83 (changes requested):

.github/workflows/ci.yml
kartikjoshi21: Updated and added full coverage

cmd/frontend/coverage.go
```diff
@@ -0,0 +1,105 @@
+// cmd/frontend/coverage.go
```
cpuguy83: Shouldn't be needed.

cmd/frontend/coverage.go
```diff
@@ -0,0 +1,105 @@
...
+	_ = coverage.ClearCounters()
+
+	return nil
+}
+
+func wrapWithCoverage(next gwclient.BuildFunc) gwclient.BuildFunc {
+	return func(ctx context.Context, c gwclient.Client) (*gwclient.Result, error) {
+		res, err := next(ctx, c)
+		if err != nil {
+			return nil, err
```
cpuguy83: In this case we don't get coverage for any of the error conditions.
Wondering if we can do something like stuff the coverage data into the grpc error metadata?

kartikjoshi21: The issue is that my current code returns too early on the error path in two places: the frontend wrapper and the test harness. That means coverage is only exported on success today. I’m changing both paths to preserve res, attach/export coverage first, and then return the original error

cmd/frontend/coverage.go
```diff
@@ -0,0 +1,113 @@
...
+	return func(ctx context.Context, c gwclient.Client) (*gwclient.Result, error) {
+		res, err := next(ctx, c)
+
+		// If coverage is requested, make sure we have a result object to attach
+		// metadata to even on error paths.
+		if wantFrontendCoverage(c) && res == nil {
+			res = gwclient.NewResult()
+		}
+
+		if covErr := attachFrontendCoverage(c, res); covErr != nil {
```
cpuguy83: Did we test that this worked? It doesn't seem like this affected coverage at all.
I'm wondering if we actually get this result back when there is an error.

kartikjoshi21: yess make sense, my earlier change was trying to address the error-path gap by attaching coverage even when next(...) returned an error. I think to tackle with this, i need a broader change, need to add a new proto to tackle frontend coverage payload and then attaching it to gRPC error path. Looking into changes need to be done for the same. maybe we can keep this PR focussed on success path for error scenarios i will put another PR ?

cpuguy83: Is there an issue with doing it all in this PR?
I'm not convinced we need to merge while still, in theory, inaccurate.

cpuguy83:
Comment from GPT-5.5:

I think frontend coverage is currently being converted with metadata but no counters.

Evidence:

- `go tool covdata` discovers counter files by filename pattern. Go 1.25 defines counter files as `covcounters.<hash>.<pid>.<timestamp>` via `CounterFileTempl = "%s.%x.%d.%d"` and `CounterFileRegexp = `^%s\.(\S+)\.(\d+)\.(\d+)+$`:
  https://github.com/golang/go/blob/go1.25.8/src/internal/coverage/defs.go#L335-L341
- The covdata pod collector scans filenames with that anchored regex and only associates matching counter files with matching `covmeta.<hash>` files:
  https://github.com/golang/go/blob/go1.25.8/src/internal/coverage/pods/pods.go#L121-L160
- Go's own runtime counter writer uses the same `covcounters.<hash>.<pid>.<timestamp>` template:
  https://github.com/golang/go/blob/go1.25.8/src/internal/coverage/cfile/emit.go#L365-L371

This PR writes counter files as `covcounters.<hash>.<pid>.<timestamp>.<random>` in `test/testenv/frontend_coverage.go`, so they do not match the anchored Go regex and are ignored by `go tool covdata textfmt`.

That matches the latest CI artifacts: all 14 `frontend-*.out` profiles are present, but each has `0` covered statements (`covered=0`, `total=6835`, `pct=0.00`). So the frontend profiles are currently contributing denominator, but no executed frontend coverage.

Suggested fix: write counter files using exactly `covcounters.<hash>.<pid>.<timestamp>` and add an assertion that `go tool covdata textfmt` produces at least one non-zero frontend coverage block, rather than only checking that `covmeta.*` / `covcounters.*` files exist.

cpuguy83:
Tried out a fix for this in https://github.com/project-dalec/dalec/pull/1049/changes/1784c3f740f6a32f74c9cd1fa7dacb4fce5130e8
This brings it up to 75% which looks a lot more accurate to me.

kartikjoshi21:
> Tried out a fix for this in [1784c3f](https://github.com/project-dalec/dalec/commit/1784c3f740f6a32f74c9cd1fa7dacb4fce5130e8) This brings it up to 75% which looks a lot more accurate to me.

@cpuguy83 i have added this suggested fix , Thankyou so much for that. Also added few more test and guardrails. Thanks

This is the result 
Total: 73.6%
Profiles merged: 14 integration + 14 frontend

cpuguy83 (approved):

PR #968, opened 2026-02-16
Prepare tests and implementation for adding bootstrapped deb containers implementation

invidian:

This PR refactors existing implementation of building deb containers into smaller functions which will be re-used by the upcoming bootstrapped containers implementation.

It also changes some tests to target package instead of container, so we can better confine container-specific tests and make them re-usable across different targets, namely existing deb testing target and upcoming new minimal containers.

Part of #448

I am not sure whether change of targets reduce our test coverage in some way.

### DIFF
--- a/targets/linux/deb/distro/container.go
+++ b/targets/linux/deb/distro/container.go
@@ -13,81 +13,126 @@ import (
 func (c *Config) BuildContainer(ctx context.Context, client gwclient.Client, sOpt dalec.SourceOpts, spec *dalec.Spec, targetKey string, debSt llb.State, opts ...llb.ConstraintsOpt) llb.State {
 	opts = append(opts, frontend.IgnoreCache(client), dalec.ProgressGroup("Build Container Image"))
 
-	baseImg := llb.Image(c.DefaultOutputImage, llb.WithMetaResolver(sOpt.Resolver), dalec.WithConstraints(opts...))
-
-	bi, err := spec.GetSingleBase(targetKey)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


targets/linux/deb/distro/container.go
```diff
@@ -13,81 +13,124 @@ import (
...
+	// Those base repos come from distro configuration.
+	repos := dalec.GetExtraRepos(input.Config.ExtraRepos, "install")
+
+	// These are user specified via spec.
+	repos = append(repos, input.Spec.GetInstallRepos(input.Target)...)
+
+	return input.Config.RepoMounts(repos, input.SOpt, input.Opts...)
+}
+
+func installPackagesInContainer(baseImg llb.State, input buildContainerInput, ro []llb.RunOption) llb.State {
```
cpuguy83: Maybe make this an llb.StateOption?

cpuguy83: Might actually be worth binding these as methods on buildContainerInput.

invidian: >Maybe make this an llb.StateOption?

Sure, that make perfect sense.

>Might actually be worth binding these as methods on buildContainerInput.

I'd skip that if you don't mind. I think binding a method complects data structure with behaviour, so it yields higher complexity without justification (since we don't deal with e.g. state here like with e.g. with stack implementation), so I'd rather keep it separate. Of course it's a small and not very important piece of code, but I think it still contributes to spreading patterns for keeping the complexity low where possible.

PR #973, opened 2026-02-20
Flatcar: add target for Flatcar-compatible systemd sysext images

### LINKED ISSUE #969: Flatcar distro support
**Background**
Dalec already has a */testing/sysext route under some existing distro targets (e.g. azlinux3/testing/sysext, noble/testing/sysext).
However, these sysext images are currently produced “based on other distros” (Deb/RPM paths + workers), not via a Flatcar-native build environment.

**Problem**
Flatcar extensions are expected to be compatible with Flatcar’s userspace ABI and with Flatcar’s sysext conventions. Flatcar sysext images must follow specific naming + metadata rules:

1. Image should be named NAME.raw
2. Image must contain usr/lib/extension-release.d/extension-release.NAME
3. For Flatcar matching, ID=flatcar plus one of VERSION_ID or SYSEXT_LEVEL is required
So “sysext output exists” is not sufficient; we need a dedicated Flatcar target and a Flatcar-native worker/toolchain.

**Goals (Scope)**

1. Add a dedicated build target for Flatcar
2. Build sysext images using a Flatcar-native build environment
3. Ensure produced sysext artifacts are directly usable on Flatcar:
4. Document usage for Flatcar

kartikjoshi21:
This PR adds an initial Flatcar build target to Dalec focused on producing Flatcar-compatible systemd system extension images (.raw).

- Adds flatcar/testing/sysext which builds a .raw sysext using the Flatcar SDK image as the sysext worker.
- Introduces an optional SysextEnv(...) hook so a target can supply default                                                                           DALEC_SYSEXT_* values for build_sysext.sh.DALEC_SYSEXT_* build args always override defaults.
- If DALEC_SYSEXT_OS_VERSION_ID is set and DALEC_SYSEXT_SYSEXT_LEVEL is not explicitly set, the default SYSEXT_LEVEL is dropped to avoid over-constraining extension matching.

          
**Default behavior for Flatcar sysext**
The Flatcar target provides defaults aligned with Flatcar expectations:

- DALEC_SYSEXT_OS_ID=flatcar
- default output image name is the spec name (produces <name>.raw)
- default DALEC_SYSEXT_SYSEXT_LEVEL=1.0 (unless OS_VERSION_ID is pinned without an explicit level)

**Tests**
Adds unit coverage for:

- filtering of DALEC_SYSEXT_* build args
- merging target defaults with build-arg overrides
- the OS_VERSION_ID vs SYSEXT_LEVEL precedence rule

Fixes: #969


targets/linux/flatcar/flatcar.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/project-dalec/dalec"
+	"github.com/project-dalec/dalec/frontend"
+	"github.com/project-dalec/dalec/targets/linux"
+	"github.com/project-dalec/dalec/targets/linux/deb/ubuntu"
+)
+
+const TargetKey = "flatcar"
+
+// Default SDK image. Keep it overrideable by code later if you want.
+const DefaultSDKImage = "ghcr.io/flatcar/flatcar-sdk-all:4593.0.0"
```
cpuguy83: Is all we need from this sdk mkfs.erofs (and other basic tools)?
This image is huge (almost 6GB).

cpuguy83 (commented):
What's the advantage of this change other than setting flatcar specific vars?

targets/linux/flatcar/flatcar.go
kartikjoshi21: @cpuguy83  good point. The Flatcar SDK image is not needed here since this path only needs `mkfs.erofs` and basic tooling, so I removed that dependency and now reuse the existing sysext worker.

The current value of this PR is adding a first-class `flatcar/testing/sysext` target with Flatcar-compatible defaults: output naming, `extension-release` metadata, and `DALEC_SYSEXT_*` override support.

I agree this is not full Flatcar-native build support yet. But we can have follow up PRs for full support.

cmd/worker-img-matrix/main.go
```diff
@@ -44,7 +45,15 @@ func main() {
 	}
 
 	var out []Info
+	workerlessTargets := map[string]struct{}{
+		// Flatcar currently exposes only a sysext assembly target.
+		flatcar.TargetKey: {},
+	}
 	for _, r := range plugins.Graph(filter) {
+		if _, ok := workerlessTargets[r.ID]; ok {
```
cpuguy83: The worker here appears to be ubuntu noble.
Let's use that for a worker target.

cpuguy83 (changes requested):

cmd/worker-img-matrix/main.go
kartikjoshi21: Done, thanks. Since the Flatcar sysext path is using the Noble worker underneath, I added a `flatcar/worker` target that delegates to the base Noble worker.

cpuguy83 (approved):

PR #980, opened 2026-03-02
test/testenv: run dial-stdio with Pdeathsig and process group

### LINKED ISSUE #974: [BUG] docker-buildx buildx dial-stdio --progress=plain processes leaks from integration tests when suite is interrupted
### Expected Behavior

`docker-buildx buildx dial-stdio --progress=plain` to not accumulate over time when test suite gets interrupted.

### Actual Behavior

When running `go test` right now and hitting CTRL+C while suite is running, orphan `docker-buildx buildx dial-stdio --progress=plain` process (es?) are left running and never gets cleaned up.

### Steps To Reproduce

_No response_

### Other

The following AI generated patch seems to be addressing this issue globally, so at least we don't have to make cleanup explicit. This should be robust enough when Go test binary gets SIGKILL'ed.

```diff
diff --git test/testenv/buildx.go test/testenv/buildx.go
index b099b61..5208731 100644
--- test/testenv/buildx.go
+++ test/testenv/buildx.go
@@ -13,6 +13,7 @@ import (
        "os/exec"
        "strings"
        "sync"
+       "syscall"
        "testing"
        "time"


@@ -105,6 +110,15 @@ func (b *BuildxEnv) dialStdio(ctx context.Context) error {
                // the buildx dial-stdio process from cleaning up its resources properly.
                cmd := exec.Command("docker", args...)
                cmd.Env = os.Environ()
+               cmd.SysProcAttr = &syscall.SysProcAttr{
+                       // Put the child in its own process group so we can kill the entire
+                       // group (docker + docker-buildx plugin) during cleanup.
+                       Setpgid: true,
+                       // Send SIGTERM to the child process when the parent (test process) dies.
+                       // This prevents dial-stdio processes from being orphaned when the test
+                       // suite is interrupted or crashes.
+                       Pdeathsig: syscall.SIGTERM,
+               }

### PR DESCRIPTION
**Pull request by invidian**
**What this PR does / why we need it**:

That should ensure we don't leave orphaned dial-stdio processes if the test suite is interrupted or crashes, and that all processes are cleaned up properly when the test finishes.

This approach is easier than explicit cleanup, since we delegate the cleanup to the OS.

**Which issue(s) this PR fixes** *(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*:
Closes #974

**Special notes for your reviewer**:

AI generated patch with manual testing. I am not familiar with nuances of `Pdeathsig` and `runtime.LockOSThread()`, but it seems to be addressing my issue.

### DIFF
--- a/test/testenv/buildx.go
+++ b/test/testenv/buildx.go
@@ -11,8 +11,10 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"runtime"
 	"strings"
 	"sync"
+	"syscall"
 	"testing"
 	"time"
 
@@ -104,6 +106,15 @@ func (b *BuildxEnv) dialStdio(ctx context.Context) error {
 		// the buildx dial-stdio process from cleaning up its resources properly.
 		cmd := exec.Command("docker", args...)
 		cmd.Env = os.Environ()
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			// Put the child in its own process group so we can kill the entire
+			// group (docker + docker-buildx plugin) during cleanup.
+			Setpgid: true,
+			// Send SIGTERM to the child process when the parent (test process) dies.
+			// This prevents dial-stdio processes from being orphaned when the test
+			// suite is interrupted or crashes.
+			Pdeathsig: syscall.SIGTERM,
+		}
 
 		c1, c2 := net.Pipe()
 		cmd.Stdin = c1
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: test/testenv/buildx.go [hunk 1]
```diff
@@ -117,16 +128,29 @@ func (b *BuildxEnv) dialStdio(ctx context.Context) error {
...
+			// Pdeathsig delivers the signal when the *thread* that forked the child exits,
+			// not when the process exits. Without locking, the Go runtime may destroy the
+			// thread that called cmd.Start(), prematurely delivering SIGTERM to the child.
+			runtime.LockOSThread()
+			defer runtime.UnlockOSThread()
+
+			if err := cmd.Start(); err != nil {
+				// Propagate the start error through the stderr pipe so the
+				// scanner below will surface it via scanner.Err().
+				w.CloseWithError(err)
```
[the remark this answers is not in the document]
invidian: The reason for pinning is to avoid the process getting killed when the thread might be terminated by the Go scheduler.

test/testenv/buildx.go
```diff
@@ -104,6 +106,15 @@ func (b *BuildxEnv) dialStdio(ctx context.Context) error {
...
 		cmd := exec.Command("docker", args...)
 		cmd.Env = os.Environ()
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			// Put the child in its own process group so we can kill the entire
+			// group (docker + docker-buildx plugin) during cleanup.
+			Setpgid: true,
+			// Send SIGTERM to the child process when the parent (test process) dies.
+			// This prevents dial-stdio processes from being orphaned when the test
+			// suite is interrupted or crashes.
+			Pdeathsig: syscall.SIGTERM,
```
cpuguy83, line 116: I realized this makes it so we can't invoke tests from non-Linux. We should at least gate this. But also definitely not a full solution.

invidian: I was aware, hence the revert :) https://github.com/project-dalec/dalec/pull/981

Issue #986, opened 2026-03-05
[REQ] Multiple image definitions


cpuguy83:
### What kind of request is this?

None

## Problem Statement

Dalec's current model is **one spec = one build = one package = one container image**. This creates three practical problems:

1. **Redundant builds**: When a single source tree produces multiple deliverables (e.g., a Bazel workspace that outputs standard, contrib, and debug binaries, or a Go monorepo with multiple commands), each deliverable requires its own spec with its own full build. The compilation is identical across specs — only artifact selection differs.

2. **No supplemental packages**: Linux packaging systems (RPM and deb) natively support subpackages — a single build producing multiple related packages (e.g., `foo`, `foo-devel`, `foo-dbg`). Dalec has no way to express this. (See [#607](https://github.com/project-dalec/dalec/issues/607))

---

## Current Spec Structure (Relevant Parts)

For reference, the parts of today's spec that would be affected:

```yaml
name: foo
version: 1.0.0
# ... metadata ...

sources: { ... }
build: { ... }

artifacts:         # Single artifact set → single package
  binaries: { ... }
  libs: { ... }

dependencies:
  build: { ... }
  runtime: { ... }

image:             # Single image config
  entrypoint: ...
  post:
    symlinks: { ... }

targets:           # Per-distro overrides (artifacts, image, deps, etc.)
  azlinux3:
    artifacts: { ... }
    image: { ... }
    dependencies: { ... }
  jammy:
    dependencies: { ... }
```

The `targets` map lets you override `artifacts`, `image`, `dependencies`, `conflicts`, `provides`, `replaces`, `package_config`, and `tests` per distro. But it cannot produce multiple packages — it's a customization mechanism, not a multiplexing mechanism.

---

## How RPM and Deb Handle Subpackages

Before proposing dalec changes, it's worth understanding how existing packaging systems solve this problem, since dalec ultimately produces RPMs and debs.

### RPM Subpackages

In an RPM spec file, subpackages are declared with `%package <name>` sections. The build is shared — there is one `%build` section and one `%install` section. Each subpackage gets its own `%files` section that selects which installed files go into that package.

Each subpackage can define its own `Summary`, `Requires`, `Provides`, `Conflicts`, `Obsoletes`, and `%description`. It inherits `Version`, `Release`, `License`, `Vendor`, and `URL` from the parent spec.

By default, the subpackage name is `<mainname>-<subname>`. The `-n` flag allows specifying an arbitrary name instead.

[...]

Issue #924, opened 2026-01-13
[REQ] declarative targets


cpuguy83:
### What kind of request is this?

New feature

### What is your request or suggestion?

First, an example Dockerfile:

```Dockerfile
FROM golang AS build
WORKDIr /src
COPY . .
RUN go build -o /out .

FROM scrach AS final
COPY --from=build /out /
```

In the above dockerfile, we can see there are 2 targets: `build` and `final`.
You can pass these directly to `docker build` via the `--target` flag.

**Note**: let's clarify some terms here before proceeding:

- Docker target - The actual target passed to `docker build` and also returned by the `frontend.targets` API call.
- Distro target -  The OS distribution reference used inside dalec, e.g. `jammy`, `windowscross`

A "Docker target" is, in effect, an API endpoint, from here on referred to as `BT`.
A"Distro target" is the underlying implementation that a BT request is sent to, from here on referred to as `DT`.

Dalec takes a different approach where the frontend just supports a static set of BT's.
There is no way from looking at a spec to determine what BT's are actually supported.

e.g.

```yaml
name: foo
sources:
  src: ...

build:
  steps: ...

artifacts: ...
```

This is a minimal and nonsense example, but the issue is the same, or potentially worse, with more complex specs.
If you ask dalec what BT's are available (assuming I wrote a valid spec above) you'd always get a static list, regardless of the content of the spec, but dependent on the version of dalec used.
Example list here:

https://github.com/project-dalec/dalec/blob/4b293ffb19c82b0942268c5ebec0f4cc432a557f/website/docs/examples/targets.md?plain=1#L1-L71

This is a huge list, that only gets bigger and likely is like to not technically valid, in terms of actually being build-able, unless the spec author was deliberate about it.

We do have the `targets` section which lets spec authors define DT customizations, but these are just customizations not filtering down the list of BT's. It is perfectly valid to have some things defined in the root of the spec and some things not.

There is also no way to set a default BT, i.e. what to build if no BT is provided to `docker build --target=<BT>`.
In the Dockerfile format the default BT is the last build stage in the file (e.g. `final` in the above Dockerfile)

Historically dalec used to look if there is anything defined under `targets` and used that to filter the list of available BT's, but this was extremely confusing and, as mentioned above, not really inline with how the spec is parsed out.

Internally we have built tooling that adds an extension field to the spec which declares the BT's we are going to build, which our internal tooling parses and generates the appropriate build matrix for our CI:

```yaml
x-build-extensions:
[...]

PR #987, opened 2026-03-06
Refactor router

cpuguy83:
This is in preparation of #986 
It also just generally fixes issues with routing for routes with sub-routers (e.g. see that targets.md now populates certain debug routes).

It moves to using a single, flat router for all routes.
Route registration happens at request time since available routes can be spec-dependent.
For the targets json API (docker build --call frontend.targets,format=json) it now includes a field to give extra hints to clients, such as the vscode extension, on if the targets are explicit in the spec or not.


internal/gwutil/gwutil.go
```diff
@@ -0,0 +1,82 @@
...
+// and [SpecLoader] interfaces are not lost when the client is wrapped.
+//
+// If either the wrapper or the inner client implements [SpecLoader], that
+// interface is also preserved on the returned client (wrapper is preferred
+// over inner).
+func WithCurrentFrontend(inner gwclient.Client, wrapper gwclient.Client) gwclient.Client {
+	cf, ok := inner.(CurrentFrontend)
+	if !ok {
+		return wrapper
+	}
```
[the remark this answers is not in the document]
cpuguy83: Updated the comment.

frontend/router_test.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	expectedRealCount++
+
+	if count := realCount(); count != expectedRealCount {
+		t.Errorf("expected real handler call count to be %d, got %d", expectedRealCount, count)
+	}
+
+	// Register a sub-route at a fully qualified path.
```
invidian: Could those tests benefit from sub-tests to reduce the number of assertions and shared state between the test cases to drive down the complexity?

internal/frontendapi/router.go
```diff
@@ -10,37 +11,53 @@ import (
 	_ "github.com/project-dalec/dalec/targets/plugin"
 )
 
-func NewBuildRouter(ctx context.Context) (*frontend.BuildMux, error) {
-	var mux frontend.BuildMux
-	mux.Add(debug.DebugRoute, debug.Handle, nil)
+// NewRouter creates a flat Router with all routes registered eagerly.
```
invidian: eagerly? What does it mean in this context?

targets/plugin/init.go
```diff
@@ -11,23 +12,49 @@ import (
 )
 
 func init() {
```
invidian, line 18: While we're at it, if we use plugins only for building routes and we already have some statically defined routes, I wonder if we could load those explicitly in `internal/frontendapi/router.go` or somewhere along the way in the composition tree, while I guess still keeping logic for loading from plugins from some external sources (although perhaps exposing Dalec as a library would still be better so wrappers can do their own transparent composition?)

That might be out of scope of course.

frontend/mux.go
```diff
@@ -1,683 +0,0 @@
-package frontend
```
invidian, line 1: nit: split of mux.go into client.go and router.go could be done in a separate commit so changes in those files are better represented, otherwise it's hard to follow what has changed since changes are done in separate files now.

frontend/router.go
```diff
@@ -0,0 +1,496 @@
...
+	var ls TargetList
+
+	keys := maps.Keys(r.routes)
+	slices.Sort(keys)
+
+	for _, key := range keys {
+		route := r.routes[key]
+
+		// If a filter target is set, only include routes under that prefix.
+		if target != "" && key != target && !strings.HasPrefix(key, target+"/") {
```
invidian, line 226: Is this related to #924?

frontend/router.go
```diff
@@ -0,0 +1,496 @@
...
+
+		// Hidden routes are dispatchable but excluded from the target list.
+		if route.Info.Hidden {
+			continue
+		}
+
+		dt := route.Info
+		if hasSpecTargets {
+			tlk := topLevelKey(key)
+			if _, ok := spec.Targets[tlk]; ok {
```
invidian: nit: maybe `spec.Targets[topLevelKey(key)]` to avoid temporary variable naming?

Also conditional can be avoided:
```go
_, ok := spec.Targets[topLevelKey(key)]
dt.SpecDefined = ok
```

frontend/router.go
```diff
@@ -0,0 +1,496 @@
...
+				dt.SpecDefined = true
+			}
+		}
+
+		// Forwarding route: query the remote frontend for its sub-targets.
```
invidian, line 235: nit: maybe:
```go
targets := []Target{dt}

		if route.Forward != nil {
			subTargets, err := queryForwardedTargets(ctx, client, key, route.Forward)
			if err != nil {
				bklog.G(ctx).WithError(err).Warn("Could not query forwarded frontend targets")
			} else {
		  	targets = subTargets
      }
		}

		ls.Targets = append(ls.Targets, targets...)
```

That yields simpler flow I think, we only write once and we don't short-circut with continue.

frontend/router.go
```diff
@@ -0,0 +1,473 @@
...
+
+	var remote bktargets.List
+	if err := json.Unmarshal(dt, &remote); err != nil {
+		return nil, errors.Wrap(err, "error unmarshalling forwarded target list")
+	}
+
+	targets := make([]Target, 0, len(remote.Targets))
+	for _, t := range remote.Targets {
+		t.Name = path.Join(prefix, t.Name)
+		targets = append(targets, Target{Target: t})
```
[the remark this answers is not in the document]
invidian: Or pass if spec is defined from the list()?

invidian (commented):
Comments for `frontend: replace BuildMux with flat Router`.

targets/windows/handler.go
```diff
@@ -43,7 +43,15 @@ var distroConfig = &distro.Config{
...
 
 // Routes returns the flat routes for the Windows target, prefixed with the given prefix.
-func Routes(prefix string) []frontend.Route {
+func Routes(prefix string, ctx context.Context, client gwclient.Client) ([]frontend.Route, error) {
+	spec, err := frontend.LoadSpecFromClient(ctx, client)
+	if err != nil {
+		return nil, err
+	}
+
+	_, specDefined := spec.Targets[prefix]
```
invidian, line 47: Couldn't we set the spec defined by mutating the routes received from different targets to avoid having each target load the spec explicitly?

internal/frontendapi/router.go
```diff
@@ -55,7 +56,11 @@ func loadRouteProviders(ctx context.Context, r *frontend.Router) error {
 		if !ok {
 			return fmt.Errorf("plugin %T does not implement RouteProvider", v)
 		}
-		for _, route := range provider.Routes() {
+		routes, err := provider.Routes(ctx, client)
+		if err != nil {
+			return err
```
invidian: nit: would be nice to annotate errors with `fmt.Errorf` for new code, that helps tracing down errors a lot when they occur.

targets/windows/handler.go
invidian: Oh, it was done centrally first, right. Could you elaborate why is it better to do it per routes source, considering that requires a client, which is a god dependency pretty much?

invidian (commented):
For `frontend: make route registration spec-aware`

internal/gwutil/gwutil.go
```diff
@@ -0,0 +1,83 @@
...
+
+	w := &clientWithCurrentFrontend{Client: wrapper, cf: cf}
+
+	// Preserve SpecLoader if present on either wrapper or inner.
+	// Prefer wrapper (it's the more specific layer).
+	switch {
```
invidian: nit: I think you should be able to do something like:
```go
	switch v := i.(type) {
	case int:
		fmt.Printf("Twice %v is %v\n", v, v*2)
	case string:
		fmt.Printf("%q is %v bytes long\n", v, len(v))
	default:
		fmt.Printf("I don't know about type %T!\n", v)
	}
```

targets/windows/handler.go
cpuguy83: I did this for simplicity while moving everything around and figuring out what the shape is.
I figured we can come back to it easily enough, and did slightly with the cached client wrapper in a later commit.

internal/gwutil/gwutil.go
cpuguy83: These is switching on two different variables. Probably shouldn't even bother with the switch here, I think.

cpuguy83: Its weird I thought I did this totally differently, but maybe messed up while re-arranging commits.
The solution I'd come up with is to just have `clientWIthCurrentFrontend` always implement SpecLoader since the LoadSpecFromClient function will handle the check anyway.

cpuguy83: Oh right... import cycle issues.

cmd/worker-img-matrix/main.go
```diff
@@ -40,14 +47,38 @@ func main() {
...
+		v, err := p.Instance()
+		if err != nil {
+			panic(fmt.Errorf("failed to get instance for plugin %s: %w", reg.ID, err))
+		}
+
+		provider := v.(plugins.RouteProvider)
+		routes, err := provider.Routes(ctx, client)
+		if err != nil {
+			panic(fmt.Errorf("failed to get routes for plugin %s: %w", reg.ID, err))
+		}
```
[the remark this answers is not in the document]
cpuguy83: Panic is the right thing to do here. It would be a bug that the user can do nothing about.

frontend/router.go
cpuguy83: No this is preserving existing hanlding:

```console
❯ docker buildx  build --build-arg BUILDKIT_SYNTAX=ghcr.io/project-dalec/dalec/frontend:latest --call targets --target azlinux3 -<<<"{}"
[+] Building 0.5s (3/3) FINISHED                                                                                                                                                     docker:default
 => resolve image config for docker-image://ghcr.io/project-dalec/dalec/frontend:latest                                                                                                        0.4s
 => CACHED docker-image://ghcr.io/project-dalec/dalec/frontend:latest@sha256:8fef7a51dfb9852b97d53ca35ba6eb6828ab5a1bb7ae7709ba6ec23032597bad                                                  0.0s
 => => resolve ghcr.io/project-dalec/dalec/frontend:latest@sha256:8fef7a51dfb9852b97d53ca35ba6eb6828ab5a1bb7ae7709ba6ec23032597bad                                                             0.0s
 => [internal] load build definition from Dockerfile                                                                                                                                           0.0s
 => => transferring dockerfile: 40B                                                                                                                                                            0.0s
TARGET                       DESCRIPTION
azlinux3/container (default) Builds a container image for Azure Linux 3
azlinux3/container/depsonly  Builds a container image with only the runtime dependencies installed.
azlinux3/rpm                 Builds an rpm and src.rpm.
azlinux3/rpm/debug           Debug options for rpm builds.
azlinux3/testing/sysext      Builds a systemd system extension image.
azlinux3/worker              Builds the base worker image responsible for building the rpm
```

cmd/frontend/main.go
```diff
@@ -65,12 +67,14 @@ func main() {
...
-		bklog.L.WithError(err).Fatal("error creating frontend router")
-	}
 
-	if err := grpcclient.RunFromEnvironment(ctx, mux.Handler(frontend.WithTargetForwardingHandler)); err != nil {
+	if err := grpcclient.RunFromEnvironment(ctx, func(ctx context.Context, client gwclient.Client) (*gwclient.Result, error) {
+		r, err := frontendapi.NewRouter(ctx, client)
+		if err != nil {
+			return nil, err
+		}
+		return r.Handler(frontend.WithTargetForwardingHandler)(ctx, client)
```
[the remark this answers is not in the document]
cpuguy83: This is intentional. The spec used for the routes doesn't have all the context it needs for loading build args (at least historically).
So the routes get the spec without the validation that a build arg was provided but unconsumed.

internal/gwutil/gwutil.go
cpuguy83: This is no longer relevant.

internal/frontendapi/router.go
cpuguy83: Yep, done.

frontend/router.go
cpuguy83: This code is rewritten.

frontend/router.go
cpuguy83: This now ranges over spec.Targets.

frontend/mux.go
cpuguy83: Good point, didn't think about this.
Done.

internal/frontendapi/router.go
cpuguy83: Reworded this, the comment was not useful.

frontend/router_test.go
cpuguy83: Done.

frontend/client.go
DannyBrito, line 1: nit: should result.txt and result.json be constants?

frontend/router.go
DannyBrito, line 1: nit consider using constants for result.txt and result.json?

frontend/router.go
```diff
@@ -0,0 +1,469 @@
...
+func (r *Router) lookupTarget(ctx context.Context, target string) (matchedPath string, _ *Route, _ error) {
+	// 1. Exact match
+	if route, ok := r.routes[target]; ok {
+		return target, &route, nil
+	}
+
+	// 2. Empty target — no global default; prompt the user.
+	if target == "" {
+		return "", nil, handlerNotFound(target, maps.Keys(r.routes))
+	}
```
DannyBrito, line 350: should the target == "" check be before the exact match? in case there is a empty target in the routes?
or maybe add a guard against empty target in Add function?

cpuguy83: Done.

frontend/router.go
cpuguy83: Done

frontend/client.go
cpuguy83: Done

cpuguy83:
Rebased.

DannyBrito (approved):

PR #988, opened 2026-03-06
Supplemental packages

### LINKED ISSUE #607: [REQ] Supplemental packages
# Problem Statement

Dalec's current model is **one spec = one build = one package = one container image**. This creates three practical problems:

1. **Redundant builds**: When a single source tree produces multiple deliverables (e.g., a Bazel workspace that outputs standard, contrib, and debug binaries, or a Go monorepo with multiple commands), each deliverable requires its own spec with its own full build. The compilation is identical across specs — only artifact selection differs.

2. **No supplemental packages**: Linux packaging systems (RPM and deb) natively support subpackages — a single build producing multiple related packages (e.g., `foo`, `foo-devel`, `foo-dbg`). Dalec has no way to express this. (See [#607](https://github.com/project-dalec/dalec/issues/607))

---

## Current Spec Structure (Relevant Parts)

For reference, the parts of today's spec that would be affected:

```yaml
name: foo
version: 1.0.0
# ... metadata ...

sources: { ... }
build: { ... }

artifacts:         # Single artifact set → single package
  binaries: { ... }
  libs: { ... }

dependencies:
  build: { ... }
  runtime: { ... }

image:             # Single image config
  entrypoint: ...
  post:
    symlinks: { ... }

targets:           # Per-distro overrides (artifacts, image, deps, etc.)
  azlinux3:
    artifacts: { ... }
    image: { ... }
    dependencies: { ... }
  jammy:
    dependencies: { ... }
```

The `targets` map lets you override `artifacts`, `image`, `dependencies`, `conflicts`, `provides`, `replaces`, `package_config`, and `tests` per distro. But it cannot produce multiple packages — it's a customization mechanism, not a multiplexing mechanism.

---

## How RPM and Deb Handle Subpackages

Before proposing dalec changes, it's worth understanding how existing packaging systems solve this problem, since dalec ultimately produces RPMs and debs.

### RPM Subpackages

In an RPM spec file, subpackages are declared with `%package <name>` sections. The build is shared — there is one `%build` section and one `%install` section. Each subpackage gets its own `%files` section that selects which installed files go into that package.

Each subpackage can define its own `Summary`, `Requires`, `Provides`, `Conflicts`, `Obsoletes`, and `%description`. It inherits `Version`, `Release`, `License`, `Vendor`, and `URL` from the parent spec.

By default, the subpackage name is `<mainname>-<subname>`. The `-n` flag allows specifying an arbitrary name instead.

**There is no implicit dependency between subpackages and the main package.** If `foo-devel` needs `foo`, it must explicitly declare `Requires: foo = %{version}-%{release}`.

### Debian Binary Packages

In `debian/control`, the first stanza defines the source package (with `Build-Depends`). Each subsequent stanza defines a binary package with its own `Package`, `Architecture`, `Description`, `Depends`, `Conflicts`, `Provides`, etc.

The build runs once via `debian/rules`. The `debian/<package>.install` files determine which built files go into which binary package.

**There is no implicit dependency between binary packages from the same source.** If `libfoo-dev` needs `libfoo0`, it must declare `Depends: libfoo0 (= ${binary:Version})`.

### Why No Implicit Dependencies

This is intentional in both systems. Different subpackage relationships require different dependency declarations:

- `foo-devel` → **depends on** `foo` (needs the runtime libraries to be useful)
- `foo-docs` → **no dependency** on `foo` (documentation is standalone)
- `foo-debug` → **depends on** `foo` (debug symbols for the installed binary)
- `foo-contrib` → **conflicts with** `foo` (both provide `/usr/bin/foo`; only one can be installed)

Implicit dependencies would be wrong for at least some of these cases. Dalec should follow the same pattern: **supplemental packages declare their own dependencies explicitly**.

---

### Core Idea

Add a `packages` map inside `targets.<distro>` for defining **named supplemental packages** that share the spec's build output but have their own artifact selection, metadata, and dependency declarations.

The primary package remains as-is (defined by the existing top-level `artifacts`, `dependencies`, etc.). Supplemental packages are additional packages produced from the same build.

**There is no root-level `packages` map.** Supplemental packages exist only within `targets.<distro>.packages`. Each target's package definitions are self-contained. Sharing across targets is handled by YAML anchors — a standard YAML feature that requires no dalec-specific inheritance logic.

### What Each Supplemental Package Can Define

Following the RPM/deb model, each supplemental package needs:

- **artifacts** — its own selection of binaries, libs, config files, symlinks, etc. from the shared build output
- **description** — its own package description
- **dependencies.runtime** — its own runtime dependencies (can reference the primary package, other supplemental packages, or external packages)
- **conflicts** — packages it conflicts with (e.g., if a contrib variant replaces the standard binary)
- **provides** — virtual packages or capabilities it provides
- **replaces** — packages it replaces/obsoletes

What it inherits from the parent spec and cannot override:

- **version / revision** — must match the parent
- **license / vendor / website** — must match the parent
- **sources / build / patches** — the whole point is sharing the build
- **dependencies.build** — build deps are for the shared build

### Package Naming

Supplemental package names default to `<parent>-<key>`, where `<key>` is the map key under `packages`. For example, a `debug` entry in a spec named `foo` produces a package called `foo-debug`. This matches the RPM subpackage convention and common Debian practice.

An explicit `name` field on a supplemental package overrides this default, allowing fully custom names. Both packaging systems support this — RPM via `%package -n <name>`, and Debian via the `Package` field in `debian/control` being independent per binary package stanza.

```yaml
targets:
  azlinux3:
    packages:
      debug:
        # name defaults to foo-debug (parent-key)
        description: "Debug symbols for foo"

      compat:
        name: foo-compat-v2    # explicit override
        description: "Backward compatibility shim"
```

### Required Fields

Each supplemental package must specify a **description**. This is required by both RPM (`Summary` + `%description`) and Debian (`Description`) for every subpackage/binary package. Dalec validates this at spec parse time.

### Validation

Dalec validates supplemental package names at spec parse time:

- **No duplicate names within a target.** If two supplemental packages in the same target resolve to the same package name (after applying `name` overrides and `<parent>-<key>` defaults), dalec errors.
- **No conflict with the primary package.** If a supplemental package's resolved name equals the spec's `name` field, dalec errors.

### Dependency Cycles

Dalec does **not** validate runtime dependency cycles between supplemental packages. Supplemental packages can depend on each other freely, and cycle detection (if it matters) is left to the package manager at install time. This matches RPM and deb behavior — neither rpmbuild nor dpkg-buildpackage validates runtime dependency graphs at build time.

### Changelog

There is one changelog per spec, shared by all packages (primary and supplemental). This matches RPM behavior (subpackages share the parent's changelog) and Debian behavior (binary packages each get the source package changelog). Supplemental packages do not get their own changelogs.

### No Artifact Inheritance

Supplemental packages do **not** inherit any artifacts from the primary package. Each supplemental package's `artifacts` block is self-contained — it specifies exactly what goes into that package and nothing else.

This means common artifacts like license files must be repeated in each supplemental package that needs them. While this creates some repetition, it follows the principle that **explicit is better than implicit**:

- RPM and deb both require explicit `%files` / `.install` lists per subpackage — nothing is inherited from the main package.
- Implicit inheritance creates ambiguity about which artifacts are "common" vs. package-specific.
- It avoids surprising behavior when a spec author adds a license artifact to the primary package and it silently appears in all supplemental packages.
- The repetition is bounded — artifact blocks are small, and most supplemental packages only have a few entries.

### Where It Goes in the Spec

Supplemental packages are defined inside `targets.<distro>.packages`, not at the root level. Each target independently declares which supplemental packages it produces and how they are configured:

```yaml
# YAML anchors for sharing package definitions across targets
x-packages:
  debug: &debug-pkg
    description: "Debug symbols for foo"
    artifacts:
      binaries:
        dbg/foo: { name: foo.dbg }
        dbg/foo.dwp: {}
    dependencies:
      runtime:
        foo: {}              # explicit dep on main package

  contrib: &contrib-pkg
    description: "Foo with contrib extensions"
    artifacts:
      binaries:
        foo-contrib: { name: foo }
    conflicts:
      foo: {}                # can't install alongside standard
    provides:
      foo: {}                # satisfies deps on foo
    dependencies:
      runtime:
        openssl-libs: {}

targets:
  azlinux3:
    packages:
      debug: *debug-pkg
      contrib: *contrib-pkg
  jammy:
    packages:
      debug: *debug-pkg
      contrib: *contrib-pkg
```

The `x-packages` block (or any `x-` prefixed key) is ignored by dalec — it's purely a YAML anchor container. The actual package definitions live in `targets.<distro>.packages`. YAML anchors (`&name` / `*name`) handle sharing with zero dalec-specific logic.

### Why Not Root-Level `packages`?

An earlier version of this proposal defined supplemental packages at the root level with per-target overrides, mirroring how the primary package works. This was abandoned in favor of the targets-only model for several reasons:

1. **No override chain to reason about.** Root-level packages required per-field override semantics (`packages.<name>` → `targets.<distro>.packages.<name>`), a `skip` mechanism for excluding packages from targets, and rules about which fields merge vs. replace. Targets-only eliminates all of this — each target's package definition is exactly what gets built.

2. **YAML anchors are sufficient.** The only thing root-level definitions provided was sharing across targets. YAML anchors do the same thing without any dalec-specific inheritance logic. When a target needs a variation, you just write the variation directly instead of crafting an override that interacts with the root definition.

3. **Exclusion is implicit.** With root-level packages, a `skip: true` mechanism was needed to prevent a package from being built for a target. With targets-only, a package simply doesn't exist in targets that don't define it. No mechanism needed.

4. **Consistency with the primary package is unnecessary.** The primary package uses root-level fields + target overrides because it already works that way and backward compatibility requires preserving that behavior. Supplemental packages are a new feature with no backward compatibility constraint — they can use a simpler model.

### Interaction with `<distro>/rpm` and `<distro>/deb`

For package targets, the natural behavior matches how rpmbuild and dpkg-buildpackage work: **a single invocation produces all packages** (primary + supplemental). The output is a directory of `.rpm` or `.deb` files (or `.zip` files for `windowscross`).

There is no sub-path selection for package targets — `<distro>/rpm` always produces all packages defined for that target. There is no `<distro>/rpm/<name>` path to build an individual supplemental package. This matches how rpmbuild works: you build the whole spec and get all subpackages.

Only the packages defined in `targets.<distro>.packages` are produced for that target. If a target doesn't define a supplemental package, it simply isn't built for that target — no exclusion mechanism needed.

### Empty Artifacts on Windows

For `windowscross/zip` targets, dalec errors if a package (primary or supplemental) resolves to no artifacts. Zip files must contain something — an empty zip is meaningless.

This does not apply to Linux package targets (`rpm`, `deb`), where a package with no artifacts is perfectly valid. Metapackages that only declare runtime dependencies, `provides`, or `conflicts` are a standard pattern in both RPM and deb.

### Sysext Targets

Sysext targets (`<distro>/testing/sysext`) are still experimental and their interaction with supplemental packages is not defined by this proposal. The sysext model may end up being similar to containers (with a `packages` field for selection), but this is deferred until the sysext design stabilizes.

cpuguy83:
Adds **subpackages** (supplemental packages): a single build can produce multiple installable packages, defined per target under a `packages` field. Each subpackage shares the primary's build (steps, sources, version, license, …) and selects its own artifacts.

- **RPM**: emits per-subpackage `%package`/`%files` sections.
- **Debian**: emits extra `Package:` stanzas with per-package install/maintainer scripts and `dh_installsystemd` overrides.
- **windowscross**: one zip per package; the container target installs binaries from all packages.

Artifacts are self-contained — anything placed in a subpackage isn't also shipped by the primary. Install-time scriptlet requirements (systemd, users/groups) are derived per subpackage.

Closes #607

### DIFF
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -2198,6 +2198,74 @@
 			],
 			"description": "Spec is the specification for a package build."
 		},
+		"SubPackage": {
+			"required": [
+				"description"
+			],
+			"properties": {
+				"artifacts": {
+					"$ref": "#/$defs/Artifacts",
+					"description": "Artifacts specifies which build outputs go into this supplemental package.\nThis is self-contained — no artifacts are inherited from the primary package."
+				},
+				"conflicts": {
+					"$ref": "#/$defs/PackageDependencyList",
+					"description": "Conflicts is the list of packages that conflict with this supplemental package."
+				},
+				"dependencies": {
+					"$ref": "#/$defs/SubPackageDependencies",
+					"description": "Dependencies specifies runtime dependencies for this supplemental package.\nOnly runtime and recommends dependencies are allowed; build dependencies\nare shared with the primary package."
+				},
+				"description": {
+					"type": [
+						"string"
+					],
+					"description": "Description is the package description. This is required — both RPM and\nDebian require a description/summary for every subpackage."
+				},
+				"name": {
+					"type": [
+						"string",
+						"null"
+					],
+					"description": "Name overrides the default package name.\nBy default, the package name is \"\u003cparent\u003e-\u003ckey\u003e\" where \u003ckey\u003e is the map\nkey under which this SubPackage is defined. Set this to use a fully custom\npackage name instead."
+				},
+				"provides": {
+					"$ref": "#/$defs/PackageDependencyList",
+					"description": "Provides is the list of things this supplemental package provides."
+				},
+				"replaces": {
+					"$ref": "#/$defs/PackageDependencyList",
+					"description": "Replaces is the list of packages that this supplemental package replaces/obsoletes."
+				}
+			},
+			"additionalProperties": {
+				"not": {}
+			},
+			"type": [
+				"object",
+				"null"
+			],
+			"description": "SubPackage defines a supplemental package produced from the same build output as the primary package."
+		},
+		"SubPackageDependencies": {
+			"properties": {
+				"recommends": {
+					"$ref": "#/$defs/PackageDependencyList",
+					"description": "Recommends is the list of packages recommended to install with the supplemental package.\nNote: Not all package managers support this (e.g. rpm)"
+				},
+				"runtime": {
+					"$ref": "#/$defs/PackageDependencyList",
+					"description": "Runtime is the list of packages required to install/run the supplemental package."
+				}
+			},
+			"additionalProperties": {
+				"not": {}
+			},
+			"type": [
+				"object",
+				"null"
+			],
+			"description": "SubPackageDependencies contains only the dependency fields valid for supplemental packages."
+		},
 		"SymlinkTarget": {
 			"required": [
 				"path",
@@ -2374,6 +2442,16 @@
 				"package_config": {
 					"$ref": "#/$defs/PackageConfig",
 					"description": "PackageConfig is the configuration to use for artifact targets, such as\nrpms, debs, or zip files containing Windows binaries"
+				},
+				"packages": {
+					"additionalProperties": {
+						"$ref": "#/$defs/SubPackage"
+					},
+					"type": [
+						"object",
+						"null"
+					],
+					"description": "Packages defines supplemental packages produced from the same build output\nas the primary package. Each entry produces an additional package with its\nown artifact selection, runtime dependencies, and metadata.\nThe map key is used as a suffix for the default package name (\"\u003cspecName\u003e-\u003ckey\u003e\")."
 				},
 				"provides": {
 					"$ref": "#/$defs/PackageDependencyList",
--- a/load.go
+++ b/load.go
@@ -577,6 +577,9 @@ func (s Spec) Validate() error {
 		if err := t.validate(); err != nil {
 			errs = append(errs, errors.Wrapf(err, "target %s", k))
 		}
+		if err := validateSubPackageNames(s.Name, k, t.Packages); err != nil {
+			errs = append(errs, err)
+		}
 	}
 
 	if err := s.Image.validate(); err != nil {
--- a/packaging/linux/deb/debroot.go
+++ b/packaging/linux/deb/debroot.go
@@ -154,10 +154,10 @@ func Debroot(ctx context.Context, sOpt dalec.SourceOpts, spec *dalec.Spec, worke
 		return dalec.ErrorState(in, errors.Wrap(err, "error generating custom systemd postinst"))
 	}
 
-	if len(customEnable) > 0 {
+	for _, partial := range customEnable {
 		// This is not meant to be executed on its own and will instead get added
 		// to a post inst file, so need to mark this as executable.
-		states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/"+customSystemdPostinstFile), 0o600, customEnable), opts...))
+		states = append(states, dalecDir.File(llb.Mkfile(filepath.Join(dir, "dalec/"+partial.partialFile()), 0o600, partial.content), opts...))
 	}
 
 	postinst := generatePostinst(spec, target)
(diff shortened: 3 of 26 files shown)


cpuguy83:
Opened separately as a PR stack. See #1149

PR #991, opened 2026-03-11
frontend: prefer dnf over tdnf to work around tdnf GPG and forcearch limitations

cpuguy83:
tdnf fails when installing signed local RPMs (from the cmdline virtual repo) into an installroot with a populated RPM database, because it requires a gpgkey entry for cmdline which is a synthetic repo with no config. This manifests when building containers with a custom base image on azlinux/mariner distros.

Rather than working around individual tdnf bugs, prefer dnf when it is available. The install script now checks for dnf at runtime and switches from tdnf transparently. The same-platform worker bootstrap is updated to install dnf as a separate first step (mirroring the cross-arch path) so that subsequent installs benefit from dnf.

This also exposed a couple of other issues:

1. We weren't testing depsonly anywhere except in the basic "e2e" tests in docker-bake.hcl which is just a holdover from before we had integration tests. I've migrated the tests there and made it run on all distros that support this target (rpm distros only).
2. Instead of manually pulling in all packages into a dir (and incurring extra copying), just pass the list of packages to install to a new method for handling this scenario.

---

This fixes a real issue we saw in our builds where azlinux3/container is used but a custom base image is set instead of building one from scratch (as is the default).

### DIFF
--- a/targets/linux/rpm/distro/container.go
+++ b/targets/linux/rpm/distro/container.go
@@ -95,16 +95,36 @@ func (cfg *Config) HandleDepsOnly(ctx context.Context, client gwclient.Client) (
 		}
 
 		pc := dalec.Platform(platform)
-		worker := cfg.Worker(sOpt, pg, pc)
 
-		deps := dalec.SortMapKeys(rtDeps)
+		// NOTE: Deps-only allows bare specs, ie specs with just the runtime deps included.
+		// This means we may need to fill in some of the details that are required by the package manager.
+		depsSpec := &dalec.Spec{
+			Name:        spec.Name + "-runtime-deps",
+			License:     spec.License,
+			Version:     spec.Version,
+			Revision:    spec.Revision,
+			Description: "Runtime dependencies meta package",
+			Dependencies: &dalec.PackageDependencies{
+				Runtime: rtDeps,
+			},
+		}
 
-		withDownloads := worker.Run(dalec.ShArgs("set -ex; mkdir -p /tmp/rpms/RPMS/$(uname -m)")).
-			Run(cfg.Install(deps,
-				DnfDownloadAllDeps("/tmp/rpms/RPMS/$(uname -m)")), pg).Root()
-		rpmDir := llb.Scratch().File(llb.Copy(withDownloads, "/tmp/rpms", "/", dalec.WithDirContentsOnly()), pg)
+		if depsSpec.Name == "-runtime-deps" {
+			// Name cannot start with "-"
+			depsSpec.Name = "dalec-user" + depsSpec.Name
+		}
+		if depsSpec.Version == "" {
+			depsSpec.Version = "0.0.1"
+		}
+		if depsSpec.Revision == "" {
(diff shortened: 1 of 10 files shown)


test/linux_target_test.go
```diff
@@ -177,6 +179,16 @@ func testLinuxDistro(ctx context.Context, t *testing.T, testConfig testLinuxConf
 		})
 	})
 
+	t.Run("container/depsonly", func(t *testing.T) {
```
invidian: I think we already have `t.Run("container", ...)` here, so that should go under that test hierarchy, otherwise it's confusing.

targets/linux/rpm/distro/container.go
```diff
@@ -80,6 +80,67 @@ func (cfg *Config) BuildContainer(ctx context.Context, client gwclient.Client, s
...
+
+	const workPath = "/tmp/rootfs"
+
+	bi, err := spec.GetSingleBase(targetKey)
+	if err != nil {
+		return dalec.ErrorState(llb.Scratch(), err)
+	}
+
+	skipBase := bi != nil
+	rootfs := bi.ToState(sOpt, opts...)
```
invidian: this will panic if `bi == nil`?

targets/linux/rpm/distro/container.go
```diff
@@ -80,6 +80,67 @@ func (cfg *Config) BuildContainer(ctx context.Context, client gwclient.Client, s
 	return rootfs
 }
 
+// BuildContainerWithPackages builds a container image by installing the given
+// package names (resolved from repos) instead of local RPM files.
+// This is used by the deps-only target where there are no locally-built RPMs.
+func (cfg *Config) BuildContainerWithPackages(ctx context.Context, client gwclient.Client, sOpt dalec.SourceOpts, spec *dalec.Spec, targetKey string, packages []string, opts ...llb.ConstraintsOpt) llb.State {
```
invidian: nit: that could get much more thorough test suite as we have for building deb containers for example, with cases like:
- respecting cache keys
- no base image in spec
- unit tests for base packages
- symlinks handling

Or maybe deps-only should be simply passing whole container suite like it would be reusable in #957?

invidian (changes requested):

targets/linux/rpm/distro/container.go
cpuguy83: I was trying to avoid using the package infra because depsonly originally avoided it as well.
Given the test-ability concerns and even things like package constraints on runtime deps, I went ahead and just made HandleDepsOnly just be effectively a wrapper for BuildPkg->BuildContainer now.

targets/linux/rpm/distro/container.go
cpuguy83: This is gone now, but `bi.ToState` checks for nil internally and returns scratch if nil.

test/linux_target_test.go
cpuguy83: Fixed.

invidian (approved):

PR #998, opened 2026-03-24
fix: credential helper fallback for git < 2.46

cpuguy83:
Git versions before 2.46 (including Azure Linux 3's 2.45.4 and Ubuntu Noble's 2.43) do not announce capability[]=authtype. The credential helper was unconditionally using authtype/credential (v2 protocol), which git silently discarded, causing authentication failures.

Detect whether git announced the authtype capability and fall back to username/password for pre-2.46 git. For header auth with non-Basic types (e.g., Bearer), return an error since there is no v1 equivalent.

Also fix readPayload to handle blank lines (input termination) and unknown keys (silent discard) per the git credential protocol spec.

### DIFF
--- a/cmd/frontend/git_credential_helper.go
+++ b/cmd/frontend/git_credential_helper.go
@@ -245,7 +245,7 @@ func generateResponse(payload *gitPayload, secret []byte, kind string) (string, 
 }
 
 func handleSecretHeader(b []byte, payload *gitPayload) (string, error) {
-	s := string(b)
+	s := strings.TrimRight(string(b), "\r\n")
 	authtype, credential, ok := strings.Cut(s, " ")
 	if !ok {
 		return "", fmt.Errorf("improperly formatted auth header")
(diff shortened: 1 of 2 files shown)


cmd/frontend/git_credential_helper.go
```diff
@@ -142,10 +143,17 @@ func readPayload(r io.Reader) gitPayload {
...
+		if line == "" {
+			break
+		}
+
+		k, v, ok := strings.Cut(line, "=")
 		if !ok {
-			exit1("improper payload from git")
+			// Per the git credential protocol, lines without '=' are malformed.
+			// Skip them gracefully rather than crashing.
+			continue
```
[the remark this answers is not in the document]
cpuguy83: The git credential protocol is Unix-native and uses LF line endings. The credential helper runs inside a Linux container where `bufio.Scanner` splits on `\n`. Git itself does not emit `\r\n` in its credential helper input, so a bare `"\r"` line cannot occur in practice. Adding CRLF normalization here would be defensive against a scenario that does not arise, and would add complexity without a real benefit.

cmd/frontend/git_credential_helper_test.go
```diff
@@ -0,0 +1,325 @@
+package main
+
+import (
+	"encoding/base64"
+	"strings"
+	"testing"
+)
+
+func TestHandleSecretToken(t *testing.T) {
```
invidian, line 9: Maybe we could have integration tests testing with different git versions or something?

invidian (approved):
I think tests could be improved, but otherwise LGTM

cmd/frontend/git_credential_helper_test.go
cpuguy83: This is unfortunately non-trivial due to needing to use https for go modules.
I did originally start down the path of making the existing gomod auth tests target the distro tests, but TLS certs.

Instead I opted for unit tests with some manual tests to confirm against azlinux to validate that.

PR #1014, opened 2026-03-31
fix: merge User field in MergeSpecImage

### LINKED ISSUE #1013: [BUG] MergeSpecImage does not merge User field from target image config
## Expected Behavior

When `user` is set under `targets.<target>.image`, it should be applied to the built container image, consistent with how all other `ImageConfig` fields (`entrypoint`, `cmd`, `env`, `volumes`, `labels`, `working_dir`, `stop_signal`, `base`) are merged from the target into the top-level config.

## Actual Behavior

The `user` field set under `targets.<target>.image` is silently ignored. `MergeSpecImage` in `imgconfig.go` merges every other `ImageConfig` field from the target but skips `User`.

## Steps to Reproduce

Use a spec with `user` set at the target level:

```yaml
targets:
  azlinux3:
    image:
      user: "999"
```

Build a container image from this spec. The resulting image will not have the user set to `999`.

## Are you willing to submit a PR to contribute to this bug fix?

Yes.

officialasishkumar:

`MergeSpecImage` in `imgconfig.go` merges every `ImageConfig` field from the target config into the spec-level config — except for `User`. This causes the `user` field set under `targets.<target>.image` to be silently ignored when building a container image.

This PR adds the missing `User` field merge using the same override-if-non-empty pattern already used by all other string fields (`Entrypoint`, `Cmd`, `WorkingDir`, `StopSignal`, `Base`).

It also adds comprehensive unit tests for `MergeSpecImage` covering all field types: string overrides, env appending, map merging (volumes/labels), nil/empty cases, and the new `User` merge behavior.

**Which issue(s) this PR fixes** (optional, using `fixes #<issue number>` format,
will close the issue(s) when the PR gets merged):
Fixes #1013

The fix is a 4-line addition (`imgconfig.go:61-63`) following the exact same pattern as the other string fields in the function. `MergeImageConfig` (in `image.go`) already handles copying `User` from the merged `ImageConfig` into the Docker image config, so the downstream path works correctly once this merge is fixed.

### DIFF
--- a/imgconfig.go
+++ b/imgconfig.go
@@ -57,6 +57,10 @@ func MergeSpecImage(spec *Spec, targetKey string) *ImageConfig {
 		if i.Base != "" {
 			cfg.Base = i.Base
 		}
+
+		if i.User != "" {
+			cfg.User = i.User
+		}
 	}
 
 	return &cfg
--- a/imgconfig_test.go
+++ b/imgconfig_test.go
@@ -1,0 +1,183 @@
+package dalec
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
+)
+
+func TestMergeSpecImage(t *testing.T) {
+	t.Run("nil spec image returns empty config", func(t *testing.T) {
+		spec := &Spec{}
+		cfg := MergeSpecImage(spec, "foo")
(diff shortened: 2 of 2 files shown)


cpuguy83 (approved):
Nice find.
Thanks!

PR #1024, opened 2026-04-08
fix: correct series file entries and sort patch keys in deb target

burak-ok:
Fix two bugs in the deb patch handling:
1. sourcePatchesDir wrote the target source name instead of the patch source name to the series file, causing incorrect documentation.
2. createPatchScript iterated over the Patches map without sorting, causing non-deterministic patch application order unlike the RPM target which sorts keys.

Add unit tests for both issues.


cpuguy83 (approved):
LGTM, thanks!

PR #1032, opened 2026-04-09
[Debug] Use larger runner for (most) integration test suites

cpuguy83:


.github/workflows/ci.yml
```diff
@@ -163,106 +167,34 @@ jobs:
...
           fi
-
       - name: download deps
         run: go mod download
       - name: Setup QEMU
         run: docker run --rm --privileged tonistiigi/binfmt:latest --install all
       - name: Setup source policy
         if: inputs.source_policy
         uses: ./.github/actions/setup-source-policy
-      - name: Aggressive cleanup
```
cpuguy83, line 183: Disks are larger on the updated runners and this step takes 3mins by itself to run, so just not worth it anymore.

.github/workflows/ci.yml
```diff
@@ -88,7 +88,7 @@ jobs:
           fi
 
   integration:
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.suite == 'other' && 'ubuntu-22.04' || 'ubuntu-latest-4-cores' }}
```
cpuguy83, line 91: The runners we have access to are all `ubuntu-latest-<n>-cores`

PR #1061, opened 2026-05-14
frontend: add build-time source filters

cpuguy83:
Add a source-options context that can provide a YAML source filter
config at build time. The config currently supports global_excludes
which can be used to filter out paths from all sources.


source.go
```diff
@@ -474,25 +491,42 @@ type fetchOptions struct {
 	SourceOpt   SourceOpts
 }
 
-func (s *Source) fetchOptions(sOpt SourceOpts) fetchOptions {
+func (s *Source) fetchOptions(sOpt SourceOpts) (fetchOptions, error) {
+	excludes := s.Excludes
+	if s.IsDir() {
+		globalExcludes, err := sOpt.sourceFilterExcludes()
```
MorrisLaw, line 497: I think this _might_ be a little too early in the flow.

Since this gets added in `Source.fetchOptions`, it looks like the filter applies any time we materialize a directory source, including sources that are only being used as inputs to generators. IIUC that means a filter could exclude something like `go.sum`, `requirements.txt`, or `Cargo.lock` before the generator has a chance to use it.

Since we also filter the generated cache later, maybe this should live closer to the final source-package/debug-source assembly instead of in the generic source path?

Could we add a regression test for this case just to make sure?

cpuguy83: Thinking about this.
In this case I think that'd be a "don't do that" scenario.
The global filter is just that.
We can also add generator specific filters, but for now I've focused on just the global one as it is most pressing.

MorrisLaw (approved):

PR #1076, opened 2026-06-01
tests: verify worker platform selection without QEMU

cpuguy83:
## Problem

The `cross platform` integration test built packages for a foreign architecture (arm64 on amd64 CI runners), which required QEMU emulation. QEMU frequently crashes in CI during this test (e.g. https://github.com/project-dalec/dalec/actions/runs/26779889971/job/78941218259?pr=1068).

## Approach

Replace it with a native test that proves what the test actually cares about: **the frontend selects the correct platform-specific worker image for a requested build platform** — without any emulation or rebuilding the world.

- A custom multi-platform worker image is built as an **OCI layout index** with one manifest per real platform the distros ship (`linux/amd64` and `linux/arm64`). Each manifest embeds a `/platform-marker` file naming its own platform.
- The index is registered with the build session (`SolveOpt.OCIStores`) and supplied as the distro worker context via an `oci-layout:` named context.
- For each requested build platform, the test solves the **`worker` target** and asserts the returned image's `/platform-marker` matches the requested platform.

The distro worker target short-circuits when a worker context image is provided and returns it directly, so nothing is executed (no rpmbuild/dpkg, no emulation) — the worker image is only inspected. Requesting `arm64` on an `amd64` runner therefore needs no QEMU.

Real `amd64`/`arm64` platforms are used (rather than a made-up platform) because the worker handler resolves the real distro image config at the requested platform, so the requested platform must actually exist on that image.

### DIFF
--- a/test/linux_target_test.go
+++ b/test/linux_target_test.go
@@ -50,7 +50,12 @@ type workerConfig struct {
 	SignRepo   func(st llb.State, repoPath string) llb.StateOption
 	// ContextName is the name of the worker context that the build target will use
 	// to see if a custom worker is provided in a context
-	ContextName    string
+	ContextName string
+	// BaseImageRef is the distro's worker base image reference (the image the
+	// worker target resolves via the llb.Image path). It is used by the
+	// source-policy variant of the cross-platform test to build a marker image
+	// from the real base and to rewrite the base image to that marker store.
+	BaseImageRef   string
 	TestRepoConfig func(keyPath, repoPath string) map[string]dalec.Source
 	Platform       *ocispecs.Platform
 	SysextWorker   func(sOpt dalec.SourceOpts, opts ...llb.ConstraintsOpt) llb.State
@@ -108,9 +113,6 @@ type testLinuxConfig struct {
 	Release OSRelease
 
 	SupportsGomodVersionUpdate bool
-
-	Platforms         []ocispecs.Platform
-	PackageOutputPath func(spec *dalec.Spec, platform ocispecs.Platform) string
 }
 
 type OSRelease struct {
@@ -120,33 +122,6 @@ type OSRelease struct {
 
 func (cfg *testLinuxConfig) GetPackage(name string) string {
 	return cfg.Target.GetPackage(name)
-}
-
-func rpmTargetOutputPath(id string) func(spec *dalec.Spec, platform ocispecs.Platform) string {
-	return func(spec *dalec.Spec, platform ocispecs.Platform) string {
-		var arch string
-		switch platform.Architecture {
-		case "amd64":
-			arch = "x86_64"
-		case "arm64":
-			arch = "aarch64"
-		case "arm":
-			// this is not perfect, but all we really support in our tests for now, so its fine.
-			arch = "armv7l"
-		}
-		return fmt.Sprintf("/RPMS/%s/%s-%s-%s.%s.%s.rpm", arch, spec.Name, spec.Version, spec.Revision, id, arch)
-	}
-}
-
-func debTargetOutputPath(id string) func(spec *dalec.Spec, platform ocispecs.Platform) string {
-	return func(spec *dalec.Spec, platform ocispecs.Platform) string {
-		arch := platform.Architecture
-		if arch == "arm" {
-			// this is not perfect, but all we really support in our tests for now, so its fine.
-			arch = "armhf"
-		}
-		return fmt.Sprintf("%s_%s-%s_%s.deb", spec.Name, spec.Version, id+"u"+spec.Revision, arch)
-	}
 }
 
 func testLinuxDistro(ctx context.Context, t *testing.T, testConfig testLinuxConfig) {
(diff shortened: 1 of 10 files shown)


test/linux_target_test.go
```diff
@@ -3694,38 +3694,50 @@ func testTargetPlatform(ctx context.Context, t *testing.T, cfg testLinuxConfig)
 		bp := readDefaultPlatform(ctx, t, client)
```
invidian: Can we expct bp, tp and fallback names so it's easier to understand this code?

bp is host (build) platform ? So maybe `hostPlatform`?
tp is target platform so `targetPlatform`?

test/linux_target_test.go
```diff
@@ -3694,38 +3694,50 @@ func testTargetPlatform(ctx context.Context, t *testing.T, cfg testLinuxConfig)
...
 
 		var tp ocispecs.Platform
+		var fallback ocispecs.Platform
 		for _, p := range cfg.Platforms {
 			if platforms.OnlyStrict(bp).Match(p) {
 				continue
 			}
 
-			tp = p
+			if fallback.OS == "" {
```
invidian: Why we only assign fallback when OS is empty?

test/linux_target_test.go
```diff
@@ -3694,38 +3694,50 @@ func testTargetPlatform(ctx context.Context, t *testing.T, cfg testLinuxConfig)
 		bp := readDefaultPlatform(ctx, t, client)
 
 		var tp ocispecs.Platform
```
invidian: if `tp` is a result, maybe we can make a helper function for it and add some tests to make sure we don't goof up the logic here? Since it's already hard to follow IMO.

cpuguy83:
Took this in a new direction.

  Why: the existing cross platform test built a package/container for a foreign architecture and ran build steps inside that foreign-arch worker. On a single-arch CI runner that requires QEMU
  emulation, which crashes intermittently and makes the test flaky.

  What changed: instead of doing a real foreign-arch build, the test now verifies the thing that was actually at risk of regressing — that the frontend selects the correct platform-specific
  worker base image for the requested build platform — without executing any foreign-arch code.

  A small multi-platform "marker" image is built from the real distro worker base, injecting a /platform-marker file into each manifest that names its platform. The test then requests the 
  worker target for both linux/amd64 and linux/arm64 and reads the marker back to confirm the right manifest was selected. The marker is content baked into each manifest, so it proves the
  actual base bytes were fetched — not just that the result was labeled with the requested platform.

  Selection is checked through both resolution paths:

   - worker context — the marker index is supplied as the worker build context (short-circuits worker building); nothing executes.
   - source policy — a BuildKit source policy rewrites the distro base image to the marker index, exercising the normal llb.Image worker-build path. Cross-arch worker assembly runs natively 
  via dnf/apt --forcearch into a mounted foreign-arch root, so no emulation is needed.

  Both run natively regardless of the runner's architecture — no QEMU.

  Coverage notes:

   - TARGETARCH/TARGETVARIANT argument passthrough is already covered without QEMU by TestPassthroughVars (via debug/resolve).
   - The previous test's end-to-end assertions on package output naming and the final container image-config platform are no longer exercised by this test. A genuinely-different-arch real 
  build is only achievable without QEMU using a native sub-architecture (e.g. 386 on amd64), which is a possible future follow-up gated on per-distro 32-bit tooling availability.

test/platform_worker_test.go
```diff
@@ -0,0 +1,243 @@
...
+// containerd content.Store that BuildKit's SolveOpt.OCIStores requires.
+//
+// BuildKit's oci-layout source only exercises the read path: ReaderAt to fetch
+// blobs and Info to size the root descriptor. Every other content.Store method
+// is promoted from the embedded nil content.Store; none are called on the read
+// path and would panic if they were.
+type layoutStore struct {
+	content.Store // nil: unimplemented methods, never called on the read path
+	layout        ocijoin.Layout
+}
```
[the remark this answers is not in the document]
cpuguy83: This is the intent and is as documented.

test/linux_target_test.go
```diff
@@ -46,7 +46,12 @@ type workerConfig struct {
 	SignRepo   func(st llb.State, repoPath string) llb.StateOption
 	// ContextName is the name of the worker context that the build target will use
 	// to see if a custom worker is provided in a context
-	ContextName    string
+	ContextName string
+	// BaseImageRef is the distro's worker base image reference (the image the
+	// worker target resolves via the llb.Image path). It is used by the
+	// source-policy variant of the cross-platform test to build a marker image
+	// from the real base and to rewrite the base image to that marker store.
+	BaseImageRef   string
```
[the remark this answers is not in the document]
cpuguy83: CI says otherwise, also:

```console
❯ gofmt -s -w ./test/linux_target_test.go
❯ git diff
❯
```

cpuguy83:
@invidian PTAL

invidian (approved):
It feels like magic, but given it's testing code, LGTM. I don't see any obvious issues with it.

PR #1080, opened 2026-06-04
flatcar: validate sysext on a running VM

kartikjoshi21:

Adds a Flatcar runtime validation workflow that:
- builds a sysext with the current `flatcar/testing/sysext` target
- boots the pinned Flatcar Stable 4593.2.1 QEMU image
- provisions the generated `.raw` image under `/etc/extensions`
- restarts `systemd-sysext` and verifies the extension binary executes

The official Flatcar QEMU image and wrapper script are SHA-512 pinned. The test uses KVM when available and falls back to QEMU software emulation.

Refs #969

This complements artifact-level integration coverage by validating the output on a running Flatcar VM. It does not add a Flatcar-native package toolchain.

**Validation**:

- `go run ./cmd/lint ./...`
- `make test`
- `make check-generated`
- `actionlint .github/workflows/flatcar-runtime.yml`
- `flatcar-runtime` GitHub Actions job passed: Flatcar listed the extension in `systemd-sysext status` and executed its binary successfully

### DIFF
--- a/test/target_flatcar_runtime_test.go
+++ b/test/target_flatcar_runtime_test.go
@@ -1,0 +1,582 @@
+package test
+
+import (
+	"bytes"
+	"compress/bzip2"
+	"context"
+	"crypto/sha512"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+	ocispecs "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/project-dalec/dalec"
+	"github.com/project-dalec/dalec/targets/linux/flatcar"
+)
+
+const (
+	flatcarRuntimeVersion   = "4593.2.1"
+	flatcarRuntimeBaseURL   = "https://stable.release.flatcar-linux.net/amd64-usr/" + flatcarRuntimeVersion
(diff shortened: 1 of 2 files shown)


cpuguy83 (commented):
If we are going toa dd this level of testing, I think it should be in the integration suite.
We should also be able to do this with buildkit, though it will need a buildkitd that allows for higher privileged containers.

kartikjoshi21:
> If we are going toa dd this level of testing, I think it should be in the integration suite. We should also be able to do this with buildkit, though it will need a buildkitd that allows for higher privileged containers.

Yeah, that makes sense.  i was thinking of keeping it separate to keep the flow totally untouched. But reworked to use the same integration suite.

test/target_flatcar_runtime_test.go
```diff
@@ -0,0 +1,582 @@
...
+	flatcarRuntimeScriptSHA = "6398d3089f2a214c3604ebc796fa3135763142a9969fcac9a8fd2a6f3b64ad48025201ec7b96f733f275e164ff9bc447a6e23caa5b53ceb565aaa4829ec9232e"
+	flatcarRuntimeName      = "flatcar-runtime-test"
+	flatcarRuntimeOutput    = "hello from a Dalec Flatcar sysext"
+)
+
+func TestFlatcarRuntime(t *testing.T) {
+	if runtime.GOOS != "linux" {
+		t.Skip("Flatcar runtime validation requires Linux")
+	}
+	requireFlatcarRuntimeCommand(t, "qemu-system-x86_64")
```
cpuguy83, line 43: I was thinking we can do a lot of this inside a builtkit container (client.NewContainer, we have some examples of this under test/) so its not dependent on host tooling other than what we already have.

cpuguy83: Realized access to /dev/kvm is going to problematic for that.
We could possibly negotiate that over a unix socket that with the container and pass it off to qemu via `--add-fd`?

PR #1085, opened 2026-06-08
test: bind git HTTP test server to ephemeral port

cpuguy83:
## Problem

`TestSourceGitChecksumPreservesTagMetadata` (and occasionally `TestGomodGitAuth/HTTP`) failed intermittently with:

```
server error: failed to listen on 0.0.0.0:8080: listen tcp 0.0.0.0:8080: bind: address already in use
```

The integration-test git HTTP server hardcoded binding `0.0.0.0:8080`. The gateway containers that run these servers share the BuildKit worker's network namespace, so when two servers were alive at the same time the second `net.Listen` hit `EADDRINUSE`.

`TestSourceGitChecksumPreservesTagMetadata` runs two `t.Parallel()` subtests, each starting an HTTP git server on 8080, and `TestGomodGitAuth/HTTP` uses 8080 too. The failure only surfaced when these overlapped under the test scheduler, which is why it was flaky.

## Fix

Let the kernel assign a free port instead of hardcoding 8080:

- `cmd/server` (`runServe`): accepts `"0"` as the requested port, then reads the actual bound port from `listener.Addr()` and reports it in the ready event.
- `StartHTTPGitServer`: records the real port back into `Attr.HTTPPort` so the spec and client gitconfig generation that run afterward use it.
- `source_test.go` / `gomod_git_auth_test.go`: request port `"0"` instead of `8080`.

## Trade-off

Callers must start the HTTP server before generating specs or client gitconfig, since the concrete port is only known once the server is online. Current call sites already follow this order.

SSH (port 2222) is left unchanged — only one SSH server is started per test process, so it can't collide.

### DIFF
--- a/test/git_services/attributes.go
+++ b/test/git_services/attributes.go
@@ -24,7 +24,9 @@ type Attributes struct {
 	// PrivateGomoduleHost is the hostname of the git server
 	PrivateGomoduleHost string
 
-	// HTTPPort is the port on which the http git server runs
+	// HTTPPort is the requested port on which the http git server listens. Use
+	// "0" to request an ephemeral port; the actual bound port is reported back
+	// and written here once the server is online.
 	HTTPPort string
 	// SSHPort is the port on which the ssh git server runs
 	SSHPort string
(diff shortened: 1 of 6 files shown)


cpuguy83:
GHA appears to be wedged.

DannyBrito (approved):
lgtm

PR #1086, opened 2026-06-10
feat: Add abilility to drop gomod requires

chrischangcode:
## Problem

When a Go sub-module gets absorbed into its parent module in a newer version, upgrading the parent via a `replace` directive leaves an orphaned `require` in `go.mod` that can no longer be resolved.

**Real-world example:** `google.golang.org/grpc/stats/opentelemetry` was a separate module (with its own `go.mod`) up through grpc v1.68.x. Starting with grpc v1.79.0, it was merged into the main `google.golang.org/grpc` module — the code still exists as a package, but there is no longer a separate `go.mod` at that path.

When a CVE fix upgrades `google.golang.org/grpc` to v1.79.3 via a replace directive, `go mod tidy` fails because it cannot resolve the orphaned `require google.golang.org/grpc/stats/opentelemetry v0.0.0-...` — no module exists at that path in the newer version.

### Current workaround

Spec authors must create a **source patch** that adds a stub `go.mod` file and a **local directory replace** pointing to it. This works but adds maintenance burden.

See [Azure/dalec-build-defs#9887](https://github.com/Azure/dalec-build-defs/pull/9887) for an example.

## Solution

Add a dedicated `Drop` field to `GomodEdits` that emits `go mod edit -droprequire=<module>` arguments. The subsequent `go mod tidy` then cleanly resolves the dependency graph.

### Spec syntax

```yaml
sources:
  kube-state-metrics:
    generate:
      - gomod:
          edits:
            replace:
              - google.golang.org/grpc => google.golang.org/grpc@v1.79.3
            drop:
              - google.golang.org/grpc/stats/opentelemetry
```

### Implementation

- **`spec.go`**: Added `Drop []string` field to `GomodEdits` struct with documentation.
- **`preprocess.go`**: Updated `gomodEditArgs()` to iterate `Drop` entries and emit `-droprequire=` arguments.
- **`generator_gomod_test.go`**: Added tests covering: basic drop, multiple drops, mixed drop+replace, empty-path error, and nil-edits passthrough.
- **`gomod-patch.sh`**: No changes needed — the shell script already iterates edit args line-by-line and passes each to `go mod edit`, which natively supports `-droprequire=`.

### Why a dedicated `Drop` field?

Per reviewer feedback, a dedicated field is preferable to overloading `Replace` with a sentinel value:
- **Clear intent** — `drop:` is unambiguous; readers don't need to know about magic values
- **Type-safe** — `[]string` for module paths, no struct needed
- **No collision risk** — doesn't repurpose an existing field's semantics
- **Validates naturally** — empty strings are the only invalid input

### Verified

Tested locally against `kubernetes/kube-state-metrics` v2.16.0:

```sh
go mod edit \
  -replace google.golang.org/grpc=google.golang.org/grpc@v1.79.3 \
  -droprequire google.golang.org/grpc/stats/opentelemetry
go mod tidy   # succeeds
go build ./...  # succeeds
```

All existing tests pass (`go test ./...`).

### DIFF
--- a/docs/spec.schema.json
+++ b/docs/spec.schema.json
@@ -1115,6 +1115,17 @@
 		},
 		"GomodEdits": {
 			"properties": {
+				"drop": {
+					"items": {
+						"type": [
+							"string"
+						]
+					},
+					"type": [
+						"array"
+					],
+					"description": "Drop removes module requirements from go.mod via go mod edit -droprequire.\nEach entry is a module path to remove. This handles cases where a sub-module\nhas been absorbed into its parent module in a newer version\n(e.g. google.golang.org/grpc/stats/opentelemetry merged into grpc at v1.79.0)."
+				},
 				"replace": {
 					"items": {
 						"$ref": "#/$defs/GomodReplace"
--- a/generator_gomod_test.go
+++ b/generator_gomod_test.go
@@ -210,3 +210,89 @@ func TestGomodReplaceGoModEditArg(t *testing.T) {
 		})
 	}
 }
+
+func TestGomodEditArgs_DropRequire(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name         string
+		gen          *GeneratorGomod
+		expectErr    bool
+		expectedArgs string
+	}{
+		{
+			name: "drop converts to droprequire",
+			gen: &GeneratorGomod{
+				Edits: &GomodEdits{
+					Drop: []string{"example.com/old-submodule"},
+				},
+			},
+			expectedArgs: "-droprequire=example.com/old-submodule",
+		},
+		{
+			name: "multiple drops",
+			gen: &GeneratorGomod{
+				Edits: &GomodEdits{
+					Drop: []string{
+						"example.com/mod-a",
+						"example.com/mod-b",
+					},
+				},
+			},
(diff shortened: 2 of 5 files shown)


cpuguy83:
Why isn't `go mod tidy` dropping this automatically if the module isn't required anymore?

cpuguy83:
Ok I think I understand now.
I think it would be best to movet his to a `Drop` field rather than making `Replace` do it.

chrischangcode:
> Ok I think I understand now. I think it would be best to movet his to a `Drop` field rather than making `Replace` do it.

Agreed! But yeah upstream go.mod required the deprecated package. https://github.com/kubernetes/kube-state-metrics/blob/release-2.16/go.mod#L241

I also think it'd be possible to add logic to calculate the replaces with more context of the overall dependency graph for more successful PRs.

https://github.com/Azure/dalec-build-defs/issues/9801

cpuguy83 (changes requested):
Design looks good.
Linter is failing due (needs gofmt).
Should update the website docs to include the new field.

cpuguy83 (approved):

chrischangcode:
@cpuguy83  Thanks for the review! Will this get merged in closer to the release?

cpuguy83:
Yes, sorry. Cleaning up some bugs on the latest release so we don't need to manage release branches for now. I've got this one slotted in for v0.22

PR #1091, opened 2026-06-11
tests: avoid cross-distro build in cache dir test

### LINKED ISSUE #1034: Cache dir test builds separate distro
See https://github.com/project-dalec/dalec/pull/948

The PR is attempting to remove the need to build an entirely different distro to run the test.
The current, problematic implementation causes issues for CI because we try to split integration tests by distro so that we don't have to build the same distro multiple times and get maximal cache re-use. The test messes that up.

kartikjoshi21:
Add test-only alternate target registration behind the alt_testing_targets build tag, and update the cache dir test to use the current target's alternate key instead of building a different distro.

For DEB targets, copy the distro config and tweak VersionID so the alt target has a distinct cache namespace without requiring a separate distro worker stack. This preserves cache isolation coverage while keeping CI integration jobs scoped to their assigned distro.

Fixes #1034

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,14 +6,15 @@ COPY . .
 ENV CGO_ENABLED=0
 ARG TARGETARCH TARGETOS GOFLAGS=-trimpath
 ARG DALEC_FRONTEND_COVERAGE=0
+ARG EXTRA_BUILD_FLAGS=""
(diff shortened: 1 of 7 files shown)


cpuguy83 (approved):
This looks ok.
I still think we should rework how cache keys are derived.
I don't think deriving key by target is the right move but that's not something this PR is changing.

PR #1097, opened 2026-06-15
ci: bump buildkit to v0.31.0

cpuguy83:
Pins the CI builders to `moby/buildkit:v0.31.0` and bumps the buildkit Go module to match so both the integration (distro) matrix and the e2e suite build against the released buildkitd.

### Changes
- Add a top-level `BUILDKIT_TEST_IMAGE` env as the single source of truth for the buildkit image under test.
- **integration** job: create and `--use` a buildx container builder pinned to the test image instead of relying on the engine's embedded buildkit. The test harness (`buildx dial-stdio`) and the `bake frontend`/`bake worker` pre-builds all target this builder. The failure log dump now also collects the `buildx_buildkit_*` container logs.
- **e2e** job: create a pinned container builder for *both* diff/merge matrix legs (`USE_BUILDX=1` for both), rather than only the diff/merge-enabled leg using a container builder and the other falling back to the default docker driver.
- Bump `github.com/moby/buildkit` v0.30.0 -> v0.31.0 (with the transitive updates from `go mod tidy`).


.github/workflows/ci.yml
```diff
@@ -383,15 +398,20 @@ jobs:
           set -e
 
           dir="$(mktemp -d)"
-          f="${dir}/dockerd.log"
-          echo "DOCKERD_LOG_PATH=${f}" >> $GITHUB_OUTPUT
-          sudo journalctl -u docker > "${f}"
+          echo "DUMP_DIR=${dir}" >> $GITHUB_OUTPUT
```
invidian, line 401: I wonder if this kind of logic could be moved to script files or Go, so that we could more easily run it locally and test it, given the difficulties of doing that with GHA.

invidian (approved):

PR #1098, opened 2026-06-16
Use native buildkit symlink file action for post-install symlinks

cpuguy83:

Post-install symlinks (`image.post.symlinks`) were created by executing a shell
script (`ln -s`, `chown -h`, `chgrp -h`) in the worker, mounting `/etc/passwd`
and `/etc/group` to resolve ownership by name. Buildkit now provides a native
symlink file action, so this switches to it when it is available.

Support is detected via the `pb.CapFileSymlinkCreate` LLB capability, mirroring
the existing diff/merge capability detection. When the connected buildkit does
not advertise the capability — or `DALEC_DISABLE_SYMLINK=1` is set — Dalec falls
back to the original shell-exec implementation, so older buildkit versions keep
working.

The native path chains `llb.Mkdir` (for parent directories, which the symlink
action does not create itself) and `llb.Symlink` into a single file op resolved
against the installed rootfs. This avoids the extra `exec` and the `/etc/passwd`
/ `/etc/group` mounts.

Ownership semantics are preserved exactly. Chowning by user name alone would
also set the group to that user's primary group, whereas the exec fallback
leaves the unspecified side as root. To match, both user and group are always
set explicitly and an unspecified side stays `0`, mirroring `chown -h user` /
`chgrp -h group`.

Fixes #

- Capability detection lives in `frontend/gateway.go` (`SupportsSymlink`) and is
  wired in `frontend/router.go`, just like `DisableDiffMerge`.
- `DALEC_DISABLE_SYMLINK` is registered as a known build-arg in `load.go` and can
  be used to force the legacy behavior.
- Unit tests in `helpers_symlink_test.go` assert the marshaled LLB for both the
  native and exec paths (symlink actions, parent mkdirs, and ownership).
- The integration test `testLinuxPostInstallSymlinks` runs the post-install
  symlink scenario under **both** paths via subtests — native by default and the
  exec fallback with `DALEC_DISABLE_SYMLINK=1`. The equivalent block was removed
  from the large combined container test to avoid duplicating the coverage.

### DIFF
--- a/frontend/support.go
+++ b/frontend/support.go
@@ -1,0 +1,25 @@
+package frontend
+
+import (
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
+	"github.com/moby/buildkit/solver/pb"
+	"github.com/project-dalec/dalec"
+)
+
+// NativeSymlinkSupport reports whether the given client's buildkit supports the
+// native LLB symlink file action.
+func NativeSymlinkSupport(client gwclient.Client) dalec.NativeSymlinkSupport {
+	return &nativeSymlinkSupport{client}
+}
+
+type nativeSymlinkSupport struct {
+	client gwclient.Client
+}
+
+func (c *nativeSymlinkSupport) SupportsNativeSymlinks() bool {
+	opts := c.client.BuildOpts()
+	if opts.Opts["build-arg:DALEC_DISABLE_SYMLINK"] == "1" {
+		return false
+	}
+	return opts.LLBCaps.Supports(pb.CapFileSymlinkCreate) == nil
+}
--- a/helpers.go
+++ b/helpers.go
@@ -304,11 +304,13 @@ func ShArgsf(format string, args ...interface{}) llb.RunOption {
 	return ShArgs(fmt.Sprintf(format, args...))
 }
 
-// InstallPostSymlinks returns a RunOption that adds symlinks defined in the [PostInstall] underneath the provided rootfs path.
-func InstallPostSymlinks(post *PostInstall, worker llb.State, opts ...llb.ConstraintsOpt) llb.StateOption {
-	return func(in llb.State) llb.State {
-		const rootfsPath = "/tmp/rootfs"
+type NativeSymlinkSupport interface {
+	SupportsNativeSymlinks() bool
+}
 
+// InstallPostSymlinks returns a RunOption that adds symlinks defined in the [PostInstall] underneath the provided rootfs path.
+func InstallPostSymlinks(post *PostInstall, worker llb.State, symlinkSupport NativeSymlinkSupport, opts ...llb.ConstraintsOpt) llb.StateOption {
+	return func(in llb.State) llb.State {
 		if post == nil {
 			return in
 		}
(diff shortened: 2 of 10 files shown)


helpers.go
```diff
@@ -301,46 +310,134 @@ func InstallPostSymlinks(post *PostInstall, worker llb.State, opts ...llb.Constr
...
+
+	buf := bytes.NewBuffer(nil)
+	buf.WriteString("set -ex\n")
+
+	sortedKeys := SortMapKeys(post.Symlinks)
+	var needsMount bool
+	for _, oldpath := range sortedKeys {
+		cfg := post.Symlinks[oldpath]
+		newpaths := cfg.Paths
+		sort.Strings(newpaths)
```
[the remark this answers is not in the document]
cpuguy83: Good catch — fixed. `installPostSymlinksExec` now uses `slices.Clone(cfg.Paths)` before sorting, matching the native path, so neither path mutates the `SymlinkTarget.Paths` stored in `post.Symlinks`.

helpers_symlink_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"context"
+	"testing"
+
+	"github.com/moby/buildkit/client/llb"
+	"github.com/moby/buildkit/solver/pb"
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
+)
+
+func marshalToOps(ctx context.Context, t *testing.T, st llb.State) []*pb.Op {
```
invidian: We have `LLBOpsFromState` already, maybe we can use it?

invidian:
@cpuguy83 can you resolve conflicts?

frontend/gateway.go
```diff
@@ -174,6 +174,28 @@ func checkDiffMerge(client gwclient.Client) bool {
 	return true
 }
 
+var (
+	supportsSymlinkOnce sync.Once
+	supportsSymlink     atomic.Bool
+)
+
+// SupportsSymlink checks if the given client supports the native LLB symlink file action.
+func SupportsSymlink(client gwclient.Client) bool {
```
invidian: Is this called often enough to justify the need for a global variable? or alternatively, could we discover the capabilities like this when client is initialized and just respond with that? Perhaps gwclient.Client interface needs to be extended for this kind of things?

helpers.go
```diff
@@ -301,46 +310,134 @@ func InstallPostSymlinks(post *PostInstall, worker llb.State, opts ...llb.Constr
...
-			ProgressGroup("Add post-install symlinks"),
-		).AddMount(rootfsPath, in)
+	appendMkdir := func(dir string) {
+		// The symlink file action does not create parent directories and will
+		// fail if they do not exist, so create them first (without changing
+		// ownership, matching the `mkdir -p` behavior of the exec fallback).
+		if dir == "/" || dir == "." || dir == "" {
+			return
+		}
+		mkdir := llb.Mkdir(dir, 0o755, llb.WithParents(true))
```
invidian: nit: Maybe to avoid duplicate parameters?

```go
mkdir := llb.Mkdir
if chain != nil {
  mkdir = chain.Mkdir
}

chain = mkdir(dir, 0o755, llb.WithParents(true))

File: helpers.go [hunk 5]
```diff
@@ -301,46 +310,134 @@ func InstallPostSymlinks(post *PostInstall, worker llb.State, opts ...llb.Constr
...
+		if chain == nil {
+			chain = mkdir
+			return
+		}
+		chain = chain.Mkdir(dir, 0o755, llb.WithParents(true))
+	}
+
+	appendSymlink := func(oldpath, newpath string, symlinkOpts []llb.SymlinkOption) {
+		if chain == nil {
+			chain = llb.Symlink(oldpath, newpath, symlinkOpts...)
```
invidian: Ditto duplicate args.

frontend/router.go
```diff
@@ -100,6 +100,10 @@ func (r *Router) Handler(opts ...func(context.Context, gwclient.Client, *Router)
 			dalec.DisableDiffMerge(true)
 		}
 
+		if !SupportsSymlink(client) {
+			dalec.DisableSymlink(true)
```
invidian: That's very place-oriented and with ambiguous side effects. Can we somehow improve it by e.g. proper dependency injection/parameters between components?

invidian (commented):
Should we also add some documentation around this feature environment variable?

frontend/gateway.go
cpuguy83: gwclient.Client is not our interface and I've also avoided pulling it into the root package.

frontend/router.go
cpuguy83: I agree in principal, however these are tightly scoped. A frontend process runs once per daemon so the value truly is a global.
Even nested solves happen against a different frontend process.

Changing requires a significant refactor that adds a fair amount of complexity for no real gain.
Not just code complexity either, it complicates using the library as right now most things are just state options or things that interact nicely with standard buildkit client types.

We *could* use context values for this, but then we still have to wire contexts through the stack for the sole purpose of stashing values.

cpuguy83:
> Should we also add some documentation around this feature environment variable?

The args are not so much a feature as they are a way for us to test both modes. I'd rather not document them for public use.

frontend/router.go
cpuguy83: That said, this isn't as pervaisive as the Diffmerge stuff is. Maybe we can inject something from the caller here.

cpuguy83: Updated, PTAL.

frontend/support.go
```diff
@@ -0,0 +1,32 @@
...
+func (c *nativeSymlinkSupport) SupportsNativeSymlinks() bool {
+	return supportsSymlink(c.client)
+}
+
+// supportsSymlink checks if the given client supports the native LLB symlink file action.
+func supportsSymlink(client gwclient.Client) bool {
+	if client.BuildOpts().Opts["build-arg:DALEC_DISABLE_SYMLINK"] == "1" {
+		return false
+	}
+	return checkSymlinkCap(client)
```
invidian: nit: that indirection does not buy us anything? Neither does `supportsSymlink` tbh. It does not reduce complexity or improves an interface, just seems like adding a costly mental indirection.
I think we could cram everything up here into a single function (method).

frontend/support.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func checkSymlinkCap(client gwclient.Client) bool {
+	caps := client.BuildOpts().LLBCaps
+	return caps.Supports(pb.CapFileSymlinkCreate) == nil
+}
+
+func NativeSymlinkSupport(client gwclient.Client) dalec.NativeSymlinkSupport {
```
invidian: nit: exported types could be at the top of the file, since they are more often accessed (interface) than unexported (implementation) ones.

helpers_symlink_test.go
```diff
@@ -0,0 +1,290 @@
...
+		Symlinks: map[string]SymlinkTarget{
+			"/usr/bin/src1": {Paths: []string{"/src1"}, User: "need"},
+			"/usr/bin/src2": {Paths: []string{"/non/existing/dir/src2"}, Group: "coffee"},
+			"/usr/bin/src3": {Paths: []string{"/non/existing/dir/src3", "/non/existing/dir2/src3"}, User: "need", Group: "coffee"},
+			"/usr/bin/src4": {Paths: []string{"/plain"}},
+		},
+	}
+}
+
+func TestInstallPostSymlinksNative(t *testing.T) {
```
invidian, line 16: nit: ditto helper and tests order, tests are more important than their helpers.

helpers_symlink_test.go
```diff
@@ -0,0 +1,290 @@
...
+	ctx := context.Background()
```
invidian: nit: we can use t.Context().

invidian (approved):
Just some nitpicks, overall LGTM

helpers_symlink_test.go
cpuguy83: Done, using LLBOpsFromState now.

frontend/gateway.go
cpuguy83: Dropped the global. Support is now injected via NativeSymlinkSupport from the frontend (where the client lives) into InstallPostSymlinks.

helpers.go
cpuguy83: Done.

helpers.go
cpuguy83: Done.

frontend/router.go
cpuguy83: Reworked. The capability is injected into InstallPostSymlinks now instead of being stashed in a package global here, so this block is gone.

frontend/support.go
cpuguy83: Collapsed into a single method.

frontend/support.go
cpuguy83: Moved the exported constructor to the top.

helpers_symlink_test.go
cpuguy83: Reordered, tests first.

helpers_symlink_test.go
cpuguy83: Done.

invidian (approved):

PR #1111, opened 2026-06-20
Make generated LLB deterministic by sorting map iteration

cpuguy83:
## What

Several code paths iterated Go maps directly while assembling LLB and while generating build artifacts. Go randomizes map iteration order, so the same input spec produced different op definitions / file contents from build to build, changing BuildKit vertex digests — defeating the cache and making output non-reproducible.

This sorts those iterations (`SortedMapIter`, or `SortMapKeys` where only keys are used) so identical specs yield identical LLB and identical generated files. The cost is sorting a few small maps; behavior is otherwise unchanged.

## Fixed sites

LLB op digest:
- `source_image.go` — image-source command env and run-step env (`AddEnv`)
- `preprocess.go` — gomod patch run env (`AddEnv`)
- `helpers.go` `GetRepoKeys` — repo GPG key mounts + returned names slice
- `generator_gomod.go` `withGomodSecretsAndSockets` — gomod `Auth` secrets/SSH sockets
- `targets/windows/handle_zip.go` — Windows build env (`AddEnv`)
- `targets/linux/rpm/distro/dnf_install.go` — dnf download argument list (the DEB equivalent was already sorted via `AppendConstraints`)
- `packaging/linux/deb/pkg.go` — `mountSources` source mounts and `TarDebSources` merge order feeding `MergeAtPath`/`Tar`

Generated build content:
- `targets/windows/handle_zip.go` — build-script `export` lines
- `packaging/linux/deb/template_rules.go` `Envs()` — `debian/rules` export lines

Narrow selection:
- `targets/linux/deb/distro/pkg.go` `buildCandidatePaths` — alt-Go candidate path search (first-match could vary)

## Intentionally not changed

- `frontend/build.go` `time.Now()` → image `Created`: image-config reproducibility (needs `SOURCE_DATE_EPOCH`), not LLB op digest.
- `template_rules.go:177` `maps.Keys(grouping)[0]`: output bytes are invariant here (`requiresCustomEnable` forces `--no-enable` in the only mixed case) — latent fragility, not a current bug.
- `ProgressGroup(identity.NewID())`: op metadata only, does not affect op digest.

### DIFF
--- a/packaging/linux/deb/determinism_test.go
+++ b/packaging/linux/deb/determinism_test.go
@@ -1,0 +1,106 @@
+package deb
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/moby/buildkit/client/llb"
+	"github.com/project-dalec/dalec"
+	"gotest.tools/v3/assert"
+)
+
+// determinismRuns is intentionally well above the 2 runs needed to spot an
+// obvious diff: marshaling is cheap, and extra runs drive the chance of a
+// randomized map iteration happening to repeat its order toward zero.
+const determinismRuns = 25
+
+// requireDeterministicLLB rebuilds the state on each run so the underlying map
+// iteration executes again, then asserts the marshaled op definitions are
+// byte-identical. Only Def (the op protos whose digests drive caching) is
+// compared; Metadata carries per-build noise and is ignored.
+func requireDeterministicLLB(ctx context.Context, t *testing.T, build func() llb.State) {
+	t.Helper()
+
+	want, err := build().Marshal(ctx)
+	assert.NilError(t, err)
+
+	for i := 1; i < determinismRuns; i++ {
+		got, err := build().Marshal(ctx)
+		assert.NilError(t, err)
+		assert.DeepEqual(t, got.Def, want.Def)
+	}
+}
+
+func manySourceStates() map[string]llb.State {
+	states := make(map[string]llb.State, 12)
+	for i := 0; i < 12; i++ {
+		name := fmt.Sprintf("src-%02d", i)
+		states[name] = llb.Scratch().File(llb.Mkfile("/"+name, 0o644, []byte(name)))
+	}
+	return states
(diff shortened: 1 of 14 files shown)


packaging/linux/deb/pkg.go
```diff
@@ -24,7 +24,7 @@ const (
 
 func mountSources(sources map[string]llb.State, dir string, mod func(string) string) llb.RunOption {
 	return dalec.RunOptFunc(func(ei *llb.ExecInfo) {
-		for key, src := range sources {
+		for key, src := range dalec.SortedMapIter(sources) {
```
invidian, line 27: Can we add some test cases (comparing generated LLBs across 2runs for diff?) to cover these?

cpuguy83: Done.

packaging/linux/deb/determinism_test.go
```diff
@@ -0,0 +1,106 @@
...
+// determinismRuns is intentionally well above the 2 runs needed to spot an
+// obvious diff: marshaling is cheap, and extra runs drive the chance of a
+// randomized map iteration happening to repeat its order toward zero.
+const determinismRuns = 25
+
+// requireDeterministicLLB rebuilds the state on each run so the underlying map
+// iteration executes again, then asserts the marshaled op definitions are
+// byte-identical. Only Def (the op protos whose digests drive caching) is
+// compared; Metadata carries per-build noise and is ignored.
+func requireDeterministicLLB(ctx context.Context, t *testing.T, build func() llb.State) {
```
invidian, line 23: Won't some integration test for it be good enough for it?

cpuguy83: I get wanting to do integration tests, but that is an entirely different change.
Likely needing to inject determinism checks across the suite rather than a specific test.
These tests validate that the modified functions are deterministic.
I don't want to deal with integration test changes in this change.

invidian (approved):

PR #1113, opened 2026-06-22
Add a post-install minimization pass for RPM container images 

### LINKED ISSUE #965: [REQ] Post-build image minimization

None

Problem statement

DALEC builds “minimal” container images by installing only what the package dependency tree implies. In practice, that tree is an unreliable proxy for runtime needs:

Over-declared deps / bad splits: packages pull in extras that the final app never uses.
Maintainer-script dependencies: shells/interpreters/tools needed only during install/config, not at runtime (e.g., bash required for postinst scripts).
Bundled functionality: one package drags in broad stacks that should be split upstream, but often aren’t.
Result: images are larger than necessary and “minimal” is constrained by packaging quality rather than actual runtime requirements.

Proposed work item: “Image Stripper” (post-build runtime minimization)
Build an offline post-processing step that removes files/packages from an already-built rootfs/image without invoking the package manager, using package metadata only as information, not as the mechanism.

Core idea: derive runtime requirements of container image of what’s actually required, then delete everything outside that list.

Goals / non-goals

Goals

Reduce final image size by removing unused files and optionally unused packages.
Do not run package manager transactions in the strip phase.
Handle maint-script deps cleanly (keep them during build, drop if not needed at runtime).

Non-goals:
Perfect static analysis of arbitrary binaries/scripts.

kartikjoshi21:
Add a post-install minimization pass for RPM container images assembled by DALEC. The pass seeds from the built spec RPMs and DALEC base RPMs, follows runtime rpmdb requirements, skips scriptlet-only requirements, and erases packages outside the runtime closure with rpm --root using --noscripts, --notriggers, and --nodeps.

This keeps rpmdb authoritative for scanners while removing package-manager and install-time leftovers. Custom base images are intentionally excluded so DALEC does not unexpectedly strip user-provided base image contents. The final rootfs is squashed after minimization so removed files do not remain in image layers.

This is a conservative RPM-first minimization step toward the image stripping work. It does not implement full file-level runtime reachability analysis or remove unused files from packages that remain installed.

Fixes #965


cpuguy83 (commented):
For reference, AZL has a minimization script for distroless containers.
I haven't reviewed either this PR or their script yet, just putting it here for reference: https://github.com/microsoft/azurelinux/blob/4.0/base/images/container-base/config.sh

I know we can't take that directly since they do things like remove the rpm db (which we want to keep).

kartikjoshi21:
> For reference, AZL has a minimization script for distroless containers. I haven't reviewed either this PR or their script yet, just putting it here for reference: https://github.com/microsoft/azurelinux/blob/4.0/base/images/container-base/config.sh
> 
> I know we can't take that directly since they do things like remove the rpm db (which we want to keep).

Thanks for the reference @cpuguy83 . I reviewed the script; its package-closure approach is similar, but the remaining cleanup is specific to AZL’s controlled distroless images. DALEC must preserve the RPM database for scanners and cannot safely remove certificates, locales, licences, documentation, or systemd files by default.
I’ll keep this PR focused on runtime package minimization and rpmdb preservation. Broader file-level stripping can be considered separately with an explicit policy.

targets/linux/rpm/distro/container_test.go
```diff
@@ -0,0 +1,191 @@
+package distro
```
invidian: Any integration tests we could re-use or add from `minimal_container` suite?

targets/linux/rpm/distro/container_test.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	assertProgressGroup(t, ops, "Minimize RPM container", false)
+	assertProgressGroup(t, ops, "Squash RPM container", false)
+}
+
+func TestRPMMinimizeScriptPreservesScannerUsableDB(t *testing.T) {
+	t.Parallel()
+
+	required := []string{
```
invidian: Seems like integration test would be better for this.

targets/linux/rpm/distro/container_test.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	state := c.BuildContainer(t.Context(), &testClient{}, sOpt, &dalec.Spec{}, "target", rpmFixture())
+
+	ops, err := test.LLBOpsFromState(t.Context(), state)
+	if err != nil {
+		t.Fatalf("failed to get llb ops from state: %v", err)
+	}
+
+	assertProgressGroup(t, ops, "Minimize RPM container", true)
```
invidian: Perhaps integration tests would allow us to assert on layers and other properties similarly to how minimal_container suite does it, e.g. in `squash_produces_single_layer`.

targets/linux/rpm/distro/minimize.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+seed_from_rpms() {
+	local dir="$1"
+
+	[ -d "${dir}" ] || return 0
+
+	while IFS= read -r -d '' rpm_file; do
+		rpm -qp --qf '%{NAME}\n' "${rpm_file}" >> "${seed_file}"
+	done < <(find "${dir}" -type f -name '*.rpm' -print0)
```
invidian: Can the RPMs be nested in the given directories or are they going to be always in the flat structure? If the latter, maybe simple `ls` would do, similarly how the deb container minimization is implemented?

targets/linux/rpm/distro/minimize.go
```diff
@@ -0,0 +1,283 @@
...
+
+for path in \
+	/etc/passwd \
+	/etc/group \
+	/etc/shadow \
+	/etc/gshadow \
+	/etc/subuid \
+	/etc/subgid \
+	/etc/nsswitch.conf; do
+	seed_from_rootfs_path "${path}"
```
invidian: If `seed_from_rootfs_path` is only used once, maybe we can inline it? That should make the script more readable I think? I'm not sure having it separately buys us a lot.

targets/linux/rpm/distro/minimize.go
```diff
@@ -0,0 +1,283 @@
...
+	rpm --root "${rootfs}" "$@"
+}
+
+seed_from_rpms() {
+	local dir="$1"
+
+	[ -d "${dir}" ] || return 0
+
+	while IFS= read -r -d '' rpm_file; do
+		rpm -qp --qf '%{NAME}\n' "${rpm_file}" >> "${seed_file}"
```
invidian: Does working with state file like `seed_file` make this script much simpler? It seems to leverage side effects and being quite place oriented, which makes it harder to follow. If this function could return the list of files instead and we would append on the outer, then pass the result to the next function, I think that would make data flow much simpler to follow.

targets/linux/rpm/distro/minimize.go
```diff
@@ -0,0 +1,283 @@
...
+done
+
+sort -u "${seed_file}" -o "${seed_file}"
+if [ ! -s "${seed_file}" ]; then
+	echo "no RPM seed packages found for minimization" >&2
+	exit 1
+fi
+
+while IFS= read -r pkg; do
+	add_installed_seed "${pkg}"
```
invidian: Ditto function.

invidian (commented):
I haven't finished reviewing minimize.go script, since I find it hard to read in the current form. Will finish reviewing it in next round.

targets/linux/rpm/distro/container_test.go
kartikjoshi21: Added RPM coverage to the existing minimal_container integration-test path.

targets/linux/rpm/distro/container_test.go
kartikjoshi21: Added squash_produces_single_layer, which verifies that the final RPM image has at most one layer.

targets/linux/rpm/distro/minimize.go
kartikjoshi21: RPM output uses the /RPMS/<arch>/*.rpm layout consumed by the installer, Replaced the recursive find invocation.

targets/linux/rpm/distro/minimize.go
kartikjoshi21: Removed the file-backed seed, queue, keep, installed, and removal flow. Functions now return package lists, while arrays carry the data between each step.

kartikjoshi21:
> I haven't finished reviewing minimize.go script, since I find it hard to read in the current form. Will finish reviewing it in next round.

@invidian  Thanks for the review. I refactored minimize.go to make the data flow easier to follow. Please take a look.

targets/linux/rpm/distro/minimize.go
```diff
@@ -0,0 +1,259 @@
...
+	"${rootfs}/var/lib/dnf" \
+	"${rootfs}/var/lib/yum" \
+	"${rootfs}/var/log/dnf.log" \
+	"${rootfs}/var/log/dnf.librepo.log" \
+	"${rootfs}/var/log/hawkey.log" \
+	"${rootfs}/var/log/tdnf.log" \
+	"${rootfs}/var/log/yum.log"
+`
+
+func minimizeContainer(rootfs, worker, rpmDir, basePkgs llb.State, opts ...llb.ConstraintsOpt) llb.State {
```
cpuguy83: Can you make this return an `llb.StateOption`?

targets/linux/rpm/distro/minimize.go
```diff
@@ -0,0 +1,259 @@
...
+	).AddMount(workPath, rootfs)
+}
+
+func squashContainer(rootfs llb.State, opts ...llb.ConstraintsOpt) llb.State {
+	opts = append(opts, dalec.ProgressGroup("Squash RPM container"))
+
+	return llb.Scratch().File(llb.Copy(rootfs, "/", "/", &llb.CopyInfo{
+		CopyDirContentsOnly: true,
+		CreateDestPath:      true,
+		AllowWildcard:       true,
```
cpuguy83: Looks unused.

targets/linux/rpm/distro/minimize.go
```diff
@@ -0,0 +1,259 @@
...
+		llb.AddMount(rpmMountDir, rpmDir, llb.SourcePath("/RPMS"), llb.Readonly),
+		llb.AddMount(baseMountPath, basePkgs, llb.SourcePath("/RPMS"), llb.Readonly),
+		llb.Args([]string{scriptPath}),
+	).AddMount(workPath, rootfs)
+}
+
+func squashContainer(rootfs llb.State, opts ...llb.ConstraintsOpt) llb.State {
+	opts = append(opts, dalec.ProgressGroup("Squash RPM container"))
+
+	return llb.Scratch().File(llb.Copy(rootfs, "/", "/", &llb.CopyInfo{
```
cpuguy83: If we run this script as part of the container setup, we can avoid this extra, expensive op.

targets/linux/rpm/distro/container.go
```diff
@@ -73,10 +73,18 @@ func (cfg *Config) BuildContainer(ctx context.Context, client gwclient.Client, s
 		frontend.IgnoreCache(client, targets.IgnoreCacheKeyContainer),
 	).AddMount(workPath, rootfs)
 
+	if !skipBase {
+		rootfs = minimizeContainer(rootfs, worker, rpmDir, basePkgs, opts...)
```
cpuguy83: I have a feeling we should make the minimization opt-in so we don't break existing builds.

Either a new target (`container/v2`?) or something else passed into the build.

I suppose the spec could itself have an opt-in, something like:

```yaml
image:
  minimizatonProfile: default # or v1? Not sure what to name it. Empty would mean no minimization
```

cpuguy83 (changes requested):

targets/linux/rpm/distro/minimize.go
kartikjoshi21: Minimization is now applied directly as a rootfs state option, and the separate scratch copy/squash operation has been removed.

targets/linux/rpm/distro/container.go
kartikjoshi21: Implemented with image.minimization_profile: default

cpuguy83 (approved):

PR #1120, opened 2026-06-25
test: verify frontend produces deterministic LLB

cpuguy83:
Re-runs each successful frontend solve with `Evaluate: false` and compares the generated op definitions, failing the test if they differ.

Wired as the outermost client wrapper in the test harness, so all integration solves are covered (including direct `gwc.Solve` calls that bypass the `solveT` helpers). Skips raw-definition solves and negative-path (errored) solves. The re-solve doesn't execute build steps, so overhead is just LLB generation.

Complements the unit-level checks in `determinism_test.go` by exercising the full frontend build path.

### DIFF
--- a/internal/testrunner/determinism_test.go
+++ b/internal/testrunner/determinism_test.go
@@ -1,0 +1,88 @@
+package testrunner
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/moby/buildkit/client/llb"
+	"github.com/project-dalec/dalec"
+	"gotest.tools/v3/assert"
+)
+
+func TestFileChecksProduceDeterministicLLB(t *testing.T) {
+	spec := &dalec.TestSpec{Files: manyFileChecks()}
+	base := llb.Image("example.com/base:latest")
+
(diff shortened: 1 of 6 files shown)


internal/testrunner/determinism_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+func manyTestEnv() map[string]string {
+	env := make(map[string]string, 12)
+	for i := 0; i < 12; i++ {
+		env[fmt.Sprintf("VAR_%02d", i)] = fmt.Sprintf("value-%d", i)
+	}
+	return env
+}
+
+func TestFileChecksProduceDeterministicLLB(t *testing.T) {
```
invidian, line 13: nit: test cases could go at the top of the file, since they are the most important content of this file, while all helper functions could land at the bottom, being the implementation details.

invidian (approved):

internal/testrunner/determinism_test.go
cpuguy83: Sure done.

PR #1132, opened 2026-07-06
deb: skip dh_strip_nondeterminism and dh_shlibdeps when disabled

cpuguy83:
`disable_strip` and `disable_auto_requires` are meant to turn off stripping and automatic dependency discovery, but each only covers part of what debhelper runs:

- `disable_strip` overrides `dh_strip` but leaves `dh_strip_nondeterminism` running.
- `disable_auto_requires` drops `${shlibs:Depends}` from the control file but leaves `dh_shlibdeps` running (its output is then discarded).

Those leftover helpers can fail on inputs the user has explicitly opted out of processing (e.g. corrupt archive fixtures, foreign-arch ELF fixtures — hit when building msft-golang). The existing options should already cover them, and there's no good spec-level workaround since mid-build `debian/rules` overrides aren't honored on newer dpkg.

This extends the existing options to emit the matching empty override targets at generation time:
- `disable_strip` also skips `dh_strip_nondeterminism`
- `disable_auto_requires` also skips `dh_shlibdeps`

### DIFF
--- a/packaging/linux/deb/template_rules_test.go
+++ b/packaging/linux/deb/template_rules_test.go
@@ -193,15 +193,29 @@ Depends: ${misc:Depends},
 }
 
 func TestRules_OverrideStrip(t *testing.T) {
-	w := &rulesWrapper{
-		Spec: &dalec.Spec{},
-	}
+	t.Run("strip enabled by default emits no overrides", func(t *testing.T) {
+		w := newRulesWrapper(dalec.Artifacts{})
+		assert.Equal(t, w.OverrideStrip().String(), "")
+	})
 
-	got := w.OverrideStrip()
-	assert.Equal(t, got.String(), "")
+	t.Run("disable_strip disables dh_strip and dh_strip_nondeterminism", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


packaging/linux/deb/template_rules_test.go
```diff
@@ -193,15 +193,29 @@ Depends: ${misc:Depends},
 }
 
 func TestRules_OverrideStrip(t *testing.T) {
-	w := &rulesWrapper{
```
invidian, line 196: Do we have some integration test scenarios which would cover those?

invidian (approved):

packaging/linux/deb/template_rules_test.go
cpuguy83: We have tests to validate the strip behavior.
We have tests to validate auto-requirement sniffing is disabled when that's toggled.

We don't have tests that validate that each specific debhelper program.
e.g., dh_strip will error out on a mangled ELF header or dh_strip_non_determinism errors out on a managed png file -- stuff that's within golang's corpus of test data that also gets shipped in GOROOT when packaging.

PR #1135, opened 2026-07-08
Resolve YAML aliases and merge keys inside dependency maps and build steps

### LINKED ISSUE #1134: YAML aliases inside a dependencies package map fail to resolve ("could not find alias")
### Summary

YAML aliases (and `<<:` merge keys) are not resolved when they are referenced **inside a package map** under `dependencies.build` (or `dependencies.runtime`). Loading such a spec fails with `unmarshal package constraints: … could not find alias "…"`, even though the anchor is defined at the top level of the document.

The same alias resolves correctly when the merge is applied one level higher, at the `dependencies:` node. So this is specific to alias references that land inside the package-constraints map.

### Reproduction

Using `dalec.LoadSpec` directly (dalec v0.21.1, github.com/goccy/go-yaml v1.19.2):

```go
package main

import (
	"fmt"

	"github.com/project-dalec/dalec"
)

func try(name, spec string) {
	_, err := dalec.LoadSpec([]byte(spec))
	if err != nil {
		fmt.Printf("%s => ERR: %v\n", name, err)
		return
	}
	fmt.Printf("%s => OK\n", name)
}

func main() {
	// (A) alias used as a package's value — FAILS
	try("alias as package value", `
name: t
version: "1.0"
revision: "1"
packager: t
vendor: t
license: MIT
description: t
x-c: &c
  version: [">= 1.0"]
dependencies:
  build:
    somepkg: *c
`)

	// (B) merge key inside the package map — FAILS
	try("merge key inside package map", `
name: t
version: "1.0"
revision: "1"
packager: t
vendor: t
license: MIT
description: t
x-b: &b
  somepkg:
    version: [">= 1.0"]
dependencies:
  build:
    <<: *b
    gnupg2:
`)

	// (C) merge at the dependencies level — OK
	try("merge at dependencies level", `
name: t
version: "1.0"
revision: "1"
packager: t
vendor: t
license: MIT
description: t
x-d: &d
  build:
    somepkg:
      version: [">= 1.0"]
dependencies:
  <<: *d
`)
}
```

Output:

```
alias as package value => ERR: error unmarshalling spec: error unmarshalling to spec: unmarshal package constraints: unmarshal package constraints: [13:15] could not find alias "c"
merge key inside package map => ERR: error unmarshalling spec: error unmarshalling to spec: unmarshal package constraints: unmarshal package constraints: [14:10] could not find alias "b"
merge at dependencies level => OK
```

### Expected

All three should load successfully. A YAML alias/merge that references a top-level anchor should resolve regardless of where in the document it is used, including inside a package map.

### Actual

Alias references inside the package map (cases A and B) fail with `could not find alias`.

### Notes on root cause

`PackageDependencyList.UnmarshalYAML` (`deps.go`) iterates `mapNode.Values` and decodes each package's value node via `dec.DecodeFromNodeContext(ctx, v.Value, &pc)`, and `PackageConstraints.UnmarshalYAML` similarly decodes its node via `yaml.NodeToValue`. These per-node decodes don't appear to carry the document-level anchor definitions, so an alias node encountered there can't be resolved. Decoding the equivalent subtree at the `dependencies:` level goes through a path that still has the anchor table, which is why case (C) works.

### Environment

- dalec: v0.21.1
- github.com/goccy/go-yaml: v1.19.2

cpuguy83:
Fixes #1134.

YAML aliases and `<<:` merge keys that reference a top-level anchor failed to load
with `could not find alias` when used inside a dependency package map — and, as an
audit turned up, inside `build.steps` as well. The same anchor resolves when the
merge is applied at a level that is decoded as a whole (e.g. the `dependencies:`
node), which is the tell for the root cause.

## Why

The custom unmarshallers that attach source maps re-decoded child AST nodes in
isolation, without the document's anchor table:

- `PackageConstraints.UnmarshalYAML` used `yaml.NodeToValue` (no reference readers).
- `PackageDependencyList.UnmarshalYAML` hand-iterated the mapping and never expanded
  `<<:` merge-key nodes.
- `BuildStepList.UnmarshalYAML` hand-iterated the sequence and decoded each step with
  `yaml.NodeToValue`, duplicating what `BuildStep.UnmarshalYAML` already does.

None of these isolated decodes carried the document's anchors, so an alias landing
inside them could not be resolved. This regressed in `ef94b3d8` when the node-based
unmarshallers were introduced for source maps.

The project already has the right primitive: `getDecoder(ctx)` builds a decoder
configured with `yaml.ReferenceReaders(baseDoc)`, which repopulates the anchor
table. The fix routes these decodes through it and lets the YAML library expand
merge keys and resolve aliases natively.

## Changes

- Decode `PackageConstraints` through the reference-reader decoder so an alias used
  as a package value resolves against the top-level anchor.
- Decode the whole package mapping in one pass so `<<:` merge keys and aliases are
  expanded natively, keeping the source-map refinement (start line pulled up to the
  package key) for directly-specified entries; merged entries keep the source map of
  their anchor definition.
- Decode `build.steps` as a whole sequence and let `BuildStep` resolve aliases/merge
  keys and attach its own source map (removing the redundant, buggy manual loop).
- Fix dependency lists written in the sequence-of-names form being silently dropped:
  the decoded result was built but never assigned.

## Audit

I checked every custom `UnmarshalYAML` / node re-decode and verified each with a real
spec. The only broken sites were the dependency maps and `build.steps` (both fixed
here). Scalar `sourceMappedValue` fields (test `command`, dependency names, uid/gid)
and the full-document decodes are fine, because they are reached through a
`getDecoder`-decoded parent (or already contain the anchor definitions).

## Tests

- `deps_test.go`: alias-as-value, merge-key-inside-map, merge at the `dependencies`
  level, explicit-overrides-merged precedence, aliases for runtime lists, the
  sequence-of-names form, and source-map attribution for merged vs. direct packages.
- `spec_test.go`: aliased and merge-key build steps resolve to the anchor's step.

### DIFF
--- a/deps.go
+++ b/deps.go
@@ -44,7 +44,11 @@ func (pc *PackageConstraints) UnmarshalYAML(ctx context.Context, node ast.Node) 
 	var i internal
 
 	if node.Type() != ast.NullType {
-		if err := yaml.NodeToValue(node, &i, decodeOpts(ctx)...); err != nil {
+		// Decode via getDecoder so the document's anchor definitions (carried as
+		// reference readers) are available; this lets an alias used as a
+		// package's value resolve against a top-level anchor.
+		dec := getDecoder(ctx)
+		if err := dec.DecodeFromNodeContext(ctx, node, &i); err != nil {
 			return errors.Wrap(err, "unmarshal package constraints")
 		}
 	}
@@ -86,32 +90,55 @@ func (l *PackageDependencyList) UnmarshalYAML(ctx context.Context, node ast.Node
 			}
 		}
 
+		*l = result
+
 		return nil
 	}
 
-	result := make(PackageDependencyList)
+	// Decode the whole mapping through getDecoder so the document's anchor
+	// definitions are available. This lets the yaml library resolve aliases used
+	// as package values and expand `<<:` merge keys that reference top-level
+	// anchors. Decode into the unnamed map type so this method is not re-entered.
+	result := make(map[string]PackageConstraints)
 	dec := getDecoder(ctx)
-	for _, v := range mapNode.Values {
-		var pc PackageConstraints
-		if err := dec.DecodeFromNodeContext(ctx, v.Value, &pc); err != nil {
-			return errors.Wrap(err, "unmarshal package constraints")
-		}
+	if err := dec.DecodeFromNodeContext(ctx, mapNode, &result); err != nil {
+		return errors.Wrap(err, "unmarshal package constraints")
+	}
 
+	// Merged/aliased entries keep the source map of their anchor definition.
+	// Directly-specified entries need an adjustment: when a package constraint is
+	// specified, e.g.
+	//   pkg-name:
+	//   	version: [">=1.0.0"]
+	// the value node's source map points at the line below `pkg-name`. However,
+	// we want to include the package name itself in the source map, so pull the
+	// start line up to the package key.
+	for _, v := range mapNode.Values {
+		// A merge key (`<<:`) is not a package: its key is the literal `<<` and
+		// the packages it contributes are declared in the referenced anchor, not
+		// in this mapping. Those merged packages already have a source map
+		// pointing at their anchor definition (set when the decoder resolved the
+		// alias), and their key nodes don't exist here to adjust, so skip them.
+		if v.Key.IsMergeKey() {
+			continue
+		}
 		key, ok := v.Key.(*ast.StringNode)
 		if !ok {
 			return goerrors.New("expected string key for package dependency")
 		}
-
-		// Handle case where a package constraint is specified:
-		// e.g.
-		//   pkg-name:
-		//	 	version: [">=1.0.0"]
-		// In this case, the the line number would be pointing at the line below `pkg-name`
-		// However, we want to include the package name itself in the source map.
+		pc := result[key.Value]
+		if pc._sourceMap == nil {
+			continue
+		}
+		// The value can be an alias whose anchor is declared above this key (e.g.
+		// `somepkg: *c`), so the entry may already start above the key line. Only
+		// pull the start line up to the key when the value node sits below it, so
+		// the name itself is included without discarding an earlier anchor
+		// position. _sourceMap is shared with the map entry by pointer, so
+		// mutating it in place needs no write-back.
 		if pc._sourceMap.pos.Start.Line > int32(key.GetToken().Position.Line) {
 			pc._sourceMap.pos.Start.Line = int32(key.GetToken().Position.Line)
 		}
-		result[key.Value] = pc
 	}
 
 	*l = result
--- a/deps_test.go
+++ b/deps_test.go
@@ -1,0 +1,279 @@
+package dalec
+
+import (
+	"strings"
+	"testing"
+
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
+)
+
+func TestPackageDependencyAliasResolution(t *testing.T) {
+	t.Parallel()
+
+	t.Run("an alias referenced as a package's value resolves to the anchor's constraints", func(t *testing.T) {
+		t.Parallel()
+		spec := loadDepsSpec(t, `
+x-c: &c
+  version: [">= 1.0"]
+dependencies:
+  build:
+    somepkg: *c
+`)
+		build := spec.Dependencies.Build
+		assert.Assert(t, cmp.Contains(build, "somepkg"))
+		assert.DeepEqual(t, build["somepkg"].Version, []string{">= 1.0"})
+	})
+
+	t.Run("a merge key inside a package map merges the anchor's packages", func(t *testing.T) {
+		t.Parallel()
+		spec := loadDepsSpec(t, `
+x-b: &b
+  somepkg:
+    version: [">= 1.0"]
+dependencies:
+  build:
+    <<: *b
+    gnupg2:
+`)
+		build := spec.Dependencies.Build
+		assert.Assert(t, cmp.Contains(build, "somepkg"))
+		assert.Assert(t, cmp.Contains(build, "gnupg2"))
+		assert.DeepEqual(t, build["somepkg"].Version, []string{">= 1.0"})
+	})
+
+	t.Run("a merge key at the dependencies level merges the anchor's lists", func(t *testing.T) {
+		t.Parallel()
+		spec := loadDepsSpec(t, `
+x-d: &d
+  build:
+    somepkg:
+      version: [">= 1.0"]
+dependencies:
+  <<: *d
+`)
+		build := spec.Dependencies.Build
+		assert.Assert(t, cmp.Contains(build, "somepkg"))
+		assert.DeepEqual(t, build["somepkg"].Version, []string{">= 1.0"})
+	})
+
+	t.Run("an explicit package overrides a merged package of the same name", func(t *testing.T) {
+		t.Parallel()
+		spec := loadDepsSpec(t, `
+x-b: &b
+  somepkg:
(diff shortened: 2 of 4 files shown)


spec_test.go
```diff
@@ -10,6 +10,42 @@ import (
 	"gotest.tools/v3/assert/cmp"
 )
 
+func TestBuildStepListAliasResolution(t *testing.T) {
```
invidian, line 13: nit: all should probably run in parallel (as default).

deps_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
+)
+
+func TestPackageDependencyAliasResolution(t *testing.T) {
```
invidian, line 11: Ditto parallel.

deps_test.go
```diff
@@ -0,0 +1,173 @@
...
+  runtime:
+    somepkg: *c
+`)
+		runtime := spec.Dependencies.Runtime
+		assert.Assert(t, cmp.Contains(runtime, "somepkg"))
+		assert.DeepEqual(t, runtime["somepkg"].Version, []string{">= 1.0"})
+	})
+}
+
+func TestPackageDependencyListSequenceForm(t *testing.T) {
```
invidian, line 91: nit: some extra test cases to eliminate few mutants:
```diff
diff --git deps_test.go deps_test.go
index 6bbe283c..9806c9ea 100644
--- deps_test.go
+++ deps_test.go
@@ -95,6 +95,55 @@ dependencies:
 	})
 }

+func TestPackageDependencyListSequenceFormSourceMap(t *testing.T) {
+	t.Run("a package listed in sequence form keeps a source map pointing at its list entry", func(t *testing.T) {
+		doc := depsSpecHeader + `
+dependencies:
+  build:
+    - somepkg
+`
+		spec, err := LoadSpec([]byte(doc))
+		assert.NilError(t, err)
+
+		pc := spec.Dependencies.Build["somepkg"]
+		assert.Assert(t, pc._sourceMap != nil)
+		assert.Equal(t, int(pc._sourceMap.pos.Start.Line), lineOf(doc, "- somepkg"))
+	})
+}
+
+func TestPackageDependencyEmptySequence(t *testing.T) {
+	t.Run("an empty dependency sequence leaves the list unset rather than an empty map", func(t *testing.T) {
+		spec := loadDepsSpec(t, `
+dependencies:
+  build: []
+`)
+		assert.Assert(t, spec.Dependencies.Build == nil)
+	})
+}
+
+func TestPackageDependencyDecodeErrors(t *testing.T) {
+	t.Run("a package constraint that cannot be decoded surfaces an error", func(t *testing.T) {
+		err := loadDepsSpecErr(t, `
+dependencies:
+  build:
+    somepkg:
+      version:
+        not: a-list
+`)
+		assert.ErrorContains(t, err, "unmarshal package constraints")
+	})
+
+	t.Run("a non-string package key surfaces an error", func(t *testing.T) {
+		err := loadDepsSpecErr(t, `
+dependencies:
+  build:
+    123:
+      version: [">= 1.0"]
+`)
+		assert.ErrorContains(t, err, "expected string key")
+	})
+}
+
 func TestPackageDependencySourceMaps(t *testing.T) {
 	// Line numbers are derived from the document text below so the assertions
 	// are not tied to a hand-counted offset. `merged-pkg` is declared inside the
@@ -138,13 +187,9 @@ dependencies:
 	})
 }

-// loadDepsSpec loads a spec composed of a minimal valid header plus the given
-// body (typically an anchor and a dependencies block) and fails the test if it
-// cannot be loaded.
-func loadDepsSpec(t *testing.T, body string) *Spec {
-	t.Helper()
-
-	const header = `
+// depsSpecHeader is a minimal valid spec header that the dependency tests prefix
+// to a dependencies block so the document loads.
+const depsSpecHeader = `
 name: t
 version: "1.0"
 revision: "1"
@@ -153,12 +198,28 @@ vendor: t
 license: MIT
 description: t
 `
-	spec, err := LoadSpec([]byte(header + body))
+
+// loadDepsSpec loads a spec composed of a minimal valid header plus the given
+// body (typically an anchor and a dependencies block) and fails the test if it
+// cannot be loaded.
+func loadDepsSpec(t *testing.T, body string) *Spec {
+	t.Helper()
+
+	spec, err := LoadSpec([]byte(depsSpecHeader + body))
 	assert.NilError(t, err)
 	assert.Assert(t, spec.Dependencies != nil)
 	return spec
 }

+// loadDepsSpecErr loads a spec composed of the minimal header plus body and
+// returns the resulting error (expected to be non-nil for the caller).
+func loadDepsSpecErr(t *testing.T, body string) error {
+	t.Helper()
+
+	_, err := LoadSpec([]byte(depsSpecHeader + body))
+	return err
+}
+
 // lineOf returns the 1-based line number of the first line in doc that contains
 // substr, or 0 if it is not found.
 func lineOf(doc, substr string) int {
```

deps.go
```diff
@@ -86,32 +90,50 @@ func (l *PackageDependencyList) UnmarshalYAML(ctx context.Context, node ast.Node
...
 			return goerrors.New("expected string key for package dependency")
 		}
-
-		// Handle case where a package constraint is specified:
-		// e.g.
-		//   pkg-name:
-		//	 	version: [">=1.0.0"]
-		// In this case, the the line number would be pointing at the line below `pkg-name`
-		// However, we want to include the package name itself in the source map.
+		pc, ok := result[key.Value]
```
invidian: nit: one less variable and small simplification based on mutation tests from Copilot.
```diff
diff --git deps.go deps.go
index 66a52900..807324c3 100644
--- deps.go
+++ deps.go
@@ -126,14 +126,15 @@ func (l *PackageDependencyList) UnmarshalYAML(ctx context.Context, node ast.Node
                if !ok {
                        return goerrors.New("expected string key for package dependency")
                }
-               pc, ok := result[key.Value]
-               if !ok || pc._sourceMap == nil {
+               pc := result[key.Value]
+               if pc._sourceMap == nil {
                        continue
                }
-               if pc._sourceMap.pos.Start.Line > int32(key.GetToken().Position.Line) {
-                       pc._sourceMap.pos.Start.Line = int32(key.GetToken().Position.Line)
-                       result[key.Value] = pc
-               }
+               // The value node never precedes its key in the document, so the entry's
+               // start line is always at or below the key line; pulling it up to the key
+               // line is always the wanted result. _sourceMap is a pointer shared with
+               // the map entry, so mutating it in place needs no write-back.
+               pc._sourceMap.pos.Start.Line = int32(key.GetToken().Position.Line)
        }

        *l = result
```

invidian (approved):
It's hardly my domain, but patches looks reasonable. I left few comments with test improvements suggestions, but they are pretty optional.

spec_test.go
cpuguy83: Done, added t.Parallel() to the func and both subtests.

deps_test.go
cpuguy83: Done, all deps tests run in parallel now.

deps_test.go
cpuguy83: Added the sequence-form source-map, empty-sequence, and decode-error cases, plus the depsSpecHeader const and loadDepsSpecErr helper. Thanks.

deps.go
cpuguy83: Adopted the ok-var removal and dropped the redundant write-back since _sourceMap is shared by pointer. Kept the guard though: for an alias used as a value (somepkg: *c) the entry can start above its key line, so pulling up unconditionally would move it off the anchor. Added TestPackageDependencyAliasValueSourceMap to pin that.

PR #1140, opened 2026-07-13
targets: support BuildKit proxy package installs

### LINKED ISSUE #1114: [REQ] add support for buildkit proxy mode

None

Buildkit 0.31 added a new "proxy" mode which forces all egress traffic through an HTTP(S) proxy.
The traffic that flows through the proxy is recorded in provenance.
HTTP(S)_PROXY envs and root CA's for TLS termination are injected into the container env.

- https://github.com/moby/buildkit/pull/6740
- https://github.com/moby/buildkit/pull/6816

Some definite places where we can add proxy mode:

- go/pip/rust module fetches
- dnf install
- apt install (requires setting apt/aptitude configs)

Notably npm, IIRC, requires configuring the CA pool manually as it doesn't use the system cert pool.

Another potential case is for sandbox build mode.
This can help capture egress traffic but only works for HTTP(S), all other traffic would be blocked and could cause unintended breakages.
An arg could be used to disable the proxy, perhaps?

kartikjoshi21:
Thread BuildKit build args through source options so package install helpers can detect proxy-network configuration and users can opt out with DALEC_DISABLE_PROXY_CONFIG.

Configure npm, apt, and dnf/tdnf package installs to trust the BuildKit proxy CA when proxy-network is enabled, then clean up temporary package-manager proxy config after installs.

Add focused unit coverage for the generated proxy scripts and E2E proxy-network coverage for package manager installs and sandbox build network access across Linux targets.

Fixes: #1114

### DIFF
--- a/targets/linux/deb/distro/install.go
+++ b/targets/linux/deb/distro/install.go
@@ -11,6 +11,108 @@ import (
 	"github.com/project-dalec/dalec/packaging/linux/deb"
 )
 
+const aptProxyConfigScript = `
+cleanup_apt_proxy() {
+	if [ -n "${DALEC_APT_PROXY_CONFIG_ACTIVE:-}" ]; then
+		rm -f "${DALEC_APT_PROXY_CONFIG_ACTIVE}" 2>/dev/null || true
+		if [ "${APT_CONFIG:-}" = "${DALEC_APT_PROXY_CONFIG_ACTIVE}" ]; then
+			if [ "${DALEC_APT_PROXY_CONFIG_OLD_APT_CONFIG_WAS_SET:-0}" = "1" ]; then
+				export APT_CONFIG="${DALEC_APT_PROXY_CONFIG_OLD_APT_CONFIG}"
+			else
+				unset APT_CONFIG
+			fi
+		fi
+		unset DALEC_APT_PROXY_CONFIG_ACTIVE
+		unset DALEC_APT_PROXY_CONFIG_OLD_APT_CONFIG
+		unset DALEC_APT_PROXY_CONFIG_OLD_APT_CONFIG_WAS_SET
+	fi
+}
+
+configure_apt_proxy() {
+	restore_xtrace=0
+	case "$-" in
+		*x*) set +x; restore_xtrace=1 ;;
+	esac
+
+	if [ "${DALEC_DISABLE_PROXY_CONFIG:-}" = "1" ]; then
+		if [ "${restore_xtrace}" = "1" ]; then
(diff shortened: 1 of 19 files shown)


test/testenv/buildx.go
```diff
@@ -489,9 +495,17 @@ func (b *BuildxEnv) RunTest(ctx context.Context, t *testing.T, f TestFunc, opts
 	defer cancelOutput()
 
 	ch := make(chan *client.SolveStatus)
-	go fowardToSolveStatusFn(ctx, ch, statusFn, cfg.SolveStatusFn)
+	var wg sync.WaitGroup
+	wg.Add(1)
```
invidian: nit: maybe use `wg.Go` for less code.

targets/linux/deb/distro/install.go
```diff
@@ -20,11 +122,15 @@ import (
 func AptInstall(packages []string, opts ...llb.ConstraintsOpt) llb.RunOption {
 	return dalec.RunOptFunc(func(ei *llb.ExecInfo) {
 		const installScript = `#!/usr/bin/env sh
-set -ex
+	set -ex
```
invidian: Some unnecessary whitespace sneaked in.

invidian (commented):
LGTM, just some nits. I'll let @cpuguy83 do a final approval, since maybe there is some code missing beyond this which I'm not aware of, since I'm not familiar with BuildKit proxy.

invidian (approved):

cpuguy83:
I need to understand this better.
Originally the proxy stuff was opt-in via network mode, which you can set per llb ExecOp.
Now it seems like its per build which is a little more problematic.
Doing some extra research here.

cpuguy83:
Ah ok. I see now. We are making it so dalec has a chance of working when proxy mode is enabled by the client/daemon.
We can't choose which step is run in the mode.
This makes sense.

Overall LGTM but there are conflicts.
Docs might need some extra explanation on what is meant by "proxy".

kartikjoshi21:
> Ah ok. I see now. We are making it so dalec has a chance of working when proxy mode is enabled by the client/daemon. We can't choose which step is run in the mode. This makes sense.
> 
> Overall LGTM but there are conflicts. Docs might need some extra explanation on what is meant by "proxy".

I’ve rebased the branch onto main and resolved the conflicts and  expanded the documentation to clarify that proxy mode is enabled at the BuildKit client/daemon solve level, not selected per Dalec step.

cpuguy83 (approved):

PR #1141, opened 2026-07-15
generator/gomod: support GOPROXY build arg

### LINKED ISSUE #1072: [REQ] Support injecting custom GOPROXY into gomod generator

None

We'd like to be able to use a custom `GOPROXY` in the gomod generator so that it can download mods using our own proxy instead of the default one.
This should be able to use build-args like we do for other injectable values and not require it to appear in the spec file.

kartikjoshi21:
Add DALEC_GOMOD_PROXY as a known frontend build arg and carry it through SourceOpts into gomod dependency generation and gomod patch preprocessing.

Fixes: #1072

### DIFF
--- a/frontend/gateway_test.go
+++ b/frontend/gateway_test.go
@@ -1,0 +1,20 @@
+package frontend
+
+import (
+	"testing"
+
+	"github.com/project-dalec/dalec"
+)
+
+func TestSourceOptFromUIClientReadsGomodProxyBuildArgIntoExtraEnvs(t *testing.T) {
+	t.Parallel()
+
+	const proxy = "http://proxy.example:5000,direct"
+	client := newStubClient()
+	client.opts["build-arg:"+dalec.BuildArgDalecGomodProxy] = proxy
+
+	sOpt := SourceOptFromUIClient(t.Context(), client, nil, nil)
+	if sOpt.ExtraEnvs["GOPROXY"] != proxy {
+		t.Fatalf("expected GOPROXY extra env %q, got %q", proxy, sOpt.ExtraEnvs["GOPROXY"])
+	}
+}
--- a/generator_gomod.go
+++ b/generator_gomod.go
@@ -19,6 +19,9 @@ const (
 	// GoModCacheKey is the key used to identify the go module cache in the buildkit cache.
 	// It is exported only for testing purposes.
 	GomodCacheKey = "dalec-gomod-proxy-cache"
+	// BuildArgDalecGomodProxy is the frontend build arg used to override GOPROXY
+	// for gomod dependency generation.
+	BuildArgDalecGomodProxy = "DALEC_GOMOD_PROXY"
 )
 
 func (g *GeneratorGomod) processBuildArgs(args map[string]string, allowArg func(key string) bool) error {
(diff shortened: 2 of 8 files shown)


preprocess.go
```diff
@@ -220,7 +220,7 @@ func buildGomodPatchEnv(editArgs string, paths []string, gen *SourceGenerator, s
 // generateGomodPatchStateForSource generates a single merged patch LLB state for all paths
 // in a gomod generator by running go mod edit + tidy and capturing the diff.
 // Returns the LLB state containing the patch file, or nil if no changes are needed.
-func (s *Spec) generateGomodPatchStateForSource(sourceName string, gen *SourceGenerator, baseState llb.State, worker llb.State, credHelper llb.RunOption, opts ...llb.ConstraintsOpt) (*llb.State, error) {
+func (s *Spec) generateGomodPatchStateForSource(sourceName string, gen *SourceGenerator, baseState llb.State, worker llb.State, credHelper llb.RunOption, gomodProxy string, opts ...llb.ConstraintsOpt) (*llb.State, error) {
```
invidian: Rather than adding yet another argument to the long list, can we extract them into a struct?

source.go
```diff
@@ -155,6 +155,10 @@ type SourceOpts struct {
 	TargetPlatform   *ocispecs.Platform
 	GitCredHelperOpt func() (llb.RunOption, error)
 	SourceFilter     func() (SourceFilterConfig, error)
+	// GomodProxy, when set, overrides GOPROXY while gomod generators download
```
invidian: This doesn't seem like a right place for this value. All fields in `SourceOpts` are generic, maybe apart from `GitCredHelperOpt`, while this is very specific. I think this needs some different (or more generic) solution.

generator_gomod.go
```diff
@@ -63,7 +66,7 @@ func (s *Spec) HasGomods() bool {
 	return false
 }
 
-func withGomod(g *SourceGenerator, srcSt, worker llb.State, subPath string, credHelper llb.RunOption, opts ...llb.ConstraintsOpt) func(llb.State) llb.State {
+func withGomod(g *SourceGenerator, srcSt, worker llb.State, subPath string, credHelper llb.RunOption, gomodProxy string, opts ...llb.ConstraintsOpt) func(llb.State) llb.State {
```
invidian: ditto args.

invidian (changes requested):

source.go
cpuguy83: Could maybe do something like `ExtraEnvs map[string]string` and lookup `GOPROXY` in the generator?

preprocess.go
kartikjoshi21: I moved the gomod-specific inputs into an internal gomodGeneratorOpts struct and pass that into generateGomodPatchStateForSource

frontend/gateway_test.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func TestSourceOptFromUIClientReadsGomodProxyBuildArgIntoExtraEnvs(t *testing.T) {
+	t.Parallel()
+
+	const proxy = "http://proxy.example:5000,direct"
+	client := newStubClient()
+	client.opts["build-arg:"+dalec.BuildArgDalecGomodProxy] = proxy
+
+	sOpt := SourceOptFromUIClient(context.Background(), client, nil, nil)
```
invidian: nit: could use t.Context().

invidian (approved):
One small nit, otherwise LGTM

invidian (approved):

cpuguy83 (approved):

PR #1142, opened 2026-07-15
Add /opt artifact support

chmill-zz:

Adds /opt as a supported artifact destination, allowing packages to install files into layouts such as /opt/<package>/bin or /opt/bin

Fixes #

### DIFF
--- a/packaging/linux/deb/debroot.go
+++ b/packaging/linux/deb/debroot.go
@@ -582,12 +582,10 @@ func createInstallScripts(worker llb.State, spec *dalec.Spec, dir, target string
 		}
 	}
 
-	if len(artifacts.Opt) > 0 {
-		for key, cfg := range dalec.SortedMapIter(artifacts.Opt) {
-			resolved := cfg.ResolveName(key)
-			targetDir := filepath.Join(OptPath, cfg.SubPath)
-			writeInstall(key, targetDir, resolved)
-		}
+	for key, cfg := range dalec.SortedMapIter(artifacts.Opt) {
+		resolved := cfg.ResolveName(key)
+		targetDir := filepath.Join(OptPath, cfg.SubPath)
+		writeInstall(key, targetDir, resolved)
 	}
 
 	if len(artifacts.Libs) > 0 {
--- a/packaging/linux/deb/debroot.go
+++ b/packaging/linux/deb/debroot.go
@@ -31,6 +31,7 @@ const (
 	DocsPath                  = "/usr/share/doc"
 	LibsPath                  = "/usr/lib"
 	LibexecPath               = "/usr/libexec"
+	OptPath                   = "/opt"
 	DataDirsPath              = "/usr/share"
 )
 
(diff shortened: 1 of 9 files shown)


chmill-zz:
The three failed checks appear to be transient infrastructure failures rather than failures caused by this change:

- `integration (Bionic)`
- `integration (Focal)`
- `integration (Jammy)`

All three failed in the `cross_platform` subtest while building for ARM. `apt` timed out connecting to `ports.ubuntu.com:80`, leaving the package indexes unavailable and causing the subsequent build-dependency installation to fail with exit code 100. The other integration targets, unit tests, and lint checks passed, and these Ubuntu jobs have passed on neighboring CI runs.

Could a maintainer please rerun the failed jobs?

packaging/linux/deb/debroot.go
```diff
@@ -580,6 +582,16 @@ func createInstallScripts(worker llb.State, spec *dalec.Spec, dir, target string
 		}
 	}
 
+	if len(artifacts.Opt) > 0 {
+		sorted := dalec.SortMapKeys(artifacts.Opt)
```
cpuguy83: Can you use the iter form? https://pkg.go.dev/github.com/project-dalec/dalec?utm_source=godoc#SortedMapIter

packaging/linux/rpm/template.go
```diff
@@ -763,6 +768,13 @@ func (w *specWrapper) getArtifactOwnership() string {
 			setArtifactOwnership(`/%{_bindir}`, p, &cfg)
 		}
 	}
+	if artifacts.Opt != nil {
+		optKeys := dalec.SortMapKeys(artifacts.Opt)
```
cpuguy83: same, re: use iter

packaging/linux/rpm/template.go
```diff
@@ -971,6 +990,14 @@ func (w *specWrapper) Install() fmt.Stringer {
 		}
 	}
 
+	if artifacts.Opt != nil {
+		optFileKeys := dalec.SortMapKeys(artifacts.Opt)
```
cpuguy83: same, re: use iter

cpuguy83 (changes requested):
Can you add a case for opt in integration/linux_target_test.go where we test other artifacts?

packaging/linux/deb/debroot.go
```diff
@@ -580,6 +582,14 @@ func createInstallScripts(worker llb.State, spec *dalec.Spec, dir, target string
 		}
 	}
 
+	if len(artifacts.Opt) > 0 {
```
invidian: Do we need len check here? Shouldn't iterator just do nothing when slice is empty?

packaging/linux/rpm/template.go
```diff
@@ -763,6 +768,11 @@ func (w *specWrapper) getArtifactOwnership() string {
 			setArtifactOwnership(`/%{_bindir}`, p, &cfg)
 		}
 	}
+	if artifacts.Opt != nil {
```
invidian: Ditto conditon.

packaging/linux/rpm/template.go
invidian: We should probably have a helper function to avoid so much repetition for all this different fields.

test/linux_target_test.go
```diff
@@ -3420,7 +3420,7 @@ func Value() string {
 		})
 	})
 
-	t.Run("test libexec file installation", func(t *testing.T) {
+	t.Run("test libexec and opt file installation", func(t *testing.T) {
```
invidian: Why do we add to this test instead of a separate one?

invidian (commented):
Overall LGTM, I have some questions/comments.

invidian (approved):

cpuguy83 (approved):

PR #1146, opened 2026-07-16
Add supplemental package model

cpuguy83:
Adds target-level supplemental package metadata, validation, build argument expansion, defaults, and generated schema support.

This is layer 1 of the native supplemental-packages stack and provides the shared model used by later target implementations.

Related to #607.

### DIFF
--- a/subpackage.go
+++ b/subpackage.go
@@ -1,0 +1,215 @@
+package dalec
+
+import (
+	goerrors "errors"
+	"fmt"
+	"iter"
+
+	"github.com/moby/buildkit/frontend/dockerfile/shell"
+	"github.com/pkg/errors"
+)
+
+// SubPackage defines a supplemental package produced from the same build
+// output as the primary package. Each supplemental package has its own
+// artifact selection, runtime dependencies, and package metadata.
+//
+// Supplemental packages share the primary package's build steps, sources,
+// version, revision, license, vendor, and website. They cannot override
+// these fields.
+type SubPackage struct {
+	// Name overrides the default package name.
+	// By default, the package name is "<parent>-<key>" where <key> is the map
+	// key under which this SubPackage is defined. Set this to use a fully custom
+	// package name instead. Build argument references are not substituted in this field.
+	Name string `yaml:"name,omitempty" json:"name,omitempty"`
+
+	// Description is the package description. This is required — both RPM and
+	// Debian require a description/summary for every subpackage.
+	// Build argument references are not substituted in this field.
+	Description string `yaml:"description" json:"description" jsonschema:"required"`
+
+	// Artifacts specifies which build outputs go into this supplemental package.
+	// This is self-contained — no artifacts are inherited from the primary package.
+	Artifacts *Artifacts `yaml:"artifacts,omitempty" json:"artifacts,omitempty"`
+
+	// Dependencies specifies runtime dependencies for this supplemental package.
+	// Only runtime and recommends dependencies are allowed; build dependencies
+	// are shared with the primary package.
+	Dependencies *SubPackageDependencies `yaml:"dependencies,omitempty" json:"dependencies,omitempty"`
+
+	// Conflicts is the list of packages that conflict with this supplemental package.
+	Conflicts PackageDependencyList `yaml:"conflicts,omitempty" json:"conflicts,omitempty"`
+
+	// Provides is the list of things this supplemental package provides.
+	Provides PackageDependencyList `yaml:"provides,omitempty" json:"provides,omitempty"`
+
+	// Replaces is the list of packages that this supplemental package replaces/obsoletes.
+	Replaces PackageDependencyList `yaml:"replaces,omitempty" json:"replaces,omitempty"`
+}
+
+// ResolvedName returns the package name that this SubPackage will produce.
+// If [SubPackage.Name] is set, it is returned as-is.
+// Otherwise, the name is "<parentName>-<mapKey>".
+func (s *SubPackage) ResolvedName(parentName, mapKey string) string {
+	if s.Name != "" {
+		return s.Name
+	}
+	return parentName + "-" + mapKey
+}
+
+// SubPackageDependencies contains only the dependency fields valid for
+// supplemental packages. Build dependencies are shared with the primary
+// package and cannot be overridden.
+type SubPackageDependencies struct {
+	// Runtime is the list of packages required to install/run the supplemental package.
+	Runtime PackageDependencyList `yaml:"runtime,omitempty" json:"runtime,omitempty"`
+	// Recommends is the list of packages recommended to install with the supplemental package.
+	// Note: Not all package managers support this (e.g. rpm)
+	Recommends PackageDependencyList `yaml:"recommends,omitempty" json:"recommends,omitempty"`
+}
+
+func (d *SubPackageDependencies) GetRuntime() PackageDependencyList {
+	if d == nil {
+		return nil
+	}
+	return d.Runtime
+}
+
+func (d *SubPackageDependencies) GetRecommends() PackageDependencyList {
+	if d == nil {
+		return nil
+	}
+	return d.Recommends
+}
+
+func (d *SubPackageDependencies) processBuildArgs(lex *shell.Lex, args map[string]string, allowArg func(string) bool) error {
+	if d == nil {
+		return nil
+	}
+
+	var errs []error
+	for k, v := range d.Runtime {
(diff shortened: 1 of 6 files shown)


target.go
```diff
@@ -73,6 +79,12 @@ func (t *Target) validate() error {
 		errs = append(errs, errors.Wrap(err, "postinstall"))
 	}
 
+	for key, pkg := range t.Packages {
+		if err := pkg.validate(); err != nil {
+			errs = append(errs, errors.Wrapf(err, "package %s", key))
```
invidian: nit: could use `%q` to guard against empty package names producing weird errors.

subpackage.go
```diff
@@ -0,0 +1,278 @@
...
+			}
+			v.Version[i] = updated
+		}
+		d.Recommends[k] = v
+	}
+
+	return goerrors.Join(errs...)
+}
+
+// SubPackage defines a supplemental package produced from the same build
```
invidian: Maybe it could be defined on top of file, it feels more substantial than `SubPackageDependencies`.

subpackage.go
```diff
@@ -0,0 +1,278 @@
...
+	if err := validateNoBuildArgReferences(s.Name); err != nil {
+		errs = append(errs, errors.Wrap(err, "name"))
+	}
+
+	if err := validateNoBuildArgReferences(s.Description); err != nil {
+		errs = append(errs, errors.Wrap(err, "description"))
+	}
+
+	if s.Artifacts != nil {
+		if s.Artifacts.DisableStrip {
```
invidian, line 136: This looks like mismatched types. Perhaps most fields should be moved to something like `BaseArtifacts` and `RootArtifacts` or `MainArtifacts` should additionally include DisableStrip field? Similar to `SubPackageDependencies`.

subpackage.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		if err := s.Artifacts.validate(); err != nil {
+			errs = append(errs, errors.Wrap(err, "artifacts"))
+		}
+	}
+
+	return goerrors.Join(errs...)
+}
+
+func (s *SubPackage) processBuildArgs(lex *shell.Lex, args map[string]string, allowArg func(string) bool) error {
```
invidian, line 147: Do we have some documentation on build args? Just looking at the code, build args seems to be related to commands which one runs to build the package, but I'm not sure that's the right thing.

subpackage.go
```diff
@@ -0,0 +1,278 @@
...
+	var errs []error
+
+	if err := validateNoBuildArgReferences(s.Name); err != nil {
```
invidian: I'm pretty sure we are missing a lot of test cases with examples when those validations fail.

subpackage_test.go
```diff
@@ -0,0 +1,652 @@
...
+				},
+			},
+			expectErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			err := tt.pkg.validate()
```
invidian: Can we test validations at exported code level?

subpackage_test.go
```diff
@@ -0,0 +1,652 @@
...
+					assert.Check(t, cmp.ErrorContains(err, tt.errSubstr))
+				}
+			} else {
+				assert.NilError(t, err)
+			}
+		})
+	}
+}
+
+func TestSubPackageYAMLRoundTrip(t *testing.T) {
```
invidian: What does it mean round trip here?

subpackage.go
```diff
@@ -0,0 +1,278 @@
...
+	t, ok := s.Targets[targetKey]
+	if !ok {
+		return nil
+	}
+	return t.Packages
+}
+
+// GetAllPackageNames returns the resolved names of all packages (primary +
+// supplemental) for the given target. The primary package name is always first.
+func (s *Spec) GetAllPackageNames(targetKey string) []string {
```
invidian: This appears unused.

subpackage.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		seen[resolved] = key
+	}
+
+	return goerrors.Join(errs...)
+}
+
+// GetSubPackages returns the supplemental packages defined for the given target.
+// Returns nil if the target does not exist or has no supplemental packages.
+func (s *Spec) GetSubPackages(targetKey string) map[string]SubPackage {
```
invidian: Can we place exported methods higher up? 

And assuming we place this method in this file for modularity, shouldn't we change this signature to `GetSubPackages(Spec, string) ...` perhaps?

Also, it could be named `GetSubPackagesForTarget`.

subpackage.go
```diff
@@ -0,0 +1,278 @@
...
+			return false
+		}
+		if i > 0 && r != '_' && !unicode.IsLetter(r) && !unicode.IsDigit(r) {
+			return false
+		}
+	}
+	return value != ""
+}
+
+func (s *SubPackage) fillDefaults() {
```
invidian: That should probably be removed together with calling code.

invidian (commented):
Thanks for splitting the PR. That certainly makes it easier to review.

subpackage.go
cpuguy83: Yeah. On the last push I asked the clanker to remove arg substitution for name and description here (doesn't make sense and complicates other pieces).
It took that to mean "make sure there are no arg substitutions".
I don't think we should be validating this at all.

subpackage.go
cpuguy83: Given this is the only divergence, and may not even be permanent, I would personally prefer having the runtime validation (and documentation) that this field is not supported in subpackages rather than dealing with managing multiple types, embedding, whatever.

subpackage.go
cpuguy83: build-args are a docker-ism.
I imagine its just assumed in our code documentation as to what it is. I suppose it could use a formal definition.

subpackage.go
invidian: Sure. Maybe we can leave a comment then, to mark that this is a deliberate decision?

cpuguy83: That's what the error is?

invidian: No, I mean that given that this is only divergence, we deliberately violate "Make Invalid States Unrepresentable" rule and rely on runtime validation to avoid managing multiple types.

cpuguy83:
Comments addressed.

subpackage.go
cpuguy83: Done.

subpackage.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// GetSubPackagesForTarget returns the supplemental packages defined for the
+// given target in map-key order. A missing target or a target without
+// supplemental packages yields no values.
+func GetSubPackagesForTarget(spec *Spec, target string) iter.Seq2[string, SubPackage] {
+	return SortedMapIter(spec.Targets[target].Packages)
+}
+
+// ResolvedName returns the package name that this SubPackage will produce.
```
invidian: nit: maybe could be placed below `type SubPackage struct { ... }` for class-like definition, rather than being scattered and intertwined with free functions and `SubPackageDependencies`.

subpackage.go
```diff
@@ -0,0 +1,215 @@
...
+	var errs []error
+
+	if s.Description == "" {
+		errs = append(errs, fmt.Errorf("description is required"))
+	}
+
+	if s.Artifacts != nil {
+		// Subpackages deliberately reuse the root Artifacts public model. DisableStrip
+		// is currently its sole unsupported field, so accepting one representable
+		// invalid state avoids maintaining nearly identical public types.
```
invidian, line 135: I like the comment, thanks!

subpackage_test.go
```diff
@@ -0,0 +1,440 @@
...
+	})
+}
+
+func TestSubPackageDependenciesNilSafe(t *testing.T) {
+	t.Parallel()
+
+	var d *SubPackageDependencies
+	assert.Check(t, d.GetRuntime() == nil)
+	assert.Check(t, d.GetRecommends() == nil)
+	assert.NilError(t, d.processBuildArgs(nil, nil, nil))
```
invidian: nit: if we would drop/test that differently, we could make this file blackbox test with dalec_test package.

invidian (approved):

subpackage_test.go
cpuguy83: Done.

invidian (approved):

invidian (approved):

PR #1147, opened 2026-07-16
Generate supplemental RPM packages

cpuguy83:
Adds RPM subpackage sections, artifact selection, and scriptlets, with template and LLB-level coverage.

This is layer 2 of the native supplemental-packages stack and depends on #1146.

Related to #607.

### DIFF
--- a/packaging/linux/rpm/buildroot_test.go
+++ b/packaging/linux/rpm/buildroot_test.go
@@ -1,0 +1,351 @@
+package rpm
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/moby/buildkit/client/llb"
+	"github.com/project-dalec/dalec"
+	"github.com/project-dalec/dalec/internal/test"
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
+)
+
+func TestRPMSSpec_Buildroot_respects_directory_arg(t *testing.T) {
+	t.Parallel()
+
+	t.Run("when the directory specified is left empty", func(t *testing.T) {
+		t.Parallel()
+		spec := baseSpec("myapp")
+
+		state := RPMSpec(spec, llb.Scratch(), "", "")
+
+		t.Run("the build root is put into /", func(t *testing.T) {
+			t.Parallel()
+			ops := test.LLBOpsFromState(t.Context(), t, state)
+
+			dirPath := findMkdir(t, ops)
+			assert.Equal(t, dirPath, "/SPECS/myapp")
+
+			filePath, _ := findSpecMkfile(t, ops)
+			assert.Equal(t, filePath, "/SPECS/myapp/myapp.spec")
+		})
+	})
+
+	t.Run("when the directory specified is non-empty", func(t *testing.T) {
+		t.Parallel()
+		spec := baseSpec("myapp")
+		state := RPMSpec(spec, llb.Scratch(), "", "custom/dir")
+
+		t.Run("the build root is put into that directory", func(t *testing.T) {
+			t.Parallel()
+			ops := test.LLBOpsFromState(t.Context(), t, state)
+
+			dirPath := findMkdir(t, ops)
+			assert.Equal(t, dirPath, "/custom/dir")
+
+			filePath, _ := findSpecMkfile(t, ops)
+			assert.Equal(t, filePath, "/custom/dir/myapp.spec")
+		})
+	})
+}
+
+func TestRPMSpec_Subpackages(t *testing.T) {
+	t.Parallel()
+
+	t.Run("when a spec has no subpackages", func(t *testing.T) {
+		t.Parallel()
+		spec := baseSpec("foo")
+
+		state := RPMSpec(spec, llb.Scratch(), "azlinux3", "")
+		ops := test.LLBOpsFromState(t.Context(), t, state)
+
+		t.Run("then the rpm spec does not have subpackage markers", func(t *testing.T) {
+			t.Parallel()
+			_, data := findSpecMkfile(t, ops)
+			assert.Assert(t, !strings.Contains(data, "%package -n"))
+			assert.Assert(t, !strings.Contains(data, "%description -n"))
+			assert.Assert(t, !strings.Contains(data, "%files -n"))
+		})
+	})
+
+	t.Run("given a spec has subpackages", func(t *testing.T) {
+		t.Parallel()
+
+		t.Run("when the subpackage uses the default naming", func(t *testing.T) {
+			t.Parallel()
+
+			spec := baseSpec("foo")
+			spec.Targets = map[string]dalec.Target{
+				"azlinux3": {
+					Packages: map[string]dalec.SubPackage{
+						"debug": {
+							Description: "Debug symbols for foo",
+							Artifacts: &dalec.Artifacts{
+								Binaries: map[string]dalec.ArtifactConfig{
+									"foo-debug": {},
+								},
+							},
+						},
+					},
+				},
+			}
+
+			t.Run("the rpm spec has subpackage markers matching the default name", func(t *testing.T) {
+				t.Parallel()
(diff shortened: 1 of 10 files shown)


packaging/linux/rpm/template.go
```diff
@@ -894,7 +901,28 @@ func (w *specWrapper) Install() fmt.Stringer {
...
+
+	// Install artifacts for supplemental packages (shared %install section)
+	packages := w.Spec.GetSubPackages(w.Target)
+	if len(packages) > 0 {
+		keys := dalec.SortMapKeys(packages)
+		for _, key := range keys {
+			pkg := packages[key]
+			if pkg.Artifacts != nil {
+				resolvedName := pkg.ResolvedName(w.Spec.Name, key)
+				installArtifacts(b, *pkg.Artifacts, resolvedName)
```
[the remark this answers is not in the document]
cpuguy83: DisableStrip should be invalid for sub-packages.
Fixing.

packaging/linux/rpm/template.go
```diff
@@ -1217,6 +1247,172 @@ func (w *specWrapper) DisableAutoReq() string {
...
+	}
+
+	return b, nil
+}
+
+// writePackageHeader writes the %package section with Summary and dependency fields.
+func (s *subPkgWrapper) writePackageHeader(b *strings.Builder) {
+	fmt.Fprintf(b, "%%package %s\n", s.directive())
+	fmt.Fprintf(b, "Summary: %s\n", s.Pkg.Description)
+
```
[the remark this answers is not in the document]
cpuguy83: DisableAutoRequires should be invalid for sub-packages.
Fixing.

test/subpackage_test.go
```diff
@@ -0,0 +1,323 @@
...
+		pkg.Dependencies = &dalec.SubPackageDependencies{
+			Runtime: map[string]dalec.PackageConstraints{
+				dependency: {},
+			},
+		}
+		spec.Targets[targetCfg.Key].Packages["contrib"] = pkg
+
+		testPackageOutput(ctx, t, targetCfg, spec, func(pkgFS fs.FS) {
+			packages := readSubpackagePackageMetadata(t, pkgFS, cfg)
+			assertPackageRuntimeDependency(t, packages, spec.Name+"-contrib", dependency)
```
invidian: This is only called once, so I'd inline it. Tests should be as easy to read as possible, without having to jump between test code and helpers. Helpers should not be relevant to the test content, while this hides what's being asserted on exactly.

test/subpackage_test.go
```diff
@@ -0,0 +1,323 @@
...
+func assertPackageRuntimeDependency(t *testing.T, packages map[string]subpackagePackageMetadata, packageName, dependency string) {
+	t.Helper()
+
+	pkg, ok := packages[packageName]
+	assert.Assert(t, ok, "supplemental package %q was not found: %v", packageName, packages)
+	_, ok = pkg.RuntimeDependencies[dependency]
+	assert.Assert(t, ok, "package %q does not require %q: %v", packageName, dependency, pkg.RuntimeDependencies)
+}
+
+func assertPackageArtifacts(
```
invidian: _probably_ fine with helper here, but I'd add a struct for arguments, so it's obvious what we pass around and assert on, otherwise it's tough to read.

test/subpackage_test.go
```diff
@@ -0,0 +1,323 @@
...
+	ArtifactNameMatches func(string, string) bool
+}
+
+type subpackagePackageMetadata struct {
+	Name                string
+	Path                string
+	RuntimeDependencies map[string]struct{}
+}
+
+var rpmSubpackageTests = &subpackageTestConfig{
```
invidian: nit: this could be function to avoid using global variables without good justification?

packaging/linux/rpm/buildroot_test.go
```diff
@@ -0,0 +1,325 @@
...
+	"testing"
+
+	"github.com/moby/buildkit/client/llb"
+	"github.com/project-dalec/dalec"
+	"github.com/project-dalec/dalec/internal/test"
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
+)
+
+// findSpecMkfile iterates over all LLB ops and returns the path and data of the
```
invidian: nit: I'd put tests at the top. As new reader of this file, I'd expect to see here most important test in this file, instead of a helper  which I may read speculatively, not knowing if it's actually going to be significant for the test which uses it later on.

packaging/linux/rpm/buildroot_test.go
```diff
@@ -0,0 +1,325 @@
...
+		Version:     "1.0.0",
+		Revision:    "1",
+		Description: "Test package",
+		License:     "MIT",
+	}
+}
+
+func TestRPMSpec_LLB(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
```
invidian: nit: could use t.Context() in every test.

packaging/linux/rpm/buildroot_test.go
```diff
@@ -0,0 +1,325 @@
...
+
+	t.Run("directory_and_filename", func(t *testing.T) {
```
invidian: This test name reads as `TestRPMSpec_LLB/directory_and_filename`, then asserting on 2 different paths. I feel that does not communicate a lot why do we need this test or what behavior does it test. Could we improve that somehow?

packaging/linux/rpm/buildroot_test.go
```diff
@@ -0,0 +1,325 @@
...
+	return &dalec.Spec{
+		Name:        name,
+		Version:     "1.0.0",
+		Revision:    "1",
+		Description: "Test package",
+		License:     "MIT",
+	}
+}
+
+func TestRPMSpec_LLB(t *testing.T) {
```
invidian: Can we test this at integration test level or as blackbox?

packaging/linux/rpm/template.go
```diff
@@ -51,6 +51,7 @@ BuildArch: noarch
 {{ .PreUn -}}
 {{ .PostUn -}}
 {{ .Files -}}
+{{ .SubPackages -}}
```
invidian, line 54: With this amount of string builder code, have we tried leveraging templates more instead of string builder? Maybe it would make things more readable? Right now all this code here is rather hard to follow tbh.

invidian (approved):
Some minor comments, but LGTM overall. I did not analyze test coverage in depth, I assume it's sufficient.

packaging/linux/rpm/template.go
cpuguy83: I think its something we could take a look at.
Templates have their own trade-off/readability issues, though.

packaging/linux/rpm/buildroot_test.go
cpuguy83: I split this out into a separate test and renamed... all the tests to make things more clear.

packaging/linux/rpm/buildroot_test.go
cpuguy83: I don't think we can test it as integration, but we are not calling anything private here.

packaging/linux/rpm/buildroot_test.go
cpuguy83: Yep, done. I forgot t.Context finally became a thing.

packaging/linux/rpm/buildroot_test.go
cpuguy83: Sure, done.

test/subpackage_test.go
cpuguy83: Done.

test/subpackage_test.go
cpuguy83: Done.

test/subpackage_test.go
cpuguy83: Done.

invidian (approved):

invidian (approved):

PR #1152, opened 2026-07-20
feat(suse): add SLES15 rpm/container build target

djshah19:
Add SUSE Linux Enterprise 15 as a new rpm distro target built with zypper.

- suse/sles15.go, suse/common.go: ConfigSLES15 distro config (bci-base:15.7 worker, zypper installer, /etc/zypp/repos.d repo layout, native-arch only)
- distro/zypper_install.go: ZypperInstall installer reusing the shared dnfInstallConfig plumbing (--non-interactive, key import, no --releasever)
- distro.go/worker.go: CrossArchInstallUnsupported flag; worker paths fail fast for SUSE since zypper lacks dnf's --forcearch/--installroot
- plugin/init.go: register sles15 routes
- test/target_suse_test.go: TestSLES15 with SUSE helpers (no %{?dist} tag, createrepo_c/zypp repo dir, rpmsign via rpm-build)
- docs: add sles15 to targets.md and examples/targets.md

Copilot-Session: 89d10fb5-5e8e-4819-84f8-f0a96dc28fcd

This PR adds **SUSE Linux Enterprise 15 (SLES 15)** as a first-class rpm build target
(`sles15/rpm`, `sles15/container`, `sles15/worker`, `sles15/container/depsonly`), so
dalec specs can produce and package rpms for SLES the same way they already do for
azlinux, almalinux, and rockylinux. SLES is a widely deployed enterprise Linux that
consumers of dalec need to ship packages for, and it was previously unsupported.

The main difference from the existing rpm targets is the package manager: SLES uses
**zypper/libzypp**, not dnf. Rather than fork the rpm build path, this PR reuses the
shared `dnfInstallConfig` plumbing and adds a zypper-specific install step:

- **Distro config** (`suse/sles15.go`, `suse/common.go`): `ConfigSLES15` pins the
  `registry.suse.com/bci/bci-base` worker, wires up the zypper installer, and uses
  SUSE's repo layout (`/etc/zypp/repos.d`).
- **Installer** (`distro/zypper_install.go`): `ZypperInstall` runs `zypper` with
  `--non-interactive` and license/vendor flags, imports repo signing keys, and omits
  dnf-only concepts such as `--releasever`. Documentation exclusion (the default, since
  `IncludeDocs` is false) is handled via libzypp's `rpm.install.excludedocs` in
  `zypp.conf` — zypper has no CLI flag equivalent to dnf's `--setopt=tsflags=nodocs`.
- **Native-arch-only guard** (`distro/distro.go`, `distro/worker.go`): zypper has no
  equivalent of dnf's `--forcearch`/`--installroot`, so cross-architecture installs are
  not supported. A new `CrossArchInstallUnsupported` flag makes the worker paths fail
  fast for SLES instead of producing a broken foreign-arch rootfs.
- **Registration** (`plugin/init.go`): registers the `sles15/*` build routes.
- **Tests** (`test/target_suse_test.go`): `TestSLES15` exercises the target through the
  shared linux test harness, with SUSE-specific helpers that account for SLES quirks —
  an empty `%{?dist}` tag (unlike `el9`/`azl3`), repo creation with `createrepo_c`, and
  package signing with `rpmsign` (shipped in `rpm-build`, not a separate `rpm-sign`
  package).
- **Docs**: adds sles15 to `website/docs/targets.md` and `examples/targets.md`.

Fixes #

- **Worker base image is pinned to `bci-base:15.7`** (not 15.6). SLES 15 SP6 (15.6)
  reached end of general support on 2025-12-31, so builds start from the currently
  maintained 15.7 BCI. Bumping the tag also requires updating the expected OS release
  (`VersionID`) in the test config.
- **Cross-architecture builds are intentionally unsupported.** zypper cannot install
  into a foreign-arch root, so `ConfigSLES15` sets `CrossArchInstallUnsupported: true`
  and the worker fails fast rather than silently producing a wrong-arch image. The SLES
  target is therefore validated on its native architecture only.
- **Doc exclusion via `zypp.conf`, not a CLI flag.** An earlier revision passed
  `--rpm-installexcludedocs`, which is not a valid zypper 1.14 option and failed every
  install; this now sets `rpm.install.excludedocs = yes` in `zypp.conf` (honoring
  `--root`) when docs are excluded, matching the dnf targets' `tsflags=nodocs` behavior.
- The diff also includes an in-flight **merge with `main`**, so some unrelated upstream
  changes (e.g. the worker-resolution test refactor in `platform_worker_test.go`, the
  `GomodEdits` `Drop` field, and dependency bumps) appear in the file list. The
  SUSE-specific changes are the `suse/`, `zypper_install.go`, `distro.go`/`worker.go`,
  `plugin/init.go`, `target_suse_test.go`, and docs entries listed above.


targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,110 @@
...
+
+func zypperCommand(cfg *dnfInstallConfig, zypperSubCmd []string, zypperArgs []string) llb.RunOption {
+	const importKeysPath = "/tmp/dalec/internal/zypper/import-keys.sh"
+
+	// zypper cannot install into a foreign-architecture rootfs the way dnf can
+	// (it has no --forcearch). cfg.root is only set for cross-arch/rootfs
+	// installs, which are guarded out for zypper-based distros at the worker
+	// level (Config.CrossArchInstallUnsupported), so this is a best-effort
+	// native rootfs install only.
+	rootFlag := ""
```
invidian: Maybe make it a slice to append then join at the end, rather than string building? That should be more readable and less error prone I think.

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,110 @@
...
+	// enables for this invocation when docs should be excluded.
+	excludeDocs := "false"
+	if !cfg.includeDocs {
+		excludeDocs = "true"
+	}
+
+	installScriptDt := `#!/usr/bin/env bash
+set -eux -o pipefail
+
+import_keys_path="` + importKeysPath + `"
```
invidian: Maybe use a template rather than string joining? You could then use things like `%t` for includeDocs for example?

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,110 @@
...
+zypper_sub_cmd="` + strings.Join(zypperSubCmd, " ") + `"
+install_flags="` + installFlags + `"
+exclude_docs="` + excludeDocs + `"
+root_dir="` + cfg.root + `"
+
+if [ -x "$import_keys_path" ]; then
+	"$import_keys_path"
+fi
+
+if [ "$exclude_docs" = "true" ]; then
```
invidian: With template, this could be an optional block, rather than a runtime check.

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,110 @@
...
+fi
+
+zypper $global_flags $zypper_sub_cmd $install_flags "${@}"
+`
+	var runOpts []llb.RunOption
+
+	installScript := llb.Scratch().File(llb.Mkfile("install.sh", 0o700, []byte(installScriptDt)), cfg.constraints...)
+	const installScriptPath = "/tmp/dalec/internal/zypper/install.sh"
+
+	runOpts = append(runOpts, llb.AddMount(installScriptPath, installScript, llb.SourcePath("install.sh"), llb.Readonly))
```
invidian: Maybe initialize runOpts here? Like:
```go
runOpts := []llb.RunOption{
  llb.AddMount(installScriptPath, installScript, llb.SourcePath("install.sh"), llb.Readonly)
}
```

It's more concise code.

invidian (approved):
Some nits, otherwise LGTM. Perhaps @cpuguy83 would also like to have a look, since I might be missing stuff.

invidian (changes requested):
For sure changes to `matrix.json` file are missing, generated by `go generate ./...`.

And DCO. I also wonder why CI hasn't kicked in...

targets/linux/rpm/distro/zypper_install.go
djshah19: addressed!

targets/linux/rpm/distro/zypper_install.go
djshah19: addressed

targets/linux/rpm/distro/zypper_install.go
djshah19: addressed

targets/linux/rpm/distro/zypper_install.go
djshah19: addressed

djshah19:
> Some nits, otherwise LGTM. Perhaps @cpuguy83 would also like to have a look, since I might be missing stuff.

@invidian Thanks for taking a look. I addressed your comments. Could you please review again? Thanks

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,111 @@
...
+	installFlags := []string{"--auto-agree-with-licenses", "--allow-downgrade", "--allow-vendor-change"}
+	installFlagsStr := strings.Join(installFlags, " ")
+
+	// zypper/libzypp has no command-line flag equivalent to dnf's
+	// --setopt=tsflags=nodocs: passing "--rpm-installexcludedocs" is rejected as
+	// an unknown option and fails the whole install before any package is laid
+	// down. Documentation exclusion is instead controlled by libzypp's
+	// "rpm.install.excludedocs" option in zypp.conf. When docs are excluded,
+	// we inject a one-off config block into the generated install script.
+	excludeDocsConfigBlock := ""
```
invidian: Maybe we can use a template for all? something like:
```patch
diff --git targets/linux/rpm/distro/zypper_install.go targets/linux/rpm/distro/zypper_install.go
index f0f95f2f..48fafd2a 100644
--- targets/linux/rpm/distro/zypper_install.go
+++ targets/linux/rpm/distro/zypper_install.go
@@ -1,8 +1,10 @@
 package distro

 import (
+	"bytes"
 	"fmt"
 	"strings"
+	"text/template"

 	"github.com/project-dalec/dalec"
@@ -42,15 +44,28 @@ func zypperCommand(cfg *dnfInstallConfig, zypperSubCmd []string, zypperArgs []st
 	installFlags := []string{"--auto-agree-with-licenses", "--allow-downgrade", "--allow-vendor-change"}
 	installFlagsStr := strings.Join(installFlags, " ")

-	// zypper/libzypp has no command-line flag equivalent to dnf's
-	// --setopt=tsflags=nodocs: passing "--rpm-installexcludedocs" is rejected as
-	// an unknown option and fails the whole install before any package is laid
-	// down. Documentation exclusion is instead controlled by libzypp's
-	// "rpm.install.excludedocs" option in zypp.conf. When docs are excluded,
-	// we inject a one-off config block into the generated install script.
-	excludeDocsConfigBlock := ""
-	if !cfg.includeDocs {
-		excludeDocsConfigBlock = `
+	// zypperInstallScriptTmpl renders the bash script that drives a zypper
+	// install. The scalar inputs are emitted through the shellQuote func so they
+	// land in the script with the same %q quoting the previous fmt.Sprintf
+	// produced.
+	zypperInstallScriptTmpl := template.Must(template.New("zypper-install").Funcs(template.FuncMap{
+		"shellQuote": func(s string) string { return fmt.Sprintf("%q", s) },
+	}).Parse(`#!/usr/bin/env bash
+set -eux -o pipefail
+
+import_keys_path={{ shellQuote .ImportKeysPath }}
+global_flags={{ shellQuote .GlobalFlags }}
+zypper_sub_cmd={{ shellQuote .ZypperSubCmd }}
+install_flags={{ shellQuote .InstallFlags }}
+root_dir={{ shellQuote .Root }}
+
+if [ -x "$import_keys_path" ]; then
+	"$import_keys_path"
+fi
+{{ if not .IncludeDocs }}
+# zypper/libzypp has no command-line flag equivalent to dnf's
+# --setopt=tsflags=nodocs: passing "--rpm-installexcludedocs" is rejected as an
+# unknown option and fails the whole install before any package is laid down.
 # libzypp (not the zypper CLI) controls documentation exclusion via the
 # rpm.install.excludedocs option in zypp.conf. Enable it in the config file
 # libzypp will read for this install (honoring --root when set).
@@ -61,27 +76,30 @@ if [ -f "$zypp_conf" ] && grep -Eq '^[[:space:]]*#?[[:space:]]*rpm\.install\.exc
 else
 	printf '\nrpm.install.excludedocs = yes\n' >> "$zypp_conf"
 fi
-`
+{{ end }}
+zypper $global_flags $zypper_sub_cmd $install_flags "${@}"
+`))
+
+	var installScriptBuf bytes.Buffer
+	if err := zypperInstallScriptTmpl.Execute(&installScriptBuf, struct {
+		ImportKeysPath string
+		GlobalFlags    string
+		ZypperSubCmd   string
+		InstallFlags   string
+		Root           string
+		IncludeDocs    bool
+	}{
+		ImportKeysPath: importKeysPath,
+		GlobalFlags:    globalFlagsStr,
+		ZypperSubCmd:   strings.Join(zypperSubCmd, " "),
+		InstallFlags:   installFlagsStr,
+		Root:           cfg.root,
+		IncludeDocs:    cfg.includeDocs,
+	}); err != nil {
+		panic(fmt.Errorf("rendering zypper install script: %w", err))
 	}

-	installScriptDt := fmt.Sprintf(`#!/usr/bin/env bash
-set -eux -o pipefail
-
-import_keys_path=%q
-global_flags=%q
-zypper_sub_cmd=%q
-install_flags=%q
-root_dir=%q
-
-if [ -x "$import_keys_path" ]; then
-	"$import_keys_path"
-fi
-
-%s
-
-zypper $global_flags $zypper_sub_cmd $install_flags "${@}"
-`, importKeysPath, globalFlagsStr, strings.Join(zypperSubCmd, " "), installFlagsStr, cfg.root, excludeDocsConfigBlock)
-	installScript := llb.Scratch().File(llb.Mkfile("install.sh", 0o700, []byte(installScriptDt)), cfg.constraints...)
+	installScript := llb.Scratch().File(llb.Mkfile("install.sh", 0o700, installScriptBuf.Bytes()), cfg.constraints...)
 	const installScriptPath = "/tmp/dalec/internal/zypper/install.sh"

 	runOpts := []llb.RunOption{
```

cpuguy83:
It seems like this needs a rebase? There's a bunch of content that was already merged in the diff.

djshah19:
> It seems like this needs a rebase? There's a bunch of content that was already merged in the diff.

I had rebased my branch it got polluted white fixing DCO, but I rebased again

targets/linux/rpm/distro/worker.go
```diff
@@ -157,6 +157,14 @@ func (cfg *Config) workerWithBuildPlatform(sOpt dalec.SourceOpts, buildPlat ocis
 		).Root()
 	}
 
+	// The cross-arch path below installs into a foreign-arch rootfs using dnf's
+	// --forcearch/--installroot. Package managers that lack that capability
+	// (e.g. zypper) cannot cross-compile this way, so fail fast rather than
+	// silently invoking dnf on an image that does not have it.
+	if cfg.CrossArchInstallUnsupported {
```
cpuguy83: This should fallback to using the requested platform as the worker, which would then get executed by qemu (when the host is setup correctly).

cpuguy83 (changes requested):

cpuguy83:
It looks like the CI workflow also needs to have sles15 added to the integration suite (sorry, this does't have a `go generate`  yet).

packaging/linux/rpm/template.go
```diff
@@ -1209,6 +1235,51 @@ func (w *specWrapper) DisableStrip() string {
...
+//     brp-suse policy scripts (from brp-check-suse). Those scripts (chkstat in
+//     brp-05-permissions, the find in brp-15-strip-debug) fail to operate on
+//     dalec's tmpfs --buildroot, flooding the log with "failed to open root
+//     directory" errors and, critically, never stripping the binaries. Point SUSE
+//     at the generic /usr/lib/rpm/brp-strip instead (the exact mechanism the
+//     dnf-based distros already use successfully in the same build environment).
+//     brp-strip honors %{__strip}, so the existing DisableStrip() override
+//     (%global __strip /bin/true) continues to disable stripping when requested.
+func (w *specWrapper) DistroMacros() string {
+	if strings.HasPrefix(w.Target, "sles") {
```
cpuguy83: This should be passed down as config, not set here.

packaging/linux/rpm/template.go
```diff
@@ -1209,6 +1235,51 @@ func (w *specWrapper) DisableStrip() string {
 	return ""
 }
 
+// DistroMacros emits target-specific rpm macro overrides at the very top of the
+// spec (before the preamble). It is gated on the target key, so it is a no-op for
+// all non-SUSE distros.
+//
+//   - %{_libexecdir}: dalec's cross-distro contract installs libexec artifacts to
```
cpuguy83: This doesn't look right.
From what I see SLES15 adheres to FHS 3.0
https://en.opensuse.org/openSUSE:Packaging_Conventions_RPM_Macros#%_libexecdir

targets/linux/rpm/distro/distro.go
```diff
@@ -51,6 +51,21 @@ type Config struct {
...
 	// binfmt/QEMU when available.
 	CrossArchInstallUnsupported bool
+
+	// ContainerBaseImageRequired indicates the distro cannot bootstrap a
+	// container filesystem from scratch because its package repositories publish
+	// no installable release / os-release / base-filesystem package. SUSE's
+	// public SLE_BCI repo is such a case: the only providers of
+	// `distribution-release` and `/etc/os-release` (e.g. `sles-release`) are
+	// `@System`-only and not installable, so a from-scratch install produces a
+	// container with no `/etc/os-release` and cannot satisfy the implicit
```
cpuguy83: Can we not provide this ourselves?

targets/linux/rpm/distro/worker.go
```diff
@@ -157,6 +157,18 @@ func (cfg *Config) workerWithBuildPlatform(sOpt dalec.SourceOpts, buildPlat ocis
 		).Root()
 	}
 
+	// The cross-arch path below installs into a foreign-arch rootfs using dnf's
```
cpuguy83: Can we keep docs about this where it is set?
The field itself can be documented as to its use.

cpuguy83: I think it would also be helpful to have `if cfg.CrossArchInstallUnsupported || samePlatform(..)`

targets/linux/rpm/distro/worker.go
```diff
@@ -201,6 +213,13 @@ func (cfg *Config) SysextWorker(sOpts dalec.SourceOpts, opts ...llb.ConstraintsO
 		).Root()
 	}
 
+	if cfg.CrossArchInstallUnsupported {
```
cpuguy83: I think it would also be helpful to have if cfg.CrossArchInstallUnsupported || samePlatform(..)

test/target_suse_test.go
```diff
@@ -0,0 +1,180 @@
...
+		Target: targetConfig{
+			Key:       "sles15",
+			Package:   "sles15/rpm",
+			Container: "sles15/container",
+			DepsOnly:  "sles15/container/depsonly",
+			Worker:    "sles15/worker",
+			FormatDepEqual: func(v, _ string) string {
+				return v
+			},
+			// SUSE rpms carry no %{?dist} tag, so sign-file names have no dist component.
```
cpuguy83: We should probably set a custom dist then.
We need to be able to differentiate these files from other RPM's.

cpuguy83 (changes requested):

packaging/linux/rpm/template.go
djshah19: Agreed dropped the override, SLE15 natively uses /usr/lib; test now parameterized

targets/linux/rpm/distro/distro.go
djshah19: done!

targets/linux/rpm/distro/worker.go
djshah19: done!

test/target_suse_test.go
djshah19: done!

packaging/linux/rpm/template.go
djshah19: done!

packaging/linux/rpm/template.go
```diff
@@ -1209,6 +1255,26 @@ func (w *specWrapper) DisableStrip() string {
...
+		return ""
+	}
+
+	b := &strings.Builder{}
+	for _, m := range w.Macros {
+		directive := "%global"
+		if m.Define {
+			directive = "%define"
+		}
+		fmt.Fprintf(b, "%s %s %s\n", directive, m.Name, m.Value)
```
invidian: nit: maybe implement it as `String()` on `SpecMacro`?

packaging/linux/rpm/template.go
```diff
@@ -1223,7 +1289,13 @@ func WriteSpec(spec *dalec.Spec, target string, w io.Writer) error {
 }
 
 func WriteSpecWithSourceFilter(spec *dalec.Spec, target string, filter dalec.SourceFilterConfig, w io.Writer) error {
-	s := &specWrapper{Spec: spec, Target: target, SourceFilter: filter}
+	return WriteSpecWithMacros(spec, target, filter, nil, w)
+}
+
+// WriteSpecWithMacros is like [WriteSpecWithSourceFilter] but also emits the
+// provided distro-specific rpm macro overrides at the top of the spec.
+func WriteSpecWithMacros(spec *dalec.Spec, target string, filter dalec.SourceFilterConfig, macros []SpecMacro, w io.Writer) error {
```
invidian: Does this need to be exported?

test/linux_target_test.go
```diff
@@ -3483,19 +3487,26 @@ func Value() string {
 				t.Fatal(err)
 			}
 
-			if err := validatePathAndPermissions(ctx, ref, "/usr/libexec/no_name_no_subpath", 0o755); err != nil {
+			// %{_libexecdir} is distro-dependent: dnf-based distros (and Tumbleweed)
+			// use /usr/libexec, while openSUSE Leap / SLE 15 use /usr/lib.
+			libexecDir := testConfig.LibexecDir
```
invidian: nit: maybe flip the logic so we use more common value as default and then optionally override with specific, rather than other way around?

targets/linux/rpm/suse/common.go
```diff
@@ -0,0 +1,118 @@
...
+					//   and /etc/group entries. Without these, artifact ownership
+					//   (e.g. `User: nobody`) silently falls back to uid/gid 0 and
+					//   `getent passwd nobody` returns nothing.
+					// - coreutils/grep/gzip/gawk/sed: the core text-processing
+					//   utilities every base Linux userland is expected to carry
+					//   (on SUSE `grep`, `gzip`, `gawk`, `sed` are all standalone
+					//   packages, not part of coreutils).
+					"system-user-root":   {},
+					"system-user-nobody": {},
+					"coreutils":          {},
```
invidian, line 71: That's a lot for base runtime dependencies. All our other distros only define tzdata and ca-certificates as default runtime dependencies, since we aim to produce as small images as possible.

Also, standalone binaries and containers actually rarely actually depend on shell utilities, so that doesn't seem like a right selection.

targets/linux/rpm/suse/common.go
```diff
@@ -0,0 +1,118 @@
...
+				"distribution-release": {},
+				"sles-release":         {},
+			},
+			Sources: map[string]dalec.Source{
+				"os-release": {
+					Inline: &dalec.SourceInline{
+						File: &dalec.SourceInlineFile{
+							Permissions: 0o644,
+							Contents: `NAME="SLES"
+VERSION="15-SP7"
```
invidian, line 97: Maybe this could be pulled somehow from worker image while bootstrapping an image to avoid keeping an explicit copy here?

targets/linux/rpm/suse/common.go
```diff
@@ -0,0 +1,118 @@
...
+VERSION_ID="15.7"
```
invidian, line 98: Also, if we decide to keep it, it should probably be part of `targets/linux/rpm/suse/sles15.go`?

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,195 @@
...
+
+{{ if not .IncludeDocs }}
+# zypper/libzypp has no command-line flag equivalent to dnf's
+# --setopt=tsflags=nodocs: passing "--rpm-installexcludedocs" is rejected as an
+# unknown option and fails the whole install before any package is laid down.
+# libzypp (not the zypper CLI) controls documentation exclusion via the
+# rpm.install.excludedocs option in zypp.conf. With --installroot, libzypp still
+# reads its configuration from the host, so enable the option in the host
+# /etc/zypp/zypp.conf.
+zypp_conf="/etc/zypp/zypp.conf"
```
invidian, line 172: We always create directory for zypp.conf it seems, therefore it should be moved outside if block and deduplicated?

invidian: Even more, only yes/no value should change depending on IncludeDocs it seems?

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,195 @@
...
+		IncludeDocs    bool
+	}{
+		ImportKeysPath: importKeysPath,
+		GlobalFlags:    globalFlagsStr,
+		ZypperSubCmd:   zypperSubCmdStr,
+		InstallFlags:   installFlagsStr,
+		IncludeDocs:    cfg.includeDocs,
+	})
+	if err != nil {
+		panic(fmt.Errorf("rendering zypper install script: %w", err))
```
invidian: Maybe instead of panicking here we can return something like `sh -c 'echo "Rendering zypper install script template failed: %w"; exit 1'`?

packaging/linux/rpm/template.go
djshah19: done!

targets/linux/rpm/distro/zypper_install.go
djshah19: done!

targets/linux/rpm/distro/zypper_install.go
djshah19: done!

packaging/linux/rpm/template.go
djshah19: Done — added `String()` on `SpecMacro`; `DistroMacros()` now just joins `m.String()`.

packaging/linux/rpm/template.go
djshah19: Good catch — unexported to `writeSpecWithMacros` (only used within the package). `SpecMacro`/`RPMMacros` stay exported since distro configs construct them cross-package.

test/linux_target_test.go
djshah19: It already defaults to the common `/usr/libexec` (empty → `/usr/libexec`); only SLE15/Leap overrides to `/usr/lib`. Can switch to a `UsesFHSLibexec bool` if you prefer that shape.

targets/linux/rpm/suse/common.go
djshah19: Not for size — SUSE builds from `llb.Scratch()` (no bci-base underneath), so this is the minimal userland. Without `system-user-nobody`/`grep`, the `artifacts` and `disable_strip` tests fail. azlinux gets the same set for free via `distroless-packages-minimal`; SUSE ships no equivalent meta-package. Open to layering on bci-base instead if you prefer.

targets/linux/rpm/suse/common.go
djshah19: Could extract `/etc/os-release` from the worker image during bootstrap instead of inlining — happy to switch if you prefer that over the hardcoded copy.

targets/linux/rpm/suse/common.go
djshah19: Agreed — if we keep it, I will move it into `sles15.go` next to the other SLES15-specific config.

targets/linux/rpm/distro/zypper_install.go
djshah19: Done — deduplicated into one block; only the value varies now: `excludedocs={{ if .IncludeDocs }}no{{ else }}yes{{ end }}`.

targets/linux/rpm/distro/zypper_install.go
djshah19: Done — replaced the panic with a build-time failing run step (`sh -c 'echo …; exit 1'`, shell-escaped).

cpuguy83: You can return an error state: `dalec.ErrorState(...)`

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,192 @@
...
+		// command-line operands into a fresh rootfs. Those rpms may be signed
+		// (rpmsign --addsign) with a key that is not present in the target
+		// rootfs's rpm keyring, so zypper would reject them. --allow-unsigned-rpm
+		// only covers *unsigned* files, not signed-but-untrusted ones, so add
+		// --no-gpg-checks for this path only. This mirrors dnf/tdnf, whose default
+		// localpkg_gpgcheck=0 skips GPG verification for command-line package
+		// files while still verifying repository packages. The repository-verified
+		// negative tests use the worker install path (cfg.root == "") and are
+		// therefore unaffected.
+		globalFlags = append(globalFlags, "--no-gpg-checks")
```
cpuguy83: We can't do this. this disables all gpg checks.

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,192 @@
...
+	})
+	if err != nil {
+		// The template is a compile-time constant, so Execute realistically only
+		// fails on a programmer error. Rather than panicking (which would crash the
+		// frontend), surface the failure at build time as a run step that prints
+		// the error and exits non-zero.
+		msg := fmt.Sprintf("rendering zypper install script: %v", err)
+		// Wrap in single quotes for the shell, escaping any embedded single quotes.
+		quoted := "'" + strings.ReplaceAll(msg, "'", `'\''`) + "'"
+		return dalec.WithRunOptions(llb.Args([]string{
```
cpuguy83: `dalec.ErrorStateOption(err)`

cpuguy83 (changes requested):

targets/linux/rpm/distro/zypper_install.go
djshah19: Good catch — fixed. Dropped `--no-gpg-checks` (which disabled repo verification too) and instead set it in `zypp.conf` on the `--installroot` path only: `repo_gpgcheck = on` (repo metadata still verified) + `pkg_gpgcheck = off`. That mirrors dnf/tdnf's `localpkg_gpgcheck=0`: skip per-package sig checks so dalec's locally-built rpm file operands install, while repository packages stay verified via the signed repo metadata. `--allow-unsigned-rpm` still covers unsigned local files.

targets/linux/rpm/distro/zypper_install.go
djshah19: Done — replaced the `sh -c 'echo; exit 1'` hack with `dalec.WithRunOptions(llb.AddMount(..., dalec.ErrorState(llb.Scratch(), err)))` so the solve fails with the wrapped Go error.

targets/linux/rpm/distro/zypper_install.go
djshah19: Done — now returns the error via `dalec.ErrorState` mounted through `WithRunOptions` (the function returns `llb.RunOption`, so the errored state is surfaced via a mount rather than a bare `StateOption`).

cpuguy83: IIRC, this function returns a `llb.RunOption`.
Here we just need to `return dalec.ErrorStateOption(err)`, no need for a dummy command to get a non-zero exit.

djshah19: You're right — `llb.StateOption` implements `RunOption` (its `SetRunOption` does `ei.State = ei.State.With(fn)`), so I'll simplify to just `return dalec.ErrorStateOption(fmt.Errorf("rendering zypper install script: %w", err))` and drop the mount wrapper. Will push once the SLES15 integration run finishes.

targets/linux/rpm/distro/zypper_install.go
invidian: TIL dalec.ErrorState! I was looking for it and I couldn't find it lol.

test/target_suse_test.go
```diff
@@ -0,0 +1,201 @@
...
+		},
+		Libdir: "/usr/lib64",
+		// openSUSE Leap / SLE 15 keep %{_libexecdir} at /usr/lib (only Tumbleweed
+		// migrated to the FHS 3.0 /usr/libexec), so libexec artifacts land there.
+		LibexecDir: "/usr/lib",
+		Worker: workerConfig{
+			ContextName:    suse.ConfigSLES15.ContextRef,
+			CreateRepo:     createZypperRepo(suse.ConfigSLES15),
+			SignRepo:       signRepoZypper,
+			TestRepoConfig: azlinuxTestRepoConfig,
```
cpuguy83, line 57: Please set `BaseImageRef: suse.ConfigSLES15.ImageRef` here. Without it, `testTargetPlatform` skips both worker-context and source-policy platform-selection coverage, leaving the SLES cross-architecture path untested while CI still passes.

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	// dalec builds its own rpm files and installs them as command-line file
+	// operands (into the worker, or into a fresh rootfs via --installroot when
+	// assembling a container from a custom base image). dalec does not sign those
+	// rpms, so they are installed via the --allow-unsigned-rpm install flag
+	// below, which permits *unsigned* command-line rpm files only. libzypp has no
+	// per-file GPG bypass equivalent to dnf's localpkg_gpgcheck=0, so a local rpm
+	// that is *signed* with a key absent from the target keyring cannot install
+	// without the global --no-gpg-checks (deliberately avoided above) -- but dalec
+	// never produces such files, so --allow-unsigned-rpm is sufficient.
```
cpuguy83: This assumption is incorrect: `Config.BuildPkg` runs RPMs through `frontend.MaybeSign`, so a configured signer produces signed local RPMs that zypper rejects when the key is absent. Please support the signing key during container installation and enable `testSignedRPMCustomBaseImage` for SUSE; the PR currently omits that shared test because this path fails.

test/target_suse_test.go
```diff
@@ -0,0 +1,201 @@
...
+			ID:        "sles",
+			VersionID: "15.7",
+		},
+		SupportsGomodVersionUpdate: true,
+	}
+	testLinuxDistro(ctx, t, cfg)
+	testSuseExtra(ctx, t, cfg, suse.ConfigSLES15.ImageRef)
+}
+
+func testSuseExtra(ctx context.Context, t *testing.T, cfg testLinuxConfig, distroImageRef string) {
```
cpuguy83, line 69: Please replace this skip with SUSE coverage for signed local RPMs: extend `testSignedRPMCustomBaseImage` to provide the generated public key through an install-time `ExtraRepos[].Keys` context and require the container build to succeed. Add the matching no-key failure case to ensure repository verification was not weakened, then remove the inaccurate claim that Dalec never signs RPMs.

cpuguy83:
Note: docs CI is failing due to an unrelated change.

Rebasing should clear that up.

targets/linux/rpm/distro/zypper_install.go
djshah19: Addressed in 7064323. The zypper path now imports install-time signing keys into both the worker RPM keyring and the target root RPM keyring when using `--installroot`, allowing signed local RPMs to be verified without disabling repository signature checks.

test/target_suse_test.go
djshah19: Addressed in 7064323. Added `BaseImageRef: suse.ConfigSLES15.ImageRef`, so the shared target-platform test now exercises SLES worker-context and source-policy platform selection.

test/target_suse_test.go
djshah19: Addressed in 7064323. Enabled the shared signed-RPM custom-base-image coverage for SUSE using an install-time `ExtraRepos[].Keys` context, added the matching no-key failure case, and removed the inaccurate statement that Dalec never signs built RPMs.

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	// If we have keys to import in order to access a repo, mount a script that
+	// imports them into the rpm keyring (zypper uses the same rpm keyring).
+	// zypper's --gpg-auto-import-keys covers most cases, but importing up front
+	// keeps parity with the dnf path for repos that reference keys via file://.
+	if len(cfg.keys) > 0 {
+		// zypper --installroot verifies local RPMs against the target root's RPM
+		// keyring, so import install-time keys into both the worker and target.
+		importScript := importGPGScript(cfg.keys, cfg.root)
```
cpuguy83: Importing the signing key into `cfg.root` makes a build-time key persist in the output image's RPM database. Instead, expose the mounted Dalec RPMs as a `type=plaindir` repo with GPG checks disabled only for that trusted local repo, then install repo-scoped exact NEVRAs at priority 1; external dependencies remain verified and no key is persisted. Please add a collision test where another enabled repo contains a higher-version package with the same name to prove the Dalec artifact is selected.

djshah19: Addressed in d498ce3c. Mounted Dalec RPMs are now exposed through an invocation-scoped `type=plaindir` repository with `gpgcheck=0` and priority 1, while external repositories retain their configured verification. The install resolves each mounted RPM to a repo-scoped exact NEVRA, and install-time keys are no longer imported into `cfg.root`, so the build signing key does not persist in the output image. Added a collision test with an enabled higher-version package of the same name to verify the Dalec artifact is selected.

test/signing_test.go
```diff
@@ -690,13 +691,96 @@ func testSignedRPMCustomBaseImage(ctx context.Context, t *testing.T, targetCfg t
...
+							Config: map[string]dalec.Source{
+								"collision.repo": {
+									Inline: &dalec.SourceInline{
+										File: &dalec.SourceInlineFile{
+											Contents: fmt.Sprintf(`[collision]
+name=Collision Repository
+baseurl=file://%s
+enabled=1
+gpgcheck=0
+priority=0
```
cpuguy83: This test does not currently prove that repository qualification selects the mounted RPM. The exact local NEVRA excludes the external `99.0.0` RPM, and zypper normalizes priority `0` to `99` while the local repository uses priority `1`, so the test still passes with `dalec-local:` removed. Please use a distinguishable same-NEVRA package at a comparable priority so the unqualified case would select the external payload.

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,304 @@
...
+			local_rpms+=( "$arg" )
+		else
+			install_args+=( "$arg" )
+		fi
+		;;
+	esac
+done
+
+if (( ${#local_rpms[@]} > 0 )); then
+	local_repo_alias="dalec-local"
```
cpuguy83: Could we give this internal repository a more distinctive alias, such as `dalec-tmp-internal-local`? `dalec-local` is generic enough to collide with a user-supplied repository alias, which can make zypper refresh and install from the wrong repository.

targets/linux/rpm/distro/zypper_install_test.go
```diff
@@ -0,0 +1,111 @@
...
+`)
+	cmd.Env = []string{
+		"PATH=/bin:/usr/bin",
+		"HTTP_PROXY=******proxy.example:3128",
+		"DALEC_RPM_PROXY_CA_BUNDLE=" + caBundle,
+	}
+
+	out, err := cmd.CombinedOutput()
+	assert.NilError(t, err, string(out))
+	assert.Assert(t, !strings.Contains(string(out), "secret"), string(out))
```
cpuguy83: This assertion cannot detect a proxy-value leak because the configured value does not contain `secret`. Please put a unique sentinel in the proxy URL and assert that exact value is absent from the output; leaving xtrace enabled currently still makes this test pass.

cpuguy83 (changes requested):

targets/linux/rpm/distro/zypper_install_test.go
djshah19: Addressed in 3d67192b. The test now places a unique sentinel in the proxy URL and asserts that the exact full proxy value is absent from xtrace output.

test/signing_test.go
djshah19: Addressed in 3d67192b. The collision repository now contains a distinguishable package with the same NEVRA as the mounted Dalec RPM and uses priority 1. This removes the higher-version shortcut, so the payload assertion exercises repository-qualified selection.

targets/linux/rpm/distro/zypper_install.go
djshah19: Addressed in 3d67192b. Renamed the invocation-scoped repository alias to `dalec-tmp-internal-local` to reduce the chance of colliding with a user-provided repository alias.

cpuguy83 (approved):

test/target_almalinux_test.go
```diff
@@ -99,5 +99,5 @@ func TestAlmalinux8(t *testing.T) {
 }
 
 func testAlmalinuxExtra(ctx context.Context, t *testing.T, cfg testLinuxConfig, distroImageRef string) {
-	testSignedRPMCustomBaseImage(ctx, t, cfg.Target, distroImageRef)
+	testSignedRPMCustomBaseImage(ctx, t, cfg.Target, distroImageRef, false)
```
invidian, line 102: nit: that `false` is cryptic. We could perhaps use a struct for args to name it.

test/signing_test.go
```diff
@@ -690,13 +691,95 @@ func testSignedRPMCustomBaseImage(ctx context.Context, t *testing.T, targetCfg t
...
 			}
 
-			containerSr := newSolveRequest(
+			containerOpts := []srOpt{
 				withSpec(ctx, t, spec),
 				withBuildTarget(targetCfg.Container),
 				withBuildContext(ctx, t, dalec.GenericPkg, signedPkgSt),
-			)
+			}
+			if withKey {
```
invidian, line 699: nit: looks like those should be 2 separate tests. Little repetition in tests is fine if that improves readability. Current version is slightly hard to read.

targets/linux/rpm/distro/zypper_install.go
```diff
@@ -0,0 +1,304 @@
...
+	// keeps parity with the dnf path for repos that reference keys via file://.
+	if len(cfg.keys) > 0 {
+		importScript := importGPGScript(cfg.keys)
+		runOpts = append(runOpts, llb.AddMount(importKeysPath,
+			llb.Scratch().File(llb.Mkfile("/import-keys.sh", 0755, []byte(importScript)), cfg.constraints...),
+			llb.Readonly,
+			llb.SourcePath("/import-keys.sh")))
+	}
+
+	cmd := make([]string, 0, len(zypperArgs)+1)
```
invidian, line 293: nit: could be `cmd := append([]string{installScriptPath}, zypperArgs...)` for brevity

invidian (approved):

targets/linux/rpm/distro/zypper_install.go
djshah19: Thanks. I considered that form, but the current allocation followed by explicit appends makes the command construction and ownership of the new slice very clear. Both forms are correct, so I would prefer to leave this as-is.

test/target_almalinux_test.go
djshah19: Good point—the boolean is less self-documenting than a named option. Since this is test-only, the current callers are limited, and the PR is now approved and green, I would prefer not to add refactoring churn here. If this helper gains another option, I will convert the arguments to an options struct.

test/signing_test.go
djshah19: Agreed that separate tests could make each scenario easier to follow. The shared runner currently keeps the substantial package build/signing setup in one place, and both scenarios are covered by named subtests. Since this is non-blocking and the current implementation is green, I would prefer to keep it as-is in this PR and split it if we need to extend these scenarios later.

PR #1158, opened 2026-07-23
feat: add registry field to nodemod generator

bchuo:
The nodemod generator ran a bare `npm install`, which always resolves from the default public npm registry. In restricted build environments where that registry is unreachable, there was no way to point the prefetch at an alternate registry or proxy.

Add an optional `registry` field to the nodemod generator that maps to `npm install --registry=<url>`, mirroring the pip generator's existing `index_url` field. When set, the generate-phase prefetch resolves all packages from the given registry.

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*:
Fixes #1159 

**Special notes for your reviewer**:

PR #1169, opened 2026-07-27
feat: add registry field to nodemod generator (#1158)

bchuo:
* feat: add registry field to nodemod generator

The nodemod generator ran a bare `npm install`, which always resolves from the default public npm registry. In restricted build environments where that registry is unreachable, there was no way to point the prefetch at an alternate registry or proxy.

Add an optional `registry` field to the nodemod generator that maps to `npm install --registry=<url>`, mirroring the pip generator's existing `index_url` field. When set, the generate-phase prefetch resolves all packages from the given registry.

* Add tests and update schema

* Address copilot comment and update test

* Update tests

* Fix test to run in parallel

---------

Fixes #


cpuguy83 (changes requested):
Can we let clients set a build-arg to set the default registry? This way we (or spec authors) don't need to define it for every node module.

PR #1177, opened 2026-07-29
expand gomod paths with glob patterns

t0rr3sp3dr0:
Support [glob patterns](https://pkg.go.dev/path/filepath#Match) in paths of the GoMod generator, allowing the DALEC manifest to remain unchanged when new modules are introduced or existing ones removed.

N/A

N/A

### DIFF
--- a/preprocess.go
+++ b/preprocess.go
@@ -4,6 +4,7 @@ import (
 	_ "embed"
 	"fmt"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"golang.org/x/mod/module"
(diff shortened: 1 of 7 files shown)


preprocess.go
```diff
@@ -16,28 +17,76 @@ import (
...
+
+		for _, gen := range src.Generate {
+			if gen == nil || gen.Gomod == nil {
+				continue
+			}
+
+			basePath := filepath.Join(sourceName, gen.Subpath)
+			oldPaths := gen.Gomod.Paths
+
+			var newPaths []string
```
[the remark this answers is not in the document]
t0rr3sp3dr0: Fixed!

preprocess.go
```diff
@@ -16,28 +17,76 @@ import (
...
+			for _, pattern := range oldPaths {
+				matches, err := sOpt.Glob(patchedState, filepath.Join(basePath, pattern))
```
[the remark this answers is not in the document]
t0rr3sp3dr0: This is very unlikely to happen, would be a burden to maintain, and is unfeasible on Windows since glob escaping is not supported on that OS.

preprocess.go
```diff
@@ -31,13 +32,67 @@ const (
...
+			if gen == nil || gen.Gomod == nil || gen.Gomod.Paths == nil {
+				continue
+			}
+
+			basePath := filepath.Join(sourceName, gen.Subpath)
+			oldPaths := gen.Gomod.Paths
+
+			newPaths := make([]string, 0, len(oldPaths))
+			for _, pattern := range oldPaths {
+				matches, err := sOpt.Glob(patchedState, filepath.Join(basePath, pattern))
```
cpuguy83, line 77: This is globging for all paths, forcing eager evaluation regardless of glob patterns being present.

I'm also thinking we could handle this inside our shell where these paths are handled.

cpuguy83 (changes requested):

preprocess.go
t0rr3sp3dr0: If no glob pattern is present, this will be a no-op. Are you worried about performance? I don't think there'll be a noticeable slowdown in paths without patterns. Since there's no function to check if a path has a glob pattern, we would have to implement one ourselves and maintain it. This wouldn't be easy, as support for new patterns could be added and given that on Windows the patterns behave differently (there's no escaping support in that OS).

If we don't handle this expansion here, things like gomod edits may not work as expected. Expanding the paths here ensures the rest of Dalec behaves exactly as before, and there's no special handling related to path expansion. The way I implemented it also makes it possible to easily expand this feature to other generators without having to worry too much about the specific package manager of other languages.

cpuguy83: It's not a no-op because we have to do a full evaluation, including fetch the content immediately (actually the current implementation is doing a new solve for every path).

For using the posix shell to handle globs, I'm not sure its a problem to have to handle the globs in 2 places. It's just a prerequisite, same as how we have to tell it to traverse those paths in both places.

PR #1196, opened 2026-08-11
Namespace build caches by environment identity

### LINKED ISSUE #1096: Rework build cache key derivation: don't namespace by target key
## Summary

Follow-up from #1091 (see [approval review](https://github.com/project-dalec/dalec/pull/1091#pullrequestreview)):

> This looks ok. I still think we should rework how cache keys are derived. I don't think deriving key by target is the right move but that's not something this PR is changing.

The build cache namespace for user-declared `Build.Caches` is derived inconsistently across target families, and two of the three families key it off the **target key** (a routing/UX identifier) rather than the **build environment** (toolchain/ABI/distro). The cache key should identify the environment that produced the artifacts, not the route used to request the build.

## Current behavior

The auto-namespace prefix is built in [`cache.go`](../blob/main/cache.go) as `<distroKey>-<platform>-<userKey>`, where `distroKey` is whatever each family passes to `CacheDir.ToRunOption`:

| Family | `distroKey` passed | Source |
| --- | --- | --- |
| deb | distro `VersionID` (e.g. `ubuntu22.04`) | `targets/linux/deb/distro/pkg.go` → `packaging/linux/deb/pkg.go` (`BuildDeb`) |
| rpm | **target key** (e.g. `azlinux3`) | `packaging/linux/rpm/rpmbuild.go` (`CacheInfo.TargetKey`) |
| windows | `path.Join(distroVersionID, targetKey)` | `targets/windows/handle_zip.go` |

So deb keys on an environment identity; rpm/windows fold the routing key into the cache identity.

## Why this is a problem

- **False sharing (correctness):** the environment behind a target key can change while the key stays constant — base image bump, `ImageRef`/`ContextRef` override, differing build-args. Same key → shared cache → artifacts from one toolchain/ABI handed to another. Build caches (Go/sccache/bazel/object files) are exactly where this corrupts.
- **False isolation (waste):** two distinct keys that resolve to the same environment won't share a cache even though they safely could.
- **Inconsistency / maintenance hazard:** because there's no single notion of cache identity, #1091 needed two different mechanisms to give a test-only alternate target a distinct namespace — bump `VersionID` for deb (`registerDebRoutes`) but rely on the differing target key for rpm/windows (`registerRoutes`) in `targets/plugin/init.go`. If any family ever changes which value it keys on, that split silently becomes wrong.

This also affects what the cache test actually proves. `testArtifactBuildCacheDir` (`test/cache_test.go`) validates "caches are isolated across the namespace boundary." Post-#1091 that boundary is "distro VersionID" for deb but "target key" for rpm/windows, so the invariant under test differs by family.

## Proposed direction

Introduce a single, explicit, environment-derived cache identity that every distro config sets, and have all three families pass **that** to `CacheDir.ToRunOption` instead of `targetKey` / ad-hoc `VersionID`:

- deb already has `VersionID` (environment-derived, read from the worker when unset) — generalize that concept.
- rpm currently has no environment identity for user build caches (only `CacheName` for the package-manager cache); add one.
- windows can reuse the same identity rather than concatenating the target key.

Benefits:
- rpm/windows stop using the routing key as a cache key (removes the false-sharing hazard).
- The test bumps that one identity field uniformly, so the `registerDebRoutes` vs `registerRoutes` special-case in `targets/plugin/init.go` collapses into one path.
- The cache test goes back to asserting environment isolation across every family.

## Acceptance criteria

- [ ] A single cache-identity concept used by deb, rpm, and windows for `Build.Caches` namespacing.
- [ ] No target family derives the user build-cache namespace from the target key.
- [ ] The `alt_testing_targets` registration uses one uniform mechanism (no deb-specific `VersionID` bump).
- [ ] `testArtifactBuildCacheDir` validates the same isolation invariant across all targets.

Related: #1034, #1091, #948

kartikjoshi21:

This PR changes user `Build.Caches` namespacing to use an explicit build-environment cache identity.

Before this change, RPM used the target key as part of the cache namespace. Windows also included the 
target key in its cache namespace. That made cache behavior depend on the route name instead of the build
environment.

This could cause two problems:
- false sharing when the environment changed but the target key stayed
  the same
- false isolation when different target keys used the same environment

This PR adds `CacheIdentity` for deb, rpm, and Windows target configs. It also updates cache wiring so `Build.Caches` use that identity instead of the target key.

The `alt_testing_targets` registration now varies cache identity uniformly across deb, rpm, and Windows.

Fixes #1096

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -115,9 +115,9 @@ func WithCacheDirConstraints(opts ...llb.ConstraintsOpt) CacheConfigOption {
 	})
 }
 
-func (c *CacheConfig) ToRunOption(worker llb.State, distroKey string, opts ...CacheConfigOption) llb.RunOption {
+func (c *CacheConfig) ToRunOption(worker llb.State, cacheIdentity string, opts ...CacheConfigOption) llb.RunOption {
 	if c.Dir != nil {
-		return c.Dir.ToRunOption(distroKey, CacheDirOptionFunc(func(info *CacheDirInfo) {
+		return c.Dir.ToRunOption(cacheIdentity, CacheDirOptionFunc(func(info *CacheDirInfo) {
 			var cacheInfo CacheInfo
 			for _, opt := range opts {
 				opt.SetCacheConfigOption(&cacheInfo)
@@ -127,7 +127,7 @@ func (c *CacheConfig) ToRunOption(worker llb.State, distroKey string, opts ...Ca
 	}
 
 	if c.GoBuild != nil {
-		return c.GoBuild.ToRunOption(distroKey, GoBuildCacheOptionFunc(func(info *GoBuildCacheInfo) {
+		return c.GoBuild.ToRunOption(cacheIdentity, GoBuildCacheOptionFunc(func(info *GoBuildCacheInfo) {
 			var cacheInfo CacheInfo
 			for _, opt := range opts {
 				opt.SetCacheConfigOption(&cacheInfo)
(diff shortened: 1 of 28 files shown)


DannyBrito (commented):
I don't have much context but looks lgtm. There is some test failures

PR #1197, opened 2026-08-11
gomod: exclude go module proxy zip cache from generated source deps

cpuguy83:

`go mod download` extracts each module's contents directly under `$GOMODCACHE/<module>@<version>` in addition to keeping a redundant zip copy under `$GOMODCACHE/cache/download/**/@v/*.zip`. `GomodDeps` was packaging both into generated RPM/DEB source packages, needlessly bloating them with a duplicate copy of every module's sources.

`GomodDeps` now always excludes that zip cache using the existing `SourceFilter` mechanism, merging the mandatory exclude with (and after) any downstream build-time source filter config. The exclude is appended last rather than first because exclude patterns are matched in order with `.dockerignore`-style `!` negation support: putting it last ensures a downstream-configured filter can't accidentally (or intentionally) re-include the zip cache via a leading `!` pattern.

The extracted module sources and the small `.mod`/`.info`/`.ziphash` cache metadata files are left untouched; only the bulky `.zip` archives are dropped. Other generators (cargohome, pip, nodemodules) are unaffected.

Fixes #

- Added unit tests (`generator_gomod_test.go`) that inspect the generated LLB directly, including a regression test proving a downstream `"!cache/download/**/*.zip"` filter can't re-include the excluded zip.
- Added an integration test (`test/source_test.go`) that runs a real build through the `debug/gomods` target and checks the actual output filesystem, covering both the default-exclude case and the negation-resistance case end-to-end.
- `go.mod`: `github.com/moby/patternmatcher` promoted from indirect to direct (used by the new negation regression unit test to validate real pattern-matching semantics).

### DIFF
--- a/generator_gomod.go
+++ b/generator_gomod.go
@@ -6,6 +6,7 @@ import (
 	goerrors "errors"
 	"fmt"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/goccy/go-yaml/ast"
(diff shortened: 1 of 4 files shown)


bchuo (approved):

invidian (commented):
Perhaps we could add an integration tests inspecting what got included in source package.

cpuguy83:
Added an integration test.

PR #1217, opened 2026-09-02
fix(sles): bootstrap base packages separately

djshah19:
Install the SLES base userland before the application RPM and its runtime dependency graph. This prevents package scriptlets from invoking a partially installed shell while preserving dependency resolution behavior for other RPM targets.

- Adds an opt-in separate base-package installation transaction and enables it for SLES 15.
- Keeps the existing combined transaction for Azure Linux, AlmaLinux, and Rocky Linux so strong application dependencies can override weak base-package recommendations.
- Squashes minimized output filesystems when separate base-package installation creates an additional layer.
- Adds focused coverage for combined and separate installation order and conditional layer squashing.

None.

The generic split-transaction approach caused the Azure Linux `ca-certs_override` integration test to fail because a weak `prebuilt-ca-certificates` recommendation was committed before the application dependency could override it. The split is therefore intentionally scoped to SLES.

### DIFF
--- a/targets/linux/rpm/distro/container.go
+++ b/targets/linux/rpm/distro/container.go
@@ -50,8 +50,9 @@ func (cfg *Config) BuildContainer(ctx context.Context, client gwclient.Client, s
 	pkgs := []string{
 		filepath.Join(rpmMountDir, "**/*.rpm"),
 	}
+	hasBasePackages := !skipBase && len(cfg.BasePackages) > 0
 
-	if !skipBase && len(cfg.BasePackages) > 0 {
+	if hasBasePackages {
 		opts := append(opts, dalec.ProgressGroup("Create base virtual package"))
 
 		var basePkgStates []llb.State
@@ -61,7 +62,26 @@ func (cfg *Config) BuildContainer(ctx context.Context, client gwclient.Client, s
 		}
 
 		basePkgs = dalec.MergeAtPath(basePkgs, basePkgStates, "/", opts...)
-		pkgs = append(pkgs, filepath.Join(baseMountPath, "**/*.rpm"))
+		if !cfg.InstallBasePackagesSeparately {
+			pkgs = append(pkgs, filepath.Join(baseMountPath, "**/*.rpm"))
+		}
+	}
+
+	worker := cfg.Worker(sOpt, dalec.Platform(sOpt.TargetPlatform), dalec.WithConstraints(opts...))
+
+	if hasBasePackages && cfg.InstallBasePackagesSeparately {
+		baseInstallOpts := []DnfInstallOpt{
+			DnfAtRoot(workPath),
+			DnfWithSourceOpts(sOpt),
+			DnfInstallWithConstraints(opts),
+			IncludeDocs(spec.GetArtifacts(targetKey).HasDocs()),
+		}
+		rootfs = worker.Run(
+			dalec.WithConstraints(opts...),
+			cfg.Install([]string{filepath.Join(baseMountPath, "**/*.rpm")}, baseInstallOpts...),
+			llb.AddMount(baseMountPath, basePkgs, llb.SourcePath("/RPMS")),
+			frontend.IgnoreCache(client, targets.IgnoreCacheKeyContainer),
+		).AddMount(workPath, rootfs)
 	}
 
 	minimize := !skipBase && spec.GetImageMinimizationProfile(targetKey) == dalec.ImageMinimizationProfileDefault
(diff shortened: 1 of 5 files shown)


targets/linux/rpm/distro/container.go
```diff
@@ -61,22 +62,39 @@ func (cfg *Config) BuildContainer(ctx context.Context, client gwclient.Client, s
...
 		basePkgs = dalec.MergeAtPath(basePkgs, basePkgStates, "/", opts...)
-		pkgs = append(pkgs, filepath.Join(baseMountPath, "**/*.rpm"))
+		if !cfg.InstallBasePackagesSeparately {
+			pkgs = append(pkgs, filepath.Join(baseMountPath, "**/*.rpm"))
+		}
+	}
+
+	worker := cfg.Worker(sOpt, dalec.Platform(sOpt.TargetPlatform), dalec.WithConstraints(opts...))
+
+	if hasBasePackages && cfg.InstallBasePackagesSeparately {
```
cpuguy83, line 72: We could probably do this generically for all.
We'll just end up with 2 layers for images, and *potentially* share a base layer between builds of the same distro.

cpuguy83: Any thoughts on this @invidian

cpuguy83 (changes requested):

targets/linux/rpm/distro/container.go
invidian: Let me look back at debian implementation.

invidian:
```go
    opts = append(opts, dalec.ProgressGroup("Install base image packages"))

    // Update the base image to include the base packages.
    // This may include things that are necessary to even install the debSt package.
    // So this must be done separately from the debSt package.
    baseImg = baseImg.Run(
      InstallLocalPkg(basePackages(ctx, input), true, opts...),
      aptProxyConfig(input.SOpt),
      dalec.WithMountedAptCache(c.AptCachePrefix, opts...),
```

I mean, the implementation is quite distro-specific, but if you want to keep it as a separate layer explicitly, maybe that make sense.

Minimal deb images squashes the layers later on anyway to reduce image size.

cpuguy83: Oh that's right. We'll need to squash anyway for minimization.

djshah19: Thanks for the suggestion. I tried making the split transaction generic across all RPM targets, but the Azure Linux `ca-certs_override` integration test exposed a dependency-resolution regression: the first transaction committed the weakly recommended `prebuilt-ca-certificates`, so the second transaction could not install the application’s strong `ca-certificates` dependency because of the `ca-certificates-shared` conflict.

I have therefore kept the behavior opt-in and enabled it for SLES 15 only. Other RPM targets retain the combined transaction, while minimized SLES images squash the final filesystem so the extra bootstrap layer does not increase image size.

Validation completed against the current commit (`566fce86`):
- Azure Linux `TestAzlinux3/ca-certs_override` passes locally.
- The SLES 15 containerd RPM from Azure/dalec-build-defs#21050 builds successfully with the local frontend.
- A scratch-based minimized SLES container performs the separate installs, runs `Squash RPM container`, produces one layer, and launches Bash successfully.

test/target_suse_test.go
```diff
@@ -64,12 +67,121 @@ func TestSLES15(t *testing.T) {
...
+		ctx := startTestSpan(ctx, t)
+
+		spec := testLinuxSpec(t, dalec.Spec{
+			Dependencies: &dalec.PackageDependencies{
+				Runtime: dalec.PackageDependencyList{
+					"bash":        {},
+					"coreutils":   {},
+					"grep":        {},
+					"libopenssl3": {},
+					"moby-runc": {
```
cpuguy83, line 89: Can we simplify this to not use moby-runc specifically since that's provided by a different repo.
Just thinking about failure modes here, we are often hit by external infrastructure issues and this introduces extra external infra for this test suite.

djshah19: The `moby-runc` dependency is intentional because the failure depends on the full zypper dependency graph. I tried smaller SLES-only fixtures, but they also passed with the old combined transaction and therefore did not cover the regression. This is the smallest fixture I found that fails on the old code (`login_defs` cannot load `libreadline.so.7`) and passes with the split bootstrap. I agree the additional repository is a failure surface; I can replace it if I find a local fixture that demonstrably fails against the old implementation.

PR #1241, opened 2026-09-15
Opt in to upstream image source labels via frontend input

### LINKED ISSUE #1240: [REQ] Automatically populate upstream OCI source labels for Renovate changelog discovery

New feature

Have Dalec's default container-image build process populate `org.opencontainers.image.source` with the **upstream component's source repository URL**, using authoritative spec/source metadata and preserving explicit user overrides.

#### Motivation

[Azure/AgentBaker](https://github.com/Azure/AgentBaker) uses Renovate to update MCR images built with Dalec from [Azure/dalec-build-defs](https://github.com/Azure/dalec-build-defs). We want the resulting dependency-update PRs to include upstream release notes/changelogs without maintaining a separate repository mapping in every consuming Renovate configuration.

Renovate's [Docker datasource documentation](https://docs.renovatebot.com/modules/datasource/docker/) explains that it discovers `sourceUrl` from `org.opencontainers.image.source` (or legacy `org.label-schema.vcs-url`) on the **latest stable image**. This source URL enables subsequent changelog discovery; this request is not for a new OCI `changelogUrl` label.

The URL must identify the component's upstream repository, **not** the Dalec frontend repository, the build-definition repository, or the base image's repository. For example, a CoreDNS image should identify `https://github.com/coredns/coredns`, rather than `https://github.com/Azure/dalec-build-defs`.

#### Current building blocks

Dalec already supports explicit [`image.labels`](https://github.com/project-dalec/dalec/blob/23446b6d993f8058ab1aa92eda895e4b0a9960ac/website/content/image.md#labels), and [`MergeImageConfig`](https://github.com/project-dalec/dalec/blob/23446b6d993f8058ab1aa92eda895e4b0a9960ac/image.go) merges them into the output image configuration. The missing capability requested here is **default population**, not basic label support.

As a concrete example, the [CoreDNS build template](https://github.com/Azure/dalec-build-defs/blob/05b8e9dd6a9f843319e65769f8f1047785cb84df/specs/coredns/templates/coredns.yml) already identifies the upstream repository in `sources.coredns.git.url`, but its `image` section supplies only an entrypoint and does not explicitly set the source label.

#### Proposed behavior

- Automatically emit `org.opencontainers.image.source` for container outputs when the primary upstream repository is unambiguous. A single authoritative Git source is a useful initial case; normalize its repository URL into a form supported by consumers such as Renovate.
- Provide/document how authors identify the primary upstream when a spec has multiple sources, uses source archives, or otherwise cannot be resolved confidently. Do not choose an arbitrary source or assume `website` always points to a source repository.
- Preserve explicit spec-level and target-level label overrides and their existing precedence. Do not mistakenly retain a base-image source label as the packaged application's source. Document behavior when no upstream can be determined and how to opt out of default population.
- Write the label into the final image configuration's `config.Labels`, which Renovate can read. Manifest/index annotations alone should not substitute for this without verified consumer compatibility. Apply consistently to each platform image and supported container target, including builds with no existing `image.labels` configuration.

Illustrative expected output for CoreDNS:

```json
{
  "config": {
    "Labels": {
      "org.opencontainers.image.source": "https://github.com/coredns/coredns"
    }
  }
}
```

#### Acceptance criteria

- A representative spec with an unambiguous upstream Git source produces the correct label without manually duplicating the URL under `image.labels`.
- Tests cover explicit overrides, inherited base-image labels, ambiguous/missing upstream metadata, and platform/target consistency.
- A published image's label is discoverable by Renovate's Docker datasource as `sourceUrl`, without a consumer-side source URL override; an upstream version update with matching releases/tags can then resolve release notes.
- Documentation explains source selection, overrides, and limitations. Existing package-only output behavior remains unchanged.

#### Scope and limitations

This provides the missing metadata for **upstream** changelog discovery; it does not guarantee release-note generation for every image tag. Image tags with packaging revisions (for example, `vX.Y.Z-N`) may still require version/tag mapping in Renovate. Dalec-specific patches, toolchain changes, and revision-only CVE rebuilds are not necessarily described by upstream release notes; aggregating those changes with upstream changelogs is a separate concern.

The standardized key is defined in the [OCI annotation specification](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys).

🤖 *Generated by GitHub Copilot*

djsly:

Add **opt-in** upstream component source-label inference for container outputs. This supersedes the original automatic-default proposal: inference and inherited-provenance cleanup are disabled unless the caller supplies the **raw frontend input `dalec.image-source-label=true`**, not a build argument.

- Infer `config.Labels["org.opencontainers.image.source"]` from exactly one top-level Git source after argument substitution, per target/platform. Support HTTP(S), including SourceHut `~user` paths; retain `.git`; omit credentials, query strings and fragments; reject local/private-address and non-canonical numeric IPv4 forms. Never infer from `website`, archives, build definitions or base images.
- With the input absent or false, neither infer nor remove inherited provenance. Existing explicit global/target label merging is unchanged. A same-named build argument cannot enable inference.
- When enabled, remove inherited OCI source/revision and legacy vcs-url/vcs-ref before applying component metadata. This avoids legacy fallback to the base repository or pairing a base commit with the component source. Preserve unrelated base labels and explicit provenance; do not infer a revision.
- Explicit global/target OCI or legacy source keys override inference, including empty values; target labels win. An empty OCI override suppresses inherited legacy fallback only when enabled. Explicit legacy labels remain intentional and must separately be emptied to suppress all discovery.
- Share configuration across Linux normal/minimal and Windows container outputs. **RPM `container/depsonly` outputs do not infer the application's source**, because the application is not included. They preserve explicit image settings and apply the same enabled-only inherited-provenance cleanup. Package-only outputs remain unchanged.
- HTTP(S) does not imply a public repository: opt in only when its URL is appropriate to disclose.

Fixes #1240

Latest review fixes: `921fcb3231c6eecf8874beb01265ba75122ccb45`. Both new findings were reproduced before fixing them: SourceHut URL/config tests rejected the valid URL, and the corrected deps-only integration test reported the erroneous CoreDNS source.

Latest local gates, Linux Go 1.26.5:

- `go test -p 2 -count=1 --test.short --timeout=10m ./...` passed.
- Focused race tests for URL/source configuration and frontend-input parsing passed. Custom lint, `go generate ./...`, and byte comparisons of all generated files passed. CI-pinned golangci-lint 2.12.2 reported **0 issues**.
- **65 source-label integration cases passed**: 13 each for Bookworm normal/minimal, Azlinux3 normal/deps-only, and Windows cross-container metadata. Covers absent/false, inert build argument, invalid input, single/ambiguous/no Git source, explicit global/target OCI and legacy labels, empty overrides and explicit revisions. Windows local validation is cross-build metadata, not Windows runtime execution.
- The deps-only full-spec regression checks actual curl installation, absence of the application artifact, and absence of inferred application provenance with a CoreDNS Git source.
- **Both DEB/RPM package-only controls passed**: absent/enabled configurations match except the existing per-solve creation timestamp.

Exact successful integration selectors (with `go test -p 2 -count=1 -timeout=59m -v ./test -run=...`):

```text
^Test(Bookworm|Azlinux3)$/(^container$|^minimal_container$)/^image_configs$/^source_labels$
^Test(Bookworm|Azlinux3|Windows)$/(^container$|^minimal_container$|^depsonly$|^test_image_configs$|^package_source_labels$)/(^image_configs$/^source_labels$|^source_labels$|^full_spec$)
^Test(Bookworm|Azlinux3)$/^package_source_labels$
```

The second selector exercised Windows/deps-only matrices and the deps-only contents case; the first and third separately cover normal/minimal matrices and package-only controls. Counts were checked from the actual test output.

Rebuilt the frontend from the repository Dockerfile (`sha256:5eb7415a92b3a20b1438f9577837590b5437a0d50d1cebb5403c12716d561704`) and exported actual OCI images through it. Against a labeled Azure Linux base, all four `azlinux3/container/depsonly` exports contain curl and no CoreDNS executable:

| Input / override | Exported provenance | Actual Renovate 44.90.2 `sourceUrl` |
|---|---|---|
| `true` | All four inherited source/revision labels removed; no application inference | absent |
| absent | All four inherited provenance labels preserved | `https://example.com/base` |
| `false` | All four inherited provenance labels preserved | `https://example.com/base` |
| `true` + explicit target source/revision | Explicit target values preserved | `https://example.com/explicit-target` |

Unrelated base labels survived every case. Enabled deps-only manifest: `sha256:aa3bc4a2acb179f72aecd5819e9baf5ce89f50d2cb39052c1acf96b6c8d84aa0`.

Re-exported the representative **CoreDNS v1.12.0 normal container** with this frontend, using the previously validated compilation/package inputs. Unmodified Renovate again discovered **`https://github.com/coredns/coredns.git`** from its actual config blob. Manifest: `sha256:4cad3043e5362e19649d92e1ca2c269b10ff72f1305ca5a1e466746aa8bd4fbc`.

Prior redesign gates also validated normal-container absent/false/empty-label controls and ran the compiled CoreDNS binary read-only without network. A separate consumer follow-up used released Renovate 44.90.2 code, real OCI bytes and captured public GitHub responses: `lookupUpdates` selected `1.11.0-1` to `1.12.0-1`, and the actual PR changelog/body functions rendered the CoreDNS v1.12.0 release notes. Raw `.git` and normalized source URL controls produced the same release-note section. **This was offline HTTP replay, not a live authenticated Renovate bot PR.** Missing-token and revision-only controls did not invent release notes.

All images/registries were local task resources; no test images were pushed publicly and no Renovate dependency was added to Dalec. Harnesses, exact export/config hashes, logs and consumer evidence are retained separately. Upstream release notes do not necessarily describe Dalec-specific CVE patches, toolchain changes or revision-only rebuilds.

The previous head's upstream checks passed. CI for this follow-up and maintainer approval remain separate gates.

🤖 *Generated by GitHub Copilot*


imgconfig.go
```diff
@@ -69,3 +105,75 @@ func MergeSpecImage(spec *Spec, targetKey string) *ImageConfig {
...
+	u := url.URL{Scheme: remote.Scheme, Host: remote.Host, Path: remote.Path}
+	host := strings.TrimSuffix(strings.ToLower(u.Hostname()), ".")
+	if host == "" || host == "localhost" || strings.HasSuffix(host, ".localhost") || strings.HasSuffix(host, ".local") {
+		return ""
+	}
+	if ip, err := netip.ParseAddr(host); err == nil {
+		ip = ip.Unmap()
+		if ip.IsPrivate() || ip.IsLoopback() || ip.IsUnspecified() || ip.IsLinkLocalUnicast() {
+			return ""
+		}
```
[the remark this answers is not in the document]
djsly: Fixed in fca840dd5031f256b9a9caf55a3850cb53d28f3d. If strict IP parsing fails, inference now rejects hosts made entirely of decimal/hexadecimal numeric components rather than guessing how a consumer will resolve them. This covers shortened, integer, octal, hexadecimal, mixed-radix, zero-padded, trailing-dot/port, and overflowing forms; canonical public IPv4 and ordinary DNS names with numeric or hex-like subdomains remain supported. No DNS lookups are introduced.

Regression tests reproduced the bypass before the fix and now pass, including image-config cases confirming inherited source/revision labels stay suppressed when inference rejects the URL. The full Linux short suite, frontend build, root-package race tests, custom lint, and generation checks pass. Documentation now calls out the numeric-host exclusion.

🤖 *Generated by GitHub Copilot*

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

djsly:
Investigated the three failed jobs in CI run 35002253803:

- **lint:** fixed both PR-introduced staticcheck findings (QF1001 and QF1003) in 54e0492d7ebff471b5b1c90d13fd86b6c1f05714. The CI-pinned golangci-lint v2.12.2 now reports **0 issues** locally on Linux. The full Linux short suite, frontend build, root-package race tests, custom lint and generation checks also pass. Added port regression cases to verify the simplification preserves URL-inference behavior.
- **Both E2E matrix jobs:** fail building the unchanged Azure Linux runc fixture because the current Go toolchain rejects `GOEXPERIMENT=nosystemcrypto` with “has been removed; use MS_GO_NOSYSTEMCRYPTO=1.” The same failure occurs on `main` at 11ca9cd2cb7005c78604690c88fc15e4c14b6f2d: [E2E diff/merge=0](https://github.com/project-dalec/dalec/actions/runs/35002322397/job/104493471823), [E2E diff/merge=1](https://github.com/project-dalec/dalec/actions/runs/35002322397/job/104493472382). `test/fixtures/moby-runc.yml` is byte-identical here and on current main (blob 0fe49c1b2d43f7d106680a93a753113f171dc7de). I have kept this unrelated fixture/toolchain repair out of the source-label PR; the shared baseline needs that fix before E2E can turn green.

On the previous PR head, all 16 integration matrix jobs, unit tests, CodeQL and dependency review passed. These are results for fca840dd, not yet the new-head rerun.

🤖 *Generated by GitHub Copilot*

cpuguy83:
I think this needs to be opt-in via a frontend input (not build-arg which is a *special* frontend input, though that does mean that docker-cli users can't enable it).

We don't want to suprise people by automatically doing this and they've got some closed source project they are building.

I also think we should only do this for http(s) URL's.

No need to drop `.git` from the URL unless its some problem with tooling that's picking this up.

Github will redirect the `.git` URL, at least.

djsly:
@cpuguy83 Addressed in 41328ac2f8c47ddced1b306d03bd3ed13a56c285:

- Inference now requires the raw frontend input `dalec.image-source-label=true`, not a build argument. Absent/false preserves inherited provenance and existing behavior.
- Only HTTP(S) URLs are eligible; `.git` is retained. The docs and PR description now supersede the original automatic-default proposal.
- Fresh real-image validation confirms the new behavior: compiled and ran CoreDNS v1.12.0, then unmodified Renovate **44.90.2** selected the latest stable fixture tag and fetched its OCI config, returning exactly **`https://github.com/coredns/coredns.git`**. Absent/false controls retain the seeded base provenance; enabled plus an explicit empty source returns no `sourceUrl`, including no inherited legacy fallback.

All 50 source-label scenarios across DEB normal/minimal, RPM normal/deps-only and Windows cross-container metadata passed, as did both package-only controls, the full Linux short suite, targeted race tests and CI-pinned lint (0 issues). Precise commands and caveats are in the updated PR body. The consumer check proves source discovery, not release-note or packaging-tag mapping.

Ready for another review; new-head CI is separate from these completed local gates.

🤖 *Generated by GitHub Copilot*

djsly:
@cpuguy83 The opt-in redesign is implemented and validated, and ready for review. Could you confirm this matches the design you had in mind?

Inference requires the raw frontend input `dalec.image-source-label=true` (not a build argument). Absent/false preserves inherited provenance; enabled inference accepts HTTP(S) only and retains `.git`.

The real-image checks and Renovate 44.90.2 source discovery passed, returning `https://github.com/coredns/coredns.git`. The numeric IPv4 review finding is also fixed and its thread resolved. Full validation details are in the updated PR description.

🤖 *Generated by GitHub Copilot*

imgconfig.go
```diff
@@ -69,3 +125,80 @@ func MergeSpecImage(spec *Spec, targetKey string) *ImageConfig {
...
+		return ""
+	}
+
+	u.Path = strings.TrimSuffix(u.Path, "/")
+	u.Path = "/" + strings.TrimPrefix(u.Path, "/")
+	if u.Path == "/" || strings.ContainsAny(u.Path, "\\$") {
+		return ""
+	}
+	for part := range strings.SplitSeq(strings.TrimPrefix(u.Path, "/"), "/") {
+		if part == "" || part == "." || part == ".." || strings.HasPrefix(part, "~") {
```
cpuguy83: **P2: Allow tilde-prefixed paths in HTTPS repository URLs**

This rejects valid repositories such as `https://git.sr.ht/~sircmpwn/aerc`. Calling `BuildImageConfig` with that single Git source and `WithImageSourceLabel()` returns no source label, while the GitHub control works. A `~user` segment in an HTTPS URL is not a local home-directory path; SSH/SCP URLs are already excluded by the scheme check above. Please remove the tilde-prefix restriction and add a SourceHut regression case.

targets/linux/rpm/distro/container.go
```diff
@@ -144,7 +148,7 @@ func (cfg *Config) HandleDepsOnly(ctx context.Context, client gwclient.Client) (
 			return nil, nil, err
 		}
 
-		img, err := linux.BuildImageConfig(ctx, sOpt, spec, platform, targetKey)
+		img, err := linux.BuildImageConfig(ctx, sOpt, spec, platform, targetKey, imageOpts...)
```
cpuguy83: **P2: Do not infer application provenance for deps-only images**

This infers from the original `spec.Sources`, but the package installed in this image is built from `depsSpec`, which includes only runtime dependencies and no application sources. For example, a spec with the CoreDNS Git repository and `curl` as its runtime dependency produces a deps-only image labeled `https://github.com/coredns/coredns.git`, even though CoreDNS is not included. Consumers can then associate CoreDNS changelogs with an image containing only its dependencies. Please preserve explicit image labels but avoid inferring the application's source for this target; the deps-only source-label expectations should reflect that distinction.

cpuguy83 (changes requested):
Two correctness issues to address: valid SourceHut HTTPS repositories are excluded, and deps-only images acquire the source label of an application they do not contain.

review by GPT-6 Astra

imgconfig.go
djsly: Fixed in 921fcb3231c6eecf8874beb01265ba75122ccb45. Removed the tilde-prefix restriction for HTTP(S) paths and added your exact SourceHut URL to the URL and image-config regressions, plus HTTP/HTTPS tilde-path and local-home-path controls. The config test covers enabled/disabled inference with and without inherited labels. These regressions reproduced the rejection before the fix and pass now; SSH/SCP/local-path and numeric-host filtering remain unchanged.

🤖 *Generated by GitHub Copilot*

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

targets/linux/rpm/distro/container.go
djsly: Fixed in 921fcb3231c6eecf8874beb01265ba75122ccb45. Deps-only image configuration now uses a shallow copy of the original spec with `Sources` taken from the synthetic `depsSpec`. This retains base selection and explicit global/target image settings without inferring provenance from application sources that are not included.

Corrected the deps-only matrix and added explicit legacy URL, empty-label and revision controls. The actual-content regression enables inference with a CoreDNS Git source, confirms curl is installed, and confirms the application artifact is absent. The old implementation fails the corrected inference test.

Also rebuilt the frontend and exported real `azlinux3/container/depsonly` images against a base carrying all four OCI/legacy provenance labels. Enabled: no inferred source or base fallback; absent/false: all inherited provenance preserved; explicit target source/revision: preserved. Unrelated base labels remain, curl is present and CoreDNS is absent. Unmodified Renovate 44.90.2 confirmed those exact source-discovery outcomes. All 65 source-label scenarios across five container paths and both package-only controls pass; full Linux short tests and pinned CI lint pass too.

🤖 *Generated by GitHub Copilot*

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

test/linux_target_test.go
```diff
@@ -4608,6 +4640,128 @@ func testImageConfig(ctx context.Context, t *testing.T, target string, opts ...s
...
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			spec := &dalec.Spec{
+				Name: "test-image-source", Version: "1.12.0", Revision: "1", License: "Apache-2.0",
+				Description: "Image source metadata", Image: tc.global,
+				Args:    map[string]string{"UPSTREAM": "https://example.com/must-be-substituted"},
+				Sources: make(map[string]dalec.Source),
+			}
+			if tc.buildArg {
+				spec.Args[frontend.KeyImageSourceLabel] = "false"
```
cpuguy83: This shouldn't be needed?

cpuguy83 (approved):

